PR #139, opened 2015-02-25
fix hex delimiters for python 3

jurov:
Unhexlify produces bytes under python 3, must decode it.


edreamleo:
Thanks for this patch.  It is at rev 24a9dd5.

Since this affects only Python 3, I have used this code::

    if g.isPython3:
        delims[i] = delims[i].decode('utf-8')

Please reopen if there are problems.

PR #187, opened 2015-04-27
Added support for bash files with unused shellscript.py mode and add placeholders to param_doc()

JohnLunzer:
I am not certain that shellscript.py addresses all bash syntax but it's better than none at all.

This addresses @edit nodes of .bash files loading with @nocolor. Now they should automatically load to @language shellscript.

I figured this was the easiest way of doing this since shellscript.py was not being used. But if you think this is misleading we can find another approach.

### DIFF
--- a/leo/config/leoSettings.leo
+++ b/leo/config/leoSettings.leo
@@ -10023,7 +10023,7 @@ def param_doc(k=None):
         indent = '        '
     if k in _values and _values[k]:
         return "\n\n%s:Parameters:\n%s" % (indent, '\n'.join(
-            ["%s- `%s`: describe %s" % (
+            ["%s- `%s`: &lt;|describe %s|&gt;" % (
                 indent, i.split('=',1)[0].strip(), i.split('=',1)[0].strip()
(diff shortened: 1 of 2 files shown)


JohnLunzer:
Still new to this.

I also am proposing a slight change to param_doc() which makes it work more naturally with that comma-comma abbreviation. It simply adds placeholders to the descriptions for parameters so that you can comma-comma to move to the descriptions for the parameters.

tbnorth:
param_doc() modification sounds fine to me, my current personal param_doc() is:
```
def test(this, is, a, test):
    """test - <||>

    :param <|type this|> this: <|describe this|>
    :param <|type is|> is: <|describe is|>
    :param <|type a|> a: <|describe a|>
    :param <|type test|> test: <|describe test|>
    :return: <|returns|>
    :rtype: <|return type|>
    """
    <|code|>
```
which I think is a different markup from the default, making the default thorough and letting people customize makes sense.  Ah - I remember now, the above is Sphinx oriented.

PR #192, opened 2015-05-26
Suggesting a fix, relative paths do not work in image nodes.

downpoured:
To repro, create a directory "d" and directory "subdir".
Create "d/image.png" using any valid png file.
Create "d/test.leo"
In d/test.leo, create a node with name "@image" and body "image.png"
Create "d/subdir/test2.leo"
In d/subdir/test2.leo, create a node with name "@image" and body "../image.png"
Open both d/test.leo and d/subdir/test2.leo in Leo and use Alt-0 to show the viewrendered pane, only one of the nodes shows the image correctly in the pane and the other shows a broken placeholder. Reading the code, it looks like supporting relative paths was intended but broken by a typo.
Issue is resolved with this patch which was likely the intended functionality.

### DIFF
--- a/leo/plugins/viewrendered.py
+++ b/leo/plugins/viewrendered.py
@@ -757,9 +757,9 @@ if QtWidgets:
             w = pc.ensure_text_widget()
             ok,path = pc.get_fn(fn,'@image')
             if not ok:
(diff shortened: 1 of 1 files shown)


edreamleo:
Thanks for this fix.

PR #197, opened 2015-06-17
Add per file/folder inclusion/exclusion to active_path.py

JohnLunzer:
Per Folder file/folder inclusion and exclusion by adding flags to the body of an active path folder (either @ or /*/), can include multiple inc= and exc= flags:
+    excdirs - excludes all directories 
+    excfiles - excludes all files
+    inc= - a single item or comma separated list of strings to include in the list of files/folders
+    exc= - a single item or comma separated list of strings to exclude in the list of files/folders
+    re - search using regular expressions (otherwise a case-sensitive 'in' comparison)

Please review these changes and let me know what you think.


edreamleo:
Thanks for your continued work.

PR #259, opened 2016-04-13
Vr3fixup

pmills:
Reinstated the intended functionality of VR2/3.  There are a couple of instances where "fixing things" was extended to making new but "expected behaviour" work.  See commit comments.

Even though there is on-going polishing work (see "Work still on-going" section below), I think this point of bug fixing is worth merging.

Bugs not fixed:

* Multiple VR3 panes cannot be created (this problem may be partly outside of the VR3 plugin)
* Docstring messed up in plugin's "about" render; possibly due to break-up into nodes?
* @md node doesn't invoke proper renderer (may have never worked in VR2), but @language md appears to work fine.

Followup:

* More testing of Markdown rendering.
* Because there are so many possible combinations of node/tree content, VR3 options, and available css/stylesheets, I haven't tested every combination, so I'd appreciate others doing some testing using their own workflows and outlines.  This may not be entirely easy, given that the plugin's docstring is still inadequate.
* There have been a couple of changes to the default settings of both docutils rendering (particularly on reporting errors), VR3 options and treatment of @language directives in order to have a better "default settings" experience.  Please report if these are problematic or a nuisance.
* Edward to check and, when happy, remove the VR3 switch and all of the non-VR3 code
* Suggest renaming to viewrendered.py (and internal naming changes also) to make it clear that there is *only one VR* now (VR3 would imply a choice between 3 versions).
* VR3's Leo code/node restructure?  (very hard to understand the structure of code from outline structure)
* "Lock to node" (from the Options menu) is working, but is not integrated with the existing/old "lock" command that VR1 has.  I believe they are related.

Work still on-going (by Peter Mills), hopefully ready for Leo 5.3:

* Improve the docstring (more explanation of every feature; this is a fair bit of work)
* Get @md nodes working as per VR1 (possibly never worked with VR2)

### DIFF
--- a/leo/plugins/viewrendered3.py
+++ b/leo/plugins/viewrendered3.py
@@ -181,25 +181,29 @@ The following settings are the same as in the viewrendered.py plugin:
 
 The following settings are new in the viewrendered2.py plugin:
 
-These settings directly override the corresponiding docutils settings:
+These settings directly override the corresponding docutils settings (note the
+underscores in the docutil keywords):
 
-- ``@string vr-stylesheet-path``
-- ``@int vr-halt-level = 6``
-- ``@string vr-math-output = mathjax``
-- ``@bool vr-smart-quotes = True``
+- ``@string vr-stylesheet_path = html4css1.css``
+- ``@int vr-halt_level = 6``
+- ``@string vr-math_output = mathjax``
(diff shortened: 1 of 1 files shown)


edreamleo:
Thanks for this work, and the excellent documentation.

PR #349, opened 2016-12-13
Command line option to select type of @<file> node to when opening non-outline files

JohnLunzer:
These changes are minor and pretty straight forward, I added an option at the command line `--load-type` which allows the user to provide a string. No checking is done when parsing options. 

Checking for appropriate `@<file>` node is done within `c.refreshFromDisk` which is now what is used to load the files. If an invalid `@<file>` node is provided Leo will still open without error and the Log will inform the user that the string they provided was not a valid `@<file>` node.

Current behavior is not changed, `@edit` nodes are still used as default.

I would have liked if when providing multiple files the files all opened in the same outline (to avoid loading all plugins multiple times), but I don't have the time to look into that right now, if somebody else wants to move that forward they are welcome.

### DIFF
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -2147,6 +2147,8 @@ class LoadManager(object):
             help='enable ipython support')
         add('--gui',
             help='gui to use (qt/qttabs)')
+        add('--load-type', dest='load_type',
+            help='@<file> type for loading non-outlines from command line')
         add('--maximized', action='store_true',
             help='start maximized')
         add('--minimized', action='store_true',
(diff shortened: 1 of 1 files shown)


edreamleo:
On Tue, Dec 13, 2016 at 1:06 PM, John Lunzer <notifications@github.com>
wrote:

> These changes are minor and pretty straight forward, I added an option at
> the command line --load-type which allows the user to provide a string.
> No checking is done when parsing options.
>

​Thanks, John.  I've merged it, all unit tests pass, and the help message
(-h) is informative.

EKR

PR #398, opened 2017-02-04
Tiny changes to auto-rst and auto-md headline defaults

shakfu:
Two non-controversial small changes to rst and md writer defaults: 

1. The sequence of rst underline characters are now more consistent with sphinx defaults (http://www.sphinx-doc.org/en/stable/rest.html)

2.  Added a space between the # the title text in markdown writer so `#Header 1` becomes `# Header 1`

### DIFF
--- a/leo/plugins/importers/leo_rst.py
+++ b/leo/plugins/importers/leo_rst.py
@@ -9,7 +9,8 @@ import leo.core.leoGlobals as g
 import leo.plugins.importers.linescanner as linescanner
 Importer = linescanner.Importer
 # Used by writers.leo_rst as well as in this file.
-underlines = "!\"$%&'()*+,-./:;<=>?@[\\]^_`{|}~#"
+underlines_old = "!\"$%&'()*+,-./:;<=>?@[\\]^_`{|}~#"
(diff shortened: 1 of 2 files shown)


edreamleo:
Thanks for this work!

shakfu:
You're welcome. Thank you for your awesome leo (-:

edreamleo:
Pylint is complaining about a possibly missing r in front of the new string.  That makes sense, because the new string contains a backslash.  What do you think?

shakfu:
Ok, my bad, the backslash was escaping the double quote in the old underlines sequence which is not required in the single-quoted new version, we can safely remove it as follows:
```
 underlines_old = "!\"$%&'()*+,-./:;<=>?@[\\]^_`{|}~#"                            
 underlines = '*=-^~"+!$%&(),./:;<>?@[\\]_`{|}#' 
```
and the pylint warning disappears. Also would you like me to remove the `underlines_old` var or retain it? 

I can make another pull request if you like.

edreamleo:
Please remove underlines_old.

shakfu:
done: see #399

PR #437, opened 2017-03-12
fix: initial fix to graphcanvas PyQt5 zoom problem, second try

node-ex:
Leo crashed after holding Ctrl and scrolling up (zooming) in the Graph pane with the following exception:

```
  File "/home/nodex/Software/Leo/leo/plugins/graphcanvas.py", line 209, in wheelEvent
    scale = 1.+0.1*(event.delta() / 120)
AttributeError: 'QWheelEvent' object has no attribute 'delta'
```

It is line 209 in the `graphcanvas.py` file. After some debugging and googling I have found out, that I am using PyQt 5 and in Qt 5 `delta()` method of `QWheelEvent` class is depracated and replaced by `angleDelta()` method.

PyQt4.8: http://doc.qt.io/qt-4.8/qwheelevent.html#delta
PyQt5: http://doc.qt.io/qt-5/qwheelevent.html#angleDelta

So I have created a new variable, `direction`, and, depending on whether PyQt4 or PyQt5 is used, assigned return value of either `event.delta()` or `event.angleDelta().y()` to it. Then I used it in 2 places, where the crash can happen.

Finally I have tested it with my setup:

* Leo 5.4, build 20170312075431, Sun Mar 12 07:54:31 CDT 2017
* Python 3.6.0
* PyQt 5.6.2

and it worked and Leo did not crash.

### DIFF
--- a/leo/plugins/graphcanvas.py
+++ b/leo/plugins/graphcanvas.py
@@ -48,7 +48,7 @@ except Exception:
 
 # Fail gracefully if the gui is not qt.
 g.assertUi('qt')
-from leo.core.leoQt import QtConst, QtCore, QtGui, QtWidgets, uic
+from leo.core.leoQt import QtConst, QtCore, QtGui, QtWidgets, uic, isQt5
 
 pygraphviz = None
 if not pydot:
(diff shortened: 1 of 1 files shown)


edreamleo:
Thanks for this work. Please base your work on devel, not master, and re-issue the PR.

PR #492, opened 2017-05-16
open-url improvement for Linux.

SegundoBob:
I only use Linux and I only use the i3wm window manager.  The
following is only about Linux and my patch only changes Leo-Editor
under Linux.

Problem: When I open a file URL for most file types, the Leo-Editor
process hangs till the program used to open the file terminates.  This
means that I can't use the Leo-Editor instance that I used to open the
URL while I'm examining whatever file I opened.  In addtion,
Leo-Editor can't even redraw its window, so when I shift focus away
from the Leo-Editor window and then back, the contents of the last
non-Leo-Editor window are shown. This happens for all file viewers
that I know.  For me the usual culprit is my PDF viewer Evince.

This problem does not occur for most web browsers for reasons unknown
to me.  But at various times I have been forced to change web browsers
because the problem did occur for them.  Then later the problem would
go away.  Whether this was due to changes in Leo-Editor or in the web
browser, I never knew.  If the problem were more common for web
browsers, this problem would probably have been fixed long ago.

My fix is very short in terms of code, but is relatively complicated
in what goes on.

1) I use subprocess.Popen() to create a process to use to view the URL
resource.  This allows the Leo-Editor process to be responsive while
the resource is being viewed.  This viewing process can continue to
run after Leo-Editor terminates.

2) I use an Idle Time handler to poll the subprocess stderr and status
while Leo-Editor goes about its own business.

3) If the suprocess.Popen() PIPE option is used for stdout or stderr,
then the first read of the pipe hangs.  Consequently, the subprocess
stderr must be redirected to a binary, unbuffered disk file so that it
can be polled by non-hanging read calls.

The subprocess.Popen() option shell can be either True or False.  Both
settings work.  I chose False as the safest option.

I poll once per second.  This makes echoing to the log pane adequately
responsive without overly burdening Leo-Editor.

Many URL's can be opened at once.  This works fine.

Destroy the IdleTime object.

### DIFF
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -6167,6 +6167,53 @@ def os_path_splitext(path):
     return head, tail
 #@+node:ekr.20090829140232.6036: *3* g.os_startfile
 def os_startfile(fname):
+    #@+others
+    #@+node:bob.20170516112250.1: *4* stderr2log()
+    def stderr2log(g, ree, fname):
+        """ Display stderr output in the Leo-Editor log pane
+
+        Arguments:
+            g:  Leo-Editor globals
+            ree:  Read file descriptor for stderr
+            fname:  file pathname
+
+        Returns:
+            None
+        """
+
+        while True:
+            emsg = ree.read().decode('utf-8')
+            if emsg:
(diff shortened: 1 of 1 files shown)


edreamleo:
Many thanks, Bob, for this improvement.  All tests pass on Windows and Linux.

PR #513, opened 2017-06-30
Sqlite format

vitalije:
Allows Leo to open document from sqlite3 database file.


edreamleo:
It's a bit early for PR's on this project.  Let's get you authorized to create branches.

Issue #562, opened 2017-10-26
Fully working and clean Leo on PyPi.org


maphew:
## Background

PyPi is the [Python Package Index Page](https://pypi.python.org/pypi)

There has been a lengthy [discussion](https://groups.google.com/d/topic/leo-editor/yLCtg54aqEI/discussion,https://groups.google.com/d/topic/leo-editor/) of this issue on leo-editor.

Leo developers *should not use pip* because some files left behind when using pip (see #603). Instead, Leo devs should use git as always.

## Goals  

- Any Leo maintainer can update [Leo's PyPi page](https://pypi.python.org/pypi/leo) with a straightforward workflow. Expertise with setuptools or distutils expertise should not be required.
- People can install Leo with `pip install leo`.
- People may also have to install PyQt with their favorite package manager: `conda install pyqt`.

## What Matt has done

- Cherry picked the old setup.py branch into a new branch, setuppy-clean, uploaded the result to PyPi, and tested the resulting PyPi package on Windows and Linux.
- Merged the new into master with a pull request and deleted the setuppy-clean branch.
- Updated the docs and scripts in `leo/dist/leoDist.leo`.

## Current status

- Work is now being carried out in master.
- PyPi compliant wheel packages are built and install successfully on Windows (7,10) and Linux (Mint) with python 2.7 and 3.6.
- leoDist.leo contains the **package upload recipe** in the node `Matt's workflow for updating pypi`.  At present, a clone of this node appears at the top level.  The clone will remain at the top level until this issue is closed, or perhaps even longer.

@maphew  **I have removed the package upload recipe from this comment**.  I see no reason for the duplication.  It will be easy enough to restore if necessary.

## To do

- [ ] Update installation docs. **In Progress**. @maphew Please put the various documentation sections in leoDist.leo into separate nodes, with subsidiary discussions in child nodes, or even grandchild nodes.  That way devs can see the logical relationships between sections.  In general, **never** rely on markup when outline structure is available.

- [ ] Someone besides Matt runs through process from start to finish and is successful. @maphew Is further "internal" testing really necessary?  Can't we simply test the wheel after you upload it?

edreamleo:
Thanks for your help on this issue.

maphew:
As far as I'm concerned everything of value in the setup.py branch has been retrieved and it can now be deleted. Any reason not to do that now @edreamleo, @tbnorth , @vivainio ?

maphew:
[...]

PR #564, opened 2017-10-27
Setuppy clean

maphew:
see #562.

`MANIFEST.in` changes should be reviewed. They work for pypi and pip, but I don't know if NSIS installer and the like are impacted also.


edreamleo:
As I mentioned in the issue, the line `prune leo/extensions` probably should be deleted.

PR #594, opened 2017-11-26
Functionally ready for mainstream pip install

maphew:
Except for updating docs on how to use pip and the known limitations of a pip install versus git checkout and/or standard installer I believe this is ready for general use.


maphew:
going ahead with merge as there are no changes outside of setup.py and LeoDist.leo.

edreamleo:
@maphew Thanks, Matt, for all this work. All looks well in master.

maphew:
You're welcome! More than once I despaired of reaching the end as one quirk after another reached out to trip me ;-)

Issue #645, opened 2018-01-01
TabImporter fails on files with both tabs and spaces


etihwnad:
The specific case is importing a [vimoutliner](https://github.com/vimoutliner/vimoutliner) file which uses leading tabs to indicate the indent level and has spaces between words on the rest of the line.

```
Heading
<tab>Subheading name
               ^- space
```

The root cause for the failure happens in `TabImporter.import_files()` where the file is read as a single string then sent to `.check(s)`.  Well, `.check(s)` iterates over the string, with the implicit assumption that its argument is iterable by lines.

Later on, in `.scan()` things are handled properly.

PR #646, opened 2018-01-01
Fix TabImporter leading whitespace checking

etihwnad:
Assure that .check() is called with a list of (pre-split) lines.

Fixes leo-editor/leo-editor#645


etihwnad:
It's been awhile, my last Leo commit was from 2010.  Starting to come back up to speed on using it again.

PR #709, opened 2018-02-08
add Mit license to root so repository page will display license

OMlalala:
### Background
- the license of Leo-editor is in leo's home page: [Leo’s MIT license — Leo 5.7 documentation](http://leoeditor.com/license.html)
- but if you include a detectable license in your repository, people who visit your repository will see it at the top of the repository page, ref: [Adding a license to a repository - User Documentation](https://help.github.com/articles/adding-a-license-to-a-repository/)

### Outcome
repo-license-indicator


maphew:
Auto detection of license would be good. Lot's of people use that as a primary filter. Thanks for noticing.

I don' think it's the preamble that's preventing detection. License isn't displayed in your fork either: https://github.com/OMlalala/leo-editor/tree/mit-license

From the docs GitHub doesn't like `.TXT`. Try renaming to `LICENSE` or `LICENSE.md`. If that doesn't do it maybe rename and move the preamble to bottom, after separator bars. (Following principle of not removing other's writing when not necessary)

OMlalala:
@maphew 
Got that. 
Thanks for your sincere suggestion, I will fix that.

Issue #773, opened 2018-03-03
setup.py creates empty `dist-info\DESCRIPTION.rst` 


maphew:
the current code in setup.py for converting Markdown to rST for the pypi package *long_description* variable works, but an upstream bug causes the resultant file to be almost empty. 

I commented on this somewhere before but can't find the reference, buried in a thread somewhere, so creating this issue to help track it.

    from pypandoc import convert
    def read_md(f):
        text = open(f, 'r').read()
        #debug
        print(text)
        print('↑↑↑ Markdown ***** rST ↓↓↓')
        print(convert(text, 'rst', format='md'))
        #/debug
        return convert(text, 'rst', format='md')

### Output (what is written to `dist-info\DESCRIPTION.rst`)
~~~
> more leo-5.7.dist-info\DESCRIPTION.rst
Leo 5.7 final February 27, 2018
~~~

### Expected (what is reported to console)
~~~
Leo 5.7 final February 27, 2018

Leo 5.7 final, http://leoeditor.com, is now available on
`SourceForge <http://sourceforge.net/projects/leo/files/Leo/>`__ and on
`GitHub <https://github.com/leo-editor/leo-editor>`__.

Leo is an IDE, outliner and PIM, as described
`here <http://leoeditor.com/preface.html>`__.

Simulating Leo's features in Vim, Emacs or Eclipse is possible, just as
it is possible to simulate Python in C or assembly language...

**The highlights of Leo 5.7**

-  pip install leo.
-  A debian package for Leo.
-  A web-based Leo Viewer, written by Joe Orr.
-  Added diff-leo-files and diff-and-open-leo-files commands. These
   create outline-oriented diffs for .leo files.
-  New reload-settings command.
-  Leo optionally shows user tips on startup.
-  Added a new Tips page to Leo's documentation.
-  A visual widget for editing cvs tables.
-  Spell checking works without pyenchant
-  Improved the console\_gui plugin.
-  Added live links from urls in the log pane.
-  Unit tests use a string gui, making them faster and more robust.
-  The line\_numbering plugin shows line numbers in external files.
-  Improved the javascript importer.
-  Added a TreePad importer.
-  Added 6 new commands.
-  Dozens of bug fixes.

**Links**

-  Leo's home page: http://leoeditor.com
-  `Documentation <http://leoeditor.com/leo_toc.html>`__
-  `Tutorials <http://leoeditor.com/tutorial.html>`__
[...]

PR #785, opened 2018-03-05
Empty description for pypi fixed (#773)

maphew:


edreamleo:
Thanks for all this work.

PR #825, opened 2018-03-19
Update viewrendered.py

tbpassin:
Added conditional import for urllib.urlopen for Python 2.7 (the urllib module was re-structured for Python 3.x). Ref: https://docs.python.org/2/library/urllib.html.


edreamleo:
@tbpassin Thanks for this.

Issue #672, opened 2018-02-01
Continuous integration with travis-ci


edreamleo:
[Travis CI](https://travis-ci.org/) supports continuous integration testing for open source projects.

This could create debian packages for daily builds, run unit tests, check stuff with lint, and verify that sources are compatible with both Python 2 and 3.

**Notes**

- pytest is part of the Anaconda distro.
- travis-ci requires a working build/test script.  Some possibilities:

1.  Run unitTest.leo from a bridge from, say, leo_unit_tests.py, which pytest would then run.
2. A script to convert all `@test` nodes to separate unit tests. 
   The make_stub_files project used such a script, iirc.

maphew:
I succeeded in getting a minimal travis-ci done, in the branch **travis**. The build passes, but only because it's not testing leo ;-)  By default travis searches the project and all sub-directories and processes any file with `test` in the name. I hard coded it test only one file in `.travis.yml`

Passes:
https://travis-ci.org/leo-editor/leo-editor/builds/420717673
```
script:
    - pytest test_pass.py
```

Fails, but with a deliberate bad test file included, so successful:
https://travis-ci.org/leo-editor/leo-editor/builds/420716344
```
script:
    - pytest test_*.py
```

I've never used a continuous integration tool before. It's really neat. Each build takes about 1 minute to get to the point of being able to run the tests. I can easily see how this will help hugely in verifying and troubleshooting the pip packaging let alone the "real testing".

maphew:
Travis branch now successfully:

- fetches named release tag from github (5.8-b1)
- builds a Leo distribution wheel
- installs the wheel
- tests Leo in console mode

It's essentially the same recipe as _LeoDist.leo: How to... >> ...build PyPi package_
It takes about 2 minutes, which is much faster than my machine. Very nice. I do feel a smidgen of guilt for committing a single character typo change which then results in all that work (!), which is done for us for free.

Log - https://travis-ci.org/leo-editor/leo-editor/builds/421909199

Next up: investigate [automatic deployment to pypi ](https://docs.travis-ci.com/user/deployment/pypi/)

----
Back to the original goal, testing: can Leo run the unit tests, or a portion of them, in headless console mode? Possibly more straightforward than converting them for pytest, assuming it's works / is feasible.

    leo-console --script=./leo/test/cmdline-unittests.leo

edreamleo:
On Wed, Aug 29, 2018 at 1:17 AM matt wilkie <notifications@github.com>
[...]

PR #972, opened 2018-08-30
Enable Travis-CI on master branch

maphew:
Part of #672. Merging this .travis.yml will enable travis-ci testing. 

From that point forward every commit will trigger a travis build and test. Build status and history can be viewed at https://travis-ci.org/leo-editor/leo-editor

If the commit is a tag AND the branch the commit is in is _master_ travis will also attempt to deploy a release automatically to PyPi.org. (Pypi will reject tags that don't look like version numbers, I think; untested)

The `travis_tests` folder is optional at this point because there are no real tests in it. It can be removed, or moved elsewhere. It's there for illustration.


maphew:
PR title says 'merge into master' but finishing the PR actually only pulls into `devel`. Going ahead with that. After this Travis will start automatically testing devel branch.

technatica:
Have you looked at building an appimage?

It would make installing on Linux for non-technical users a snap. Oh,
wait... :-)

Chris

technatica:
https://docs.appimage.org/packaging-guide/other-resources/travis.html

maphew:
Thanks for the idea and pointer @technatica. I've used AppImage before (Krita, Onenote) but not looked at how they're made. After a cursory run through the docs it feels like it's at least as complicated as setup.py and pip. I've barely got my head around that so it'll be some time before I'll be ready to dig deeper. ;-) Feel free to open an enhancement issue for it though. Maybe someone else will jump in.

PR #974, opened 2018-09-10
LeoSetings.leo suggestion

frakel:
adding the '&' again in "Recent Files" in LeoSetings.leo as this was the case in previous releases
just for quicker access of the menu command...

### DIFF
--- a/leo/core/commit_timestamp.json
+++ b/leo/core/commit_timestamp.json
@@ -1,4 +1,4 @@
 {
-    "asctime": "Wed, Aug 15, 2018  9:56:10 AM",
-    "timestamp": "20180815095610"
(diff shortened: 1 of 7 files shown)


frakel:
sorry if I produced confusion here. It says *frakel wants to merge 5 commits into master from devel*  which is not exactly what I wanted. I just added an '&' in the  `@menu Recent &Files` node in `/leo/config/leoSettings.leo` as descibed ...

frakel:
sorry I was confused and made my change in the leo-editor's project instead of in my fork, arghhh

frakel:
The problem first was that I mixed up the clone with the fork, second, devel with master branch. Sorry for the mess I produced :(

PR #976, opened 2018-09-14
Update Qt example in LeoDocs to use leoQt

jcreedcmu:
I'm a comparatively new user of leo so please let me know if there's a developer-contributor's guide I should have found, or if I'm otherwise making any mistakes out of ignorance!

I tried doing a `ctrl-b` on the example
```
from PyQt4 import QtGui
w = QtGui.QWidget()
w.resize(250, 150)
w.move(300, 300)
w.setWindowTitle('Simple test')
w.show()
c.my_test = w # <-- Keep a reference to the window!
```
I found in `leo/doc/LeoDocs.leo` and got the error
```
exception executing script
ImportError: No module named PyQt4

--------------------
   line 3: 
* line 4: from PyQt4 import QtGui
   line 5: w = QtGui.QWidget()
   line 6: w.resize(250, 150)
```

I found commit f5dbba2035 suggested that at some point the set of available modules for import was refactored, and when I tried replacing the first two lines of the example with
```
from leo.core.leoQt import QtWidgets, QtCore
w = QtWidgets.QWidget()
```
executing it as a script successfully created a new Qt Window. Thus, the present PR, made on the assumption that the docs were simply a bit out of date. If there's something else going on that's supposed to make `PyQt4` directly visible to scripts, I'd be interested in knowing.


tbnorth:
I'll check it later but I think your analysis and fix are correct - thanks!

edreamleo:
On Fri, Sep 14, 2018 at 5:15 PM jcreedcmu <notifications@github.com> wrote:

> I'm a comparatively new user of leo so please let me know if there's a
> developer-contributor's guide I should have found, or if I'm otherwise
> making any mistakes out of ignorance!
>
> I tried doing a ctrl-b on the example
>
> from PyQt4 import QtGui
> w = QtGui.QWidget()
> w.resize(250, 150)
> w.move(300, 300)
> w.setWindowTitle('Simple test')
> w.show()
> c.my_test = w # <-- Keep a reference to the window!
>
> I found in leo/doc/LeoDocs.leo and got the error
>
> exception executing script
> ImportError: No module named PyQt4
>
> Thanks for this.  True, this is an example in the FAQ, but as you say,
it's misleading.

Rev aba1c611a applies your patch by hand.  It will appear online shortly.

Edward

edreamleo:
I've applied this patch by hand, so it can be closed.

Issue #968, opened 2018-08-27
make `python setup.py install` work again - from #810


brobr:
_spawned from #810. Summary:_

> For creating a distribution package one could -in versions prior to 5.7- copy the files 'leo/dist/leo-install.py' and 'leo/dist/setup.py' to the main directory and from there run
> `python setup.py install --root=$PKG` with as target the folder in which the package is created ($PKG).

...

Hi, thanks for the update and an extended leo-5.7, which I managed to compile with the leo-5.6/leo/dist/setup.py outwith pip but whether the program is now fully functional time will tell:
> There have a been changes in setup.py since 5.6 not specifically related to pip. 

>  If there are specific reasons you can't, re-open this issue and we can explore what might be done. I'll need information and feedback though as I don't know about building distribution packages.

The request for a setup.py that works via install is that it can compile the program in such a way that it can be packaged (that is 'installed') outwith the python-installation of the computer that does the compilation. With 'python setup.py install' one can direct installation of the compiled program to a folder in /tmp, say '/tmp/ leo-package', which will contain a folder structure mimicking the one of any computer that runs that distribution. Then this leo-package folder -after compressing- can be distributed among users/computers that run this distribution. The install/upgrade-scripts of that distribution will then copy the bits from the package into the python-distribution of that computer. Of course, in order to run the installed program correctly, all necessary dependencies have to be installed.

Thus, users of that distribution can just use the distributed binary or compile the program on their box as if they are an admin using the distribution-specific compile scripts as longs as their box is in a similar state as that of the admin/distributor.

As an administrator of many computers which need to get a python program installed, one would prefer to only use the distribution-specific install/upgrade tools as this provides an easy way to check what has been installed and/or what needs updating on each computer. In such a scenario, when on top of this, as not all is python on those boxes- one needs to check pip for python updates on each and every computer, things can become difficult to manage. Especially if this also would hold for modules for other programming environments that come with their own installers (e.g. ruby) if these could not be compiled into a form that can be distributed. 

The pip-install is excellent for users with one computer to manage, but if this number increases it becomes quite hard to maintain a set of computers if one is forced to, as currently for leo-5.7, to use pip outwith the common distribution-specific routine.

In the end, you run the risk of losing users who rely on distributed versions of leo:
[...]

Issue #810, opened 2018-03-14
"python setup.py install" does not complete


brobr:
Hi,

Thanks for the new release of Leo. With the pip-install, however, Leo-5.7 breaks with the previous possibility to create a distribution-specific package.

For creating a distribution package one could -in versions prior to 5.7- copy the files 'leo/dist/leo-install.py' and 'leo/dist/setup.py'  to the main directory and from there run
` "python setup.py install --root=$PKG"` with as target the folder in which the package is created ($PKG).

There is no longer a 'leo/dist/setup.py'; instead the pip-focused setup.py in the main directory, which, when run by its own -after installing 'semantic-version-2.3.0'- yields this error:

> copying leo/plugins/importers/howto.txt -> build/lib/leo/plugins/importers
running build_scripts
creating build/scripts-2.7
error: file '/tmp/SBo/leo-editor-5.7/leo-install.py' does not exist

When run in combination with the requested leo-install.py, copied over from 'leo/dist/', compilation fails; I got this error:

> .....
byte-compiling /tmp/SBo/package-leo/usr/lib64/python2.7/site-packages/leo/plugins/xemacs.py to xemacs.pyc
byte-compiling /tmp/SBo/package-leo/usr/lib64/python2.7/site-packages/leo/plugins/xml_edit.py to xml_edit.pyc
byte-compiling /tmp/SBo/package-leo/usr/lib64/python2.7/site-packages/leo/plugins/xsltWithNodes.py to xsltWithNodes.pyc
byte-compiling /tmp/SBo/package-leo/usr/lib64/python2.7/site-packages/leo/plugins/zenity_file_dialogs.py to zenity_file_dialogs.pyc
byte-compiling /tmp/SBo/package-leo/usr/lib64/python2.7/site-packages/leo/__init__.py to __init__.pyc
running install_egg_info
Copying leo.egg-info to /tmp/SBo/package-leo/usr/lib64/python2.7/site-packages/leo-5.7_final-py2.7.egg-info
running install_scripts
Traceback (most recent call last):
  File "setup.py", line 150, in <module>
    'gui_scripts': ['leo = leo.core.runLeo:run']
  File "/usr/lib64/python2.7/site-packages/setuptools/__init__.py", line 129, in setup
    return distutils.core.setup(**attrs)
  File "/usr/lib64/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
[...]

PR #983, opened 2018-10-01
Setup 968

maphew:
Branch initiated from #968, #810. Highlights:

* enable Travis continual integration testing
* move dependencies only needed for building packages for PyPi outside of `setup.py`, as per PEP566
* drop pypandoc dependency for package building: no longer necessary because new pypi.org understands markdown
* don't install Windows-only console scripts on Linux (largely cosmetic and might be reverted as it makes wheels built on Linux less useful when installed on Windows. cf. https://github.com/leo-editor/leo-editor/issues/968#issuecomment-425281750)


edreamleo:
@maphew I'm not competent to review this.  I like the TravisCI work you have done.  Feel free to merge anything you do into devel.  I'll leave this open for now.  Please close when you have resolved it to your own satisfaction.

maphew:
ok, all done. yay :)

Issue #987, opened 2018-10-08
Remove Readme conversion from setup.py; Simplify build dependencies


maphew:
* With PEP 566 pypi.org can now read Markdown, so we can rip out all the extra work and associated dependencies (pypandoc) in setup.py to convert from md to restructured text. ([ref](https://pypi.org/project/markdown-description-example/))
* With [PEP 518](https://www.python.org/dev/peps/pep-0518/) we can move the dependencies needed only for package building out of the install for normal users

Discovered while troubleshooting #810. There's no point in waiting for that to get fixed before using these simplifications though.

maphew:
x-ref: https://github.com/pypa/setuptools/issues/1495#issuecomment-423655919

edreamleo:
Thanks, Matt, for handling all of this.

edreamleo:
I am going to close this now that devel passes TravisCI.  Please reopen if necessary.

PR #988, opened 2018-10-08
Simplify 987 

maphew:
implement #987


maphew:
Hmmm. This is kind of noisy. Probably overkill to create both an Issue and a Pull Request for this.

PR #1302, opened 2019-08-25
Bugfix/redo group fix

btheado:
Fixes issue where commands which use undoGroup/redoGroup will fail on redo when the undo stack is not empty. The error message looks like `redoGroup oops: expecting bunch.items.  bunch.kind = XXXX`, where XXXX depends on what undo item is one earlier in the undo stack.

See full discussion at https://groups.google.com/d/msg/leo-editor/IgYYBwvtLf0/jRAAap2qDQAJ


edreamleo:
Thanks for the new unit test, and the clear description of it.

PR #1305, opened 2019-08-29
Recognize Super and Hyper as bare modifier Qt keys

leonidborisenko:
I'm using Leo editor on Linux. When I press just Windows key, following characters are added to text:

Windows key produces text garbage on Linux

---

Due to being designed at the time of the space-cadet keyboard, X11 defined the shift states "Meta", "Super", and "Hyper" (along with "Shift" and "Control" and "Alt" which were commonly available on keyboards). Initially on PC hardware it was just impossible to produce these shift states as there was no key to push for them.

With the appearance of keyboards with the Windows key there was now a key available on standard keyboards that could be used for one of these. Initially, circa 1996, it was most common to make this key be the "Meta" shift key. However, due to the high number of Emacs commands using "Meta", there were already long-established replacements (the Alt key acted as Meta, or typing Escape, X acted like Meta-X), so adding an actual Meta key did not provide much added functionality. This made Super the first key of interest in emulating, and therefore it became the standard assignment after a few
years.

Source: https://en.wikipedia.org/wiki/Super_key_(keyboard_button)

### DIFF
--- a/leo/plugins/qt_events.py
+++ b/leo/plugins/qt_events.py
@@ -271,6 +271,10 @@ class LeoQtEventFilter(QtCore.QObject):
             qt.Key_Shift: 'Key_Shift',
             qt.Key_NumLock: 'Num_Lock',
                 # 868.
+            qt.Key_Super_L: 'Key_Super_L',
+            qt.Key_Super_R: 'Key_Super_R',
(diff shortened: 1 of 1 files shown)


edreamleo:
Many thanks for this.  My only concern is whether the new constants are always defined.  If not, we can add conditional code.

Issue #1312, opened 2019-09-01
"time.clock" needs to be replaced with "time.process_time"


KmolYuan:
`time` module function `clock` will be removed in Python 3.8, its equivalent function is `process_time`. (https://github.com/python/cpython/pull/13270)

edreamleo:
@KmolYuan Thanks for this.  I didn't know about this change.

PR #1313, opened 2019-09-01
Replace "time.clock" into "time.process_time".

KmolYuan:
(#1312) Leo-editor can compatible with CPython 3.8.0b4+ after applied this patch.


KmolYuan:
Does leo-editor 6 also need to be compatible with Python 2? Otherwise a check should be established.

edreamleo:
@KmolYuan Leo 6.0 requires python 3.  That's one of it's most important features.

edreamleo:
In future, please make base all pull requests on devel, not master.  I want to keep master unchanged except for official releases.

PR #1355, opened 2019-09-28
Keep cursor on blank last line after end-of-line command

### LINKED ISSUE #1354: end-of-line command on blank last line moves cursor up a line
With #1353, I'm now using the `end-of-line` command more often and I found if the last line in a node body is blank and the cursor is on that last line, then the `end-of-line` command will move the cursor to the end of the previous line. I expect the cursor to stay where it was as the `end-of-line` command should never cause the cursor to change lines.

I have some failing unit tests and a fix (which I'm not 100% confident about) for this issue. I will submit it in a pull request.

btheado:
Fixes #1354. I'm not 100% sure this is the right fix. It isn't obvious to me whether or not g.getLine has the correct behavior in this case. I chose to fix it in a less obtrusive place which only affects the `end-of-line` command.

I added two unit tests which fail before the fix and pass after. I also added a few other tests to help me feel better about the change.

### DIFF
--- a/leo/commands/editCommands.py
+++ b/leo/commands/editCommands.py
@@ -2028,7 +2028,7 @@ class EditCommandsClass(BaseEditCommandsClass):
                 self.moveToHelper(event, i, extend=extend)
             elif spot == 'end-line':
                 # Bug fix: 2011/11/13: Significant in external tests.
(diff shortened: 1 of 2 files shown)


edreamleo:
Thanks for this work.  It's now merged into devel.  We may as well test there.

Issue #1435, opened 2019-11-08
mod_http plugin error


LewisNeal:
mod_http plugin enabled, open http://localhost:8130/ in a browser. The page does not load.

Console output:
```
error: uncaptured python exception, closing channel <leo.plugins.mod_http.RequestHandler connected 127.0.0.1:60397 at 0x22c87d4de80> (<class 'AttributeError'>:module 'cgi' has no attribute 'parse_qs' [N:\git\leo-editor\leo\plugins\mod_http.py|a_read|1181] [N:\git\leo-editor\leo\plugins\mod_http.py|handle_read_event|1108] [C:\Users\lewis\AppData\Local\Programs\Python\Python38\lib\asyncore.py|handle_read_event|420] [C:\Users\lewis\AppData\Local\Programs\Python\Python38\lib\asynchat.py|handle_read|171] [N:\git\leo-editor\leo\plugins\mod_http.py|found_terminator|1138] [N:\git\leo-editor\leo\plugins\mod_http.py|handle_request_line|1121])

_Leo Log Window
Leo 6.2-b1-devel, devel branch, build a52b6d87b5
2019-11-08 12:46:02 -0600
Python 3.8.0, PyQt version 5.13.2
Windows 10 AMD64 (build 10.0.18362) SP0_
```

technatica:
Works here on Linux.

Chris

Leo 6.2-b1-devel, devel branch, build a52b6d87b5

2019-11-08 12:46:02 -0600

Python 3.6.8, PyQt version 5.12.3

linux

edreamleo:
@LewisNeal Your traceback shows that you are using Python 3.8.  Python 3.8 [removes parse_qs, parse_qsl, and escape from the cgi module.](https://docs.python.org/3/whatsnew/3.8.html#api-and-feature-removals).

They have been deprecated since Python 3.2. They should be imported from the urllib.parse and html modules instead.

LewisNeal:
Edward, Thanks for setting me the challenge :)
I updated my local mod_http.py and the plugin is working again.
Since Leo requires 3.6 or later I would not expect my changes to break the plugin for others..... Ha.

edreamleo:
@LewisNeal Not sure I know what you mean.  Did you change mod_http so it works with 3.8?  If so, please create a new branch for the changes.  I'll merge it into devel.  Or just issue a PR.

LewisNeal:
[...]

PR #1439, opened 2019-11-12
Update mod_http.py plugin for py3.8

LewisNeal:
A fix for #1435 mod_http plugin error


edreamleo:
This branch should be based on devel, not master.

PR #1441, opened 2019-11-13
Update mod_http.py plugin for py3.8

LewisNeal:
Update mod_http.py plugin to work with py3.8

### DIFF
--- a/leo/plugins/mod_http.py
+++ b/leo/plugins/mod_http.py
@@ -218,7 +218,6 @@ which node is selected.
 import leo.core.leoGlobals as g
 import asynchat
 import asyncore
-import cgi
 import json
 import http.server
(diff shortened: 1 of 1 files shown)


LewisNeal:
I'm sure I created this branch based on devel.
The message above I find a bit confusing: LewisNeal wants to merge 1 commit into leo-editor:master
Is that correct?

LewisNeal:
I need to fork repository from devel.

edreamleo:
On Fri, Nov 15, 2019 at 6:29 AM LewisNeal <notifications@github.com> wrote:

> I need to fork repository from devel.
>
That shouldn't be necessary.  Perhaps I mislead you.  It should only be
necessary to create a *branch* from devel.

Edward

PR #1528, opened 2020-03-10
Vr3

tbpassin:
Viewrendered3 in ready for beta testing.  Please note that in addition to the viewrendered3.py file, there is a new directory in plugins, plugins/viewrendered3 which contains two stylesheets.  In addition, are minor changes to qt_gui.py.  They allow VR3 to be used to display doc strings and plugin module help.


edreamleo:
Please resubmit this PR. It must be based on devel, not master.

PR #1530, opened 2020-03-11
Devel - Viewrendered3 beta ready for merge

tbpassin:
This merge affects viewrendered3.py, qt_gui.py, and has a new directory plugins/viewrendered3 that contains two css stylesheets.


edreamleo:
The diffs against qt_gui show that it does not match the version in devel.

PR #1558, opened 2020-04-01
Update viewrendered3.py

tbpassin:
Change db key in layout to avoid possible conflict with viewrendered.  Merge @ekr changes.  Update docstring.


edreamleo:
I have copied vr3.py by hand to devel, to avoid weird merge conflicts.

PR #1588, opened 2020-05-10
github actions experiment

maphew:


maphew:
Believed complete and am about to merge into devel. However we won't know if it actually works until the next release (or to be more precise: then next tag that starts with `v`).

PR #1623, opened 2020-08-02
Fixes on some plugins

XCaminhante:
active_path.py: Opening a binary file caused an exception because the plugin tried to read the filestream directly.
line_numbering.py: The plugin incorrectly numbered the lines when it found a Leo directive.
nodeActions.py: The plugin was registering its double-click handler in an obsolete event. I added a command and fixed the file node recognition.


XCaminhante:
Sorry, I'll pay more attention

edreamleo:
@XCaminhante No problem :-) Thanks for your work.

PR #1624, opened 2020-08-04
Fixes on some plugins

XCaminhante:
active_path.py: Opening a binary file caused an exception because the plugin tried to read the filestream directly.
line_numbering.py: The plugin incorrectly numbered the lines when it found a Leo directive. Also, the numbering was starting at 3 in nodes that were not file descendants.
nodeActions.py: The plugin was registering its double-click handler in an obsolete event. I added a command and fixed the file node recognition.


edreamleo:
@XCaminhante I'm not sure why the TravisCI tests never ran. It's a small matter because only plugins were affected.

I've just merged your code into devel. All unit tests run on my machine, so all seems well. I expect TravisCI to pass on devel.

Thanks again for this contribution.

PR #1633, opened 2020-08-24
Add contract-all-subheads

edreamleo:


edreamleo:
The new code is exactly the reverse of the expand-all-subheads code. However, there are some problems:

1. Neither expand-all-subheads nor contract-all-subheads is undoable.
2. Unlike expand-all-subheads, contract-all-subheads does not contract c.p. I don't know why.  I can change this easily if desired.

Issue #1643, opened 2020-09-05
Make Qt docks easier to use


edreamleo:
Docks are too important a feature to abandon. What I did:

- `--use-docks` always creates global docks.
- `--global-docks` is deprecated. If present, Leo issues a warning and acts as if `--use-docks` were present.
- `--init-docks` also implies `--use-docks`.

**Update**: Leo hangs if the first-named file does not exist.

edreamleo:
All work has been completed. Please reopen if necessary.

edreamleo:
I am reopening this issue. The layout of the first .leo file loaded should apply to new files.

edreamleo:
I am closing this issue. New files now get a reasonable layout.

PR #1647, opened 2020-09-06
Widgets are moveable only when --init-docks is in effect. Improve mesages

edreamleo:
This is the first step for #1643. I am wondering whether --global-docks is necessary.


edreamleo:
Closing this. It diffs against master.

PR #1650, opened 2020-09-07
Devel

tbpassin:
Fix incorrect initialization of state machine for MD processing.


tbpassin:
The only changes here are two lines of VR3.  No other files are intended to change.

PR #1660, opened 2020-09-09
Fix #1570

### LINKED ISSUE #1570: Tab pane disappears when starting with fresh .leo directory
While testing #1568, I used a blank home directory so that all the default settings and plugins are used. I discovered when I run leo that way, the tab pane disappears.

When the leo window is first drawn, I can see the outline pane on the left with the tab pane beneath it. On the right I see the body pane. But then the tab pane disappears. After that the Leo Tips dialog box appears.

Steps to reproduce on linux:
1. mkdir -p ~/tmp/blank
2. HOME=~/tmp/blank python launchLeo.py

```
Leo 6.3-devel, devel branch, build 8390038749
2020-04-11 17:34:16 +0200
Python 3.7.5, PyQt version 5.13.0
linux
```

I used `git bisect` and found 307a4a9432 "Completed #1514 by setting @string initial-split-orientation = horizontal as the default" to be the first bad commit.

edreamleo:
The fix was as indicated in #1570.

### DIFF
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -3432,8 +3432,10 @@ class LoadManager:
         frame.setInitialWindowGeometry()
         frame.deiconify()
         frame.lift()
+        # #1570: Recalculate the ratios.
+        junk, frame.ratio, frame.secondary_ratio = frame.initialRatios()
+        # Resize the _new_ frame.
         frame.resizePanesToRatio(frame.ratio, frame.secondary_ratio)
-            # Resize the _new_ frame.
         if not g.os_path_exists(fn):
             p = c.rootPosition()
             # Create an empty @edit node unless fn is an .leo file.
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -3432,10 +3432,9 @@ class LoadManager:
         frame.setInitialWindowGeometry()
         frame.deiconify()
         frame.lift()
-        # #1570: Recalculate the ratios.
-        junk, frame.ratio, frame.secondary_ratio = frame.initialRatios()
-        # Resize the _new_ frame.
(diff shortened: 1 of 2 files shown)


btheado:
Leo crashes for me with this code when I try the procedure I reported in the original issue #1570

```
  $ HOME=~/tmp/blank python launchLeo.py
  
  setting leoID from os.getenv('USER'): 'btheado'
  Leo 6.3-devel, bug-1570 branch, build e0952aaaa2
  2020-09-09 17:01:29 -0500
  Can not create empty workbook
  
    File "..../src/leo-editor/leo/core/leoApp.py", line 2632, in doPostPluginsInit
      c1 = lm.openEmptyWorkBook()
  
    File "..../src/leo-editor/leo/core/leoApp.py", line 2675, in openEmptyWorkBook
      c = lm.loadLocalFile(fn, gui=g.app.gui, old_c=None)
  
    File "..../src/leo-editor/leo/core/leoApp.py", line 3322, in loadLocalFile
      c = lm.openFileByName(fn, gui, old_c, previousSettings)
  
    File "..../src/leo-editor/leo/core/leoApp.py", line 3367, in openFileByName
      lm.initWrapperLeoFile(c, fn)
  
    File "..../src/leo-editor/leo/core/leoApp.py", line 3436, in initWrapperLeoFile
      junk, frame.ratio, frame.secondary_ratio = frame.initialRatios()
  
  AttributeError: can't set attribute
```

edreamleo:
The message "AttributeError: can't set attribute" is one of Python's most confusing. What it means in this case is that frame.ratio and frame.secondary_ratio are read-only *properties*.

edreamleo:
Brian indicates that the new code works for him.

PR #1666, opened 2020-09-13
New files use layout of the first-loaded .leo file

### LINKED ISSUE #1665: New files should use the layout of the first loaded .leo file
Leo will no longer use a global "empty" key in the cache to access new files.

This change affects *only* app.saveWindowState and app.restoreWindowState.

*Note*: g.app.save/restoreGlobalWindowState methods no longer exist.

edreamleo:
Fixes #1665.


edreamleo:
I am going to close this. #1666 is moot now that Leo will not use Qt docks.

PR #1673, opened 2020-09-15
Remove all dock-related code from Leo

edreamleo:
See #1673.

I am considering merging this PR into devel in about a week.

There is little risk: `git checkout 96f2826` will restore the legacy code.


tbnorth (approved):
Hi :)  Not sure what you're asking me to review, the changes or the idea of removing dock support from Leo.  I have no attachment to the code being removed, and I'm not using or contributing to Leo so I don't think I have any basis to judge whether or not removal makes sense. I had enough trouble getting a proof of concept docking UI going to understand how it could be more trouble than it was worth.  So I approve :-}

PR #1676, opened 2020-09-16
Removed support for docks.

tbpassin:
Docks code is commented out, not removed.  It will be removed after the final decision on no-docks.


edreamleo:
@tbpassin Good catch. I have updated the no-docks branch by hand.

PR #1680, opened 2020-09-18
No docks

tbpassin:
Remove commented out code for docks.  Still at v3.0rc2.

### DIFF
--- a/leo/doc/LeoDocs.leo
+++ b/leo/doc/LeoDocs.leo
@@ -624,7 +624,6 @@
 </v>
 <v t="ekr.20050830115714.26"><vh>Customizing Leo</vh>
 <v t="ekr.20050830115714.29"><vh>How can I add support for a new language?</vh></v>
-<v t="ekr.20200227061507.1"><vh>How can I arrange Qt docks?</vh></v>
 <v t="ekr.20141108064630.18"><vh>How can I create a pane for matplotlib charts?</vh></v>
 <v t="ekr.20190731085747.1"><vh>How can I customize settings for a particular external file?</vh></v>
 <v t="ekr.20060915112109"><vh>How can I enable and disable support for psyco?</vh></v>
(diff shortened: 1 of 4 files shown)


tbpassin:
I think that was an unintended mistake on my part.  I must have accidentally merged into upstream-no-docks instead of origin-no-docks.  I apologise for that.  

In your version, would you please update the version number to 3.0rc2?  It's in the "About" section of the VR3 docstr.

What is the "Easter Egg interface"?

tbpassin:
Edward, I have a bone to pick with you.  You made a large number of changes to my code without discussing it with me. Some of them were cosmetic - e.g., changing the order of constant declarations - but some of them were functional.  The first example I found is this:
```
# my code
asciidoctor_exec = find_exe('asciidoc') or None

# your change (similar changes to two other lines):
asciidoctor_exec = shutil.which('asciidoctor')
```
My code uses a function ``find_exe()`` which I wrote and you removed.  But your code **will not** always find executables which mine will.  I actually wrote it because some of these external imports weren't being found with your version.

I'm happy with using PRs, but **please** it's important (and respectful) to ask for some discussion before making more than a few obvious tweaks to some non-functional change.

Now I'm going to have to go through hundreds of your changes to try to see which other ones might have affected functionality.  Also, I would have appreciated some explanation of your "Easter Egg interface" changes, which I don't know anything about.  It's not that I don't appreciate changes based on your deep knowledge of certain Leo functionality, but you might not appreciate some VR3 knowledge that I have gained the hard way, like ``find_exe()``.

Even in the non-functional case of the order of constant declarations, some of that order was just historical topsy-turvy growth, but some reflects grouping that makes it easier for me to work with, unlike the strictly alphabetical order you imposed.  We could discuss it, and I may have in the end agreed with you, who knows?

edreamleo:
On Fri, Sep 18, 2020 at 10:09 AM tbpassin <notifications@github.com> wrote:

> Edward, I have a bone to pick with you. You made a large number of changes
> to my code without discussing it with me.
>

My apologies. I don't remember making the changes, but that's not an excuse.

Edward

tbpassin:
I have just checked out the current no-docks branch, and I find that VR3 open in a floating window, is not closing when I issue the ``toggle`` command, and it is positioned at the top of the screen with the close and move controls inaccessible so I can't close it or get it out of the way.

I would like to have it open in the right-hand Leo pane as it used to do, and be able to close it when it's opened in a floating window.

edreamleo:
On Fri, Sep 18, 2020 at 9:27 AM tbpassin <notifications@github.com> wrote:

> I think that was an unintended mistake on my part. I must have
> accidentally merged into upstream-no-docks instead of origin-no-docks. I
> apologise for that.
>
In your version, would you please update the version number to 3.0rc2? It's
> in the "About" section of the VR3 docstr.
>
My copy of VR3 in the no-docks branch already has the version as 3.0rc2.
That code should be the same as your code. The only change I made to the
no-dock code in the VR3 plugins was the ViewRenderedProvider3 class.

> What is the "Easter Egg interface"?
>
When you right-click a Leo splitter you get a popup. That's the Easter Egg.
The new ViewRenderedProvider3 code improves this interface when the user
selects the "Open window" menu.

Edward

edreamleo:
@tbpassin Everything works as it should for me. Maybe we are out of sync somehow?

Does temporary moving .leo/db so it isn't the cache help?

tbpassin:
I found a clumsy way to close the floating window. On Windows, hold <ALT> and press TAB.  Keep the <ALT> key down.  You get a display of all your windows.  You can mouse over the floating VR3 window and click the "X" box that appears.  This closes the window.  However, the next time I open VR3, it appears up at the top again, with the title bar and close box out of reach above the top of the screen.

I also saw these error messages on the terminal -
```
NO _leo_viewrendered3
Wrong pane count at level 0, count:2, sizes:3
```

tbpassin:
I found out why the VR3 floating window wasn't closing.  I have the close command linked to the <ALT-0> key.  When the floating window opens, it gets the focus.  When it has focus, it's not passing key events to Leo/VR3.  After you click back in Leo, then the hot key receives the command and VR3 closes.

I don't know how to pass key events from the QT pane back to Leo, but that would apparently be a good thing to do.

And I still want the VR3 pane to open inside Leo by default, not in a floating window.

edreamleo:
@tbpassin It's probably an issue with the session machinery. That's why I suggested clearing the cache.

edreamleo:
@tbpassin I think you meant, press and hold the *alt* key.

edreamleo:
@tbpassin I thought I had vague recollection of shutil.which change, but on another look it doesn't at all seem like something I would have wanted to do.

In future, I'll be sure to create a separate PR for any change to VR3, regardless of the branch involved.

An upcoming change: g.os_path_expandExpression has been deprecated in favor of c.expand_path_expression. This will be a separate PR, affecting all the VR plugins.

tbpassin:
" I think you meant, press and hold the alt key."

Yes, that's what I meant.  Thanks.

tbpassin:
"Does temporary moving .leo/db so it isn't the cache help?"

When I removed .leo/db, the first time I opened VR3, it opened inside Leo as expected.  After I closed it, then each time I opened it again, it opened as a floating window and gave me the error messages

```
NO _leo_viewrendered3
Wrong pane count at level 0, count:2, sizes:3
```

Actually, I only get the ``Wrong pane count`` message the first time it opens as a floating window.  After that I only see the ``NO _leo_viewrendered3`` message.

edreamleo:
@tbpassin Thanks for this report. "No _leo_viewrendered"  indicates that my changes to the provider class aren't right.

This should be a separate issue, with a separate PR.

tbpassin:
"An upcoming change: g.os_path_expandExpression has been deprecated in favor of c.expand_path_expression. This will be a separate PR, affecting all the VR plugins."

Which branch is this in?  The old method will stay around, won't it?  Why is this change being made?

edreamleo:
@tbpassin The deprecation message exists in devel. Possibly even master.

The change was made to simplify the api. The changed code would be in the get_fn function in all VR plugins.

Replace `g.os_path_expandExpression(fn, c=c)` by `c.expand_path_expression(fn)`.

The new code is better: the global function makes no sense without the 'c' kwarg.

tbpassin:
It looks like I have my no-docks branch up-to-date, and that the current commit has your ns_provider and my find_exe, and my grouping of constant declarations. Thank you.  I will make the change to  os_path_expandExpression as you recommend and do a new PR for that,  Then we should be done. pending your getting the splitter/floating pane thing worked out.

tbpassin:
I got fooled by git extensions again and merged the wrong way from my no-docks repo to Leo's.  I undid this action, but it seems to have left a trace in the history.

Issue #1653, opened 2020-09-08
Region Bounds


kghenderson:
I use Ctrl+Shift+RtArrow quite a lot and it's not stopping at the next operator. Interestingly, Ctrl+Shift+LtArrow does.

For example, I have:

thing['select_me'] = 'something'.  

If I put my cursor at the beginning of select_me (inside the quote) and hit Ctrl+Shift+RtArrow, then I expect to stop before the matching end quote and select the word, but instead it skips the operators and goes to the s in 'something'. If I put the cursor at the end of select_me and Ctrl+Shift+LtArrow to go back, then it _does_ stop at the quote as expected.

Perhaps I've misconfigured something, but it's been a little frustrating.

**EKR**: By default the bindings are:
```
forward-word-extend-selection = Ctrl-Shift-RtArrow
back-word-extend-selection    = Ctrl-Shift-LtArrow
```

edreamleo:
@kghenderson The bug-1653 branch contains a proposed fix.  Please try it out and report your results in PR #1684 attempts a fix. Thanks.

edreamleo:
The bug-1653 branch has been merged into devel.

PR #1684, opened 2020-09-20
Bug 1653

edreamleo:
Proposed fix for #1653.

### DIFF
--- a/leo/commands/editCommands.py
+++ b/leo/commands/editCommands.py
@@ -2197,9 +2197,10 @@ class EditCommandsClass(BaseEditCommandsClass):
                     while 0 <= i < n and g.isWordChar(s[i]):
                         i += 1
                 else:
-                    while 0 <= i < n and g.isWordChar(s[i]):
-                        i += 1
+                    #1653. Scan for non-words *first*.
                     while 0 <= i < n and not g.isWordChar(s[i]):
+                        i += 1
+                    while 0 <= i < n and g.isWordChar(s[i]):
                         i += 1
             else:
                 i -= 1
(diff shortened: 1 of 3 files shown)


kghenderson:
Hmm ... unfortunately this still isn't what I was looking for. I use IntelliJ/PyCharm primarily as my other editor, and this doesn't do the same thing, but I also tested this with SublimeText and it _does_ seem to match what it's doing there by default.

In my example from the issue:
`thing['select_me'] = 'something'`

With cursor at the beginning of the line, before the 't' in thing, then Ctrl+Right should first jump to the end of 'thing', but _then_ to before the s in select - rather than after the 'e' in me. I want it to be aware of the operators in the middle.

I understand this is a personalized setting, so perhaps I just need a better understanding of where to look and/or perhaps a brief write-up on how this might be custom configured based on the other navigation function available? 

This is really a big deal for me because it means I need to use the mouse much more frequently than I'd like.
Of course my end goal is to try the more vim-like modes (or better yet kakoune-like) but I'm not ready to make that jump with Leo yet.

I really appreciate you taking a look at this though, thank you.

kghenderson:
I think this is better than it was, so while not quite the end goal, I think it's ok to go ahead and merge it. Thanks.

edreamleo:
On Mon, Sep 21, 2020 at 9:35 AM Kevin Henderson <notifications@github.com>
wrote:

> I think this is better than it was, so while not quite the end goal, I
> think it's ok to go ahead and merge it. Thanks.
>
Will do. Feel free to suggest your own code. Only ec.moveWordHelper is
involved.

Have you tried the "smart" option, that is, the forward-word-smart and
back-word-smart commands? They too call ec.moveWordHelpe.

Edward

Issue #1686, opened 2020-09-23
Improve how free_layout.py inits, saves & restores layouts


edreamleo:
This has become an issue for the VR3 plugin.

- [ ] If necessary, create a way to designate a layout to be used initially.
- [ ] Restore floating windows automatically.

edreamleo:
Thomas and I agree that the VR plugins work pretty well as they are.

PR #1691, opened 2020-09-26
PR for #1686: improve saving/restoring of layouts

edreamleo:
Preliminary for #1686.
- [x] Add c positional arg to ns.load_layout. Allows better traces.
- [x] Better traces for --trace=layouts.
- [x] Simplify flc.ns_do_context, flc.loadLayouts
- [x] Simplify ns.get_layout and ns.get_loadable_layout by eliminating a kwarg.


edreamleo:
The new code is not guaranteed to be precisely equivalent to the old.

However, the new code is simpler than the old, and is worth doing, despite the slight risk.

PR #1695, opened 2020-10-06
No docks

tbpassin:
Also bumped version to 3.0rc4.  This should be the last release candidate.


edreamleo:
Looks good to me.

edreamleo (approved):
Looks good to me.

PR #1705, opened 2020-10-12
Revert to rev 5501dc: Use the default_layout ivar

edreamleo:
This undoes the refactoring of flc.loadLayouts and ns.get_layout.


edreamleo:
I am going to close this. Because of the recent confusion, I did the work in devel.

PR #1724, opened 2020-11-07
add Page Views

xdvrx1:
When we add this service, it will count every hit of this repo. And this will guide us more about the visitors each day and will indicate the total views. For me, this is very helpful both for us and for those will view this repo seeing this page views. If there is the website built from this repo, it can be simply added there too.

Screenshot (1662)


edreamleo:
@xdvrx1 Hi. Thanks for the contribution.

Please change the base of the PR to "devel".  See [this page](https://stackoverflow.com/questions/24159036/how-to-modify-a-pull-request-on-github-to-change-target-branch-to-merge-into)

edreamleo (changes requested):
Please base this PR on "devel", not "master"

edreamleo (approved):

edreamleo:
@xdvrx1 Thanks for your work. Actually, the patch would also be appropriate in master.  I'll make that change myself.

PR #1736, opened 2020-11-16
Support for plantuml 

sprajagopal:
Hi,
I love the concept of this editor. I started hacking around with it recently. I think [plantuml](plantuml.com/) support would greatly benefit the editor. This would bring it  one step closer to org-mode in emacs. 

For this I modified the `viewrendered` plugin to include a `update_plantuml` function. The function uses `os.system` to run the body text through `java plantuml.jar`. The resulting png file is rendered using the text widget (I used the `update_image` as reference).

Behaviour:
1. Create a `@plantuml` node
2. The body should contain the `plantuml` definition
3. Enabling `viewrendered` plugin shows the output file in the render pane
External Dependencies:
1. Java
2. Plantuml.jar in `~/.leo` folder.


edreamleo (approved):
Thanks for this contribution.

edreamleo (changes requested):
I have changed the base to devel.

The patch should not change README.md.

sprajagopal:
Ah got it, I'll keep future PRs base as `devel`. Any other guidelines I should be following?

edreamleo (approved):
Looks good now.

Issue #1751, opened 2020-11-23
Ctrl+PgDwn crashes quickmove plugin


ar-jan:
I use `Ctrl + Page Down` and `Page Up` for navigating between multiple opened Leo files. When this key combination is used with the Nav tab active, Leo immediately crashes instead of navigating to the next/previous tab in the log pane.

> Leo 6.3, master branch, build e46e9315f6
> 2020-11-10 00:33:16 -0600
> Python 3.7.9, PyQt version 5.9.6
> Windows 10 AMD64 (build 10.0.19041) SP0

I was on v 6.2.1 before, where the same thing also happened so it's not a very recent issue.

edreamleo:
@ar-jan Thanks for this report.

I don't see the crash you describe. Please include it here.

ar-jan:
The error is:

```
Leo 6.3, master branch, build e46e9315f6
2020-11-10 00:33:16 -0600
Traceback (most recent call last):
  File "C:\Programs\Leo\leo\plugins\quicksearch.py", line 733, in onSelectItem
    indent = len(cur.text()) - len(str(cur.text()).lstrip())
AttributeError: 'NoneType' object has no attribute 'text'
```

Is that sufficient information or should I use further debug options? (not sure how/which).

edreamleo:
PR #1756 contains the proposed fix.

edreamleo:
@ar-jan The traceback should suffice. The ekr-quicksearch branch contains the proposed fix. Please test the ekr-quicksearch branch and report your experience.

ar-jan:
Yes, that solves the issue for me.

Curious why/when `cur` would not exist after doing a nav search?

edreamleo:
@ar-jan One of life's mysteries (tm).

edreamleo:
The ekr-quicksearch branch has been merged into devel.

ar-jan:
Thanks!

edreamleo:
@ar-jan You're welcome.

PR #1756, opened 2020-11-25
PR for #1751: crash in quicksearch plugin

edreamleo:
See #1751.


edreamleo (commented):
The new code deals with the crash itself. It remains to be seen whether it is a complete fix.

Issue #1591, opened 2020-05-15
Todo icons are not shown for PyQt 5.14


wangzhaohe:
Leo Log Window
Leo 6.3-devel, devel branch, build 4ea4179199
2020-05-10 09:57:52 -0500
Python 3.7.3, **PyQt version 5.14.2**
darwin

<img width="671" alt="image" src="https://user-images.githubusercontent.com/2628013/82049636-3c36eb80-96e9-11ea-9804-5e0be38f7f6d.png">

wangzhaohe:
Task icons display well after `pip install PyQt5==5.12`

<img width="419" alt="image" src="https://user-images.githubusercontent.com/2628013/86549615-b447c200-bf72-11ea-9f32-2e7e16a31bc2.png">


Leo Log Window
Leo 6.2.1 final, master branch, build 49c5e4fa3c
2020-03-30 08:39:10 -0500
Python 3.7.3, PyQt version 5.12.1
darwin

edreamleo:
@wangzhaohe Thanks for the report. I'll close this issue.

gatesphere:
Having this same issue on PyQt 5.15.1.  I'm not sure that downgrading PyQt is a real solution -- clearly some behavior has changed in PyQt.

edreamleo:
On Mon, Nov 9, 2020 at 6:40 PM Jacob Peck <notifications@github.com> wrote:

> Having this same issue on PyQt 5.15.1. I'm not sure that downgrading PyQt
> is a real solution -- clearly some behavior has changed in PyQt.
>
I'm not sure I ever understood this issue, but now I do see that icons are
missing.

I have reopened the issue and assigned it to 6.4.

Edward

edreamleo:
The backlinks plugin does display text properly, in similar buttons. I wonder whether the css is a problem.

edreamleo:
A workaround is use text rather than icons for problematic versions of PyQt. Expect this workaround soon.

edreamleo:
The ekr-1591 branch contains the fix. See PR #1728.

edreamleo:
The ekr-1591 branch has been merged into devel.

PR #1879, opened 2021-04-05
Provide an icon fix for #1591

gatesphere:
After some tinkering, I was able to restore the icons to the Task (todo.py) pane on PyQt versions > 5.12.


edreamleo (approved):
Looks good enough to merge into devel.  We'll deal with any problems with a new issue if necessary.

PR #1893, opened 2021-04-15
Tbp qt6

tbpassin:
Reverted to use of QWebEngineView when using VR3 with Qt5.  This change allows the rendered test to display correctly, including math symbols.  When using Qt6, rendering remains as before - not styled with the specified stylesheets.

Pylint: 10/10.


edreamleo (approved):
I changed the base from master to ekr-qt. With that change, only vr3.py changes.

Issue #1919, opened 2021-05-03
The tab character is visually not respecting the font size or tab width setting


XCaminhante:
The tab character is not appearing on the screen with a width consistent with the character width of the font applied to the edit field.
It is also not respecting the @tabwidth directive. The Leo Editor's character column counter is always counting the width of the tab as 4 columns.

![Screenshot from 2021-05-03 17-10-50](https://user-images.githubusercontent.com/32053103/116908660-b0d4a400-ac32-11eb-9cd5-ce2ed26dfbc4.png)
![Screenshot from 2021-05-03 17-11-10](https://user-images.githubusercontent.com/32053103/116908664-b16d3a80-ac32-11eb-875e-cdf99a9cd1fc.png)
![Screenshot from 2021-05-03 17-11-35](https://user-images.githubusercontent.com/32053103/116908666-b205d100-ac32-11eb-884f-60ff534ae530.png)
![Screenshot from 2021-05-03 17-11-54](https://user-images.githubusercontent.com/32053103/116908668-b29e6780-ac32-11eb-866e-a05215eeead5.png)

XCaminhante:
This Leo Editor instance is running on the last devel commit

edreamleo:
BaseJEditColorizer.configure_hard_tab_width contains the relevant code.

edreamleo:
There appears to be no proper way to do this in Qt. Indeed, we would like to set the tab width using QTextCharFormat, but that class has no such method. Instead, bjc.configure_hard_tab_width configures the tab width on the *entire body text widget*. This will kinda work, provided that the body pane contains a single @language directive.

edreamleo:
I am going to close this issue. Imo, PR #1922 is the best that can be done.

PR #1922, opened 2021-05-05
PR for #1919: width of hard tabs

edreamleo:
See #1919.  As discussed there, no complete fix seems possible.


XCaminhante:
Thank you very much

Issue #1950, opened 2021-05-21
Can't find /Users/swot/.leo/vr3/vr3_config.ini so VR3 cannot execute non-Python code


wangzhaohe:
At commit 949c37330a757b6a126f7fa2decc86cb8df62912

```
git reset --hard 949c37330a757b6a126f7fa2decc86cb8df62912
```

Markdown is not displayed correctly.

markdown code:

```
sudo journalctl -u memcached
```

display:

```
# #log sudo journalctl -u memcached
```

<font color="red">Can't find /Users/swot/.leo/vr3/vr3_config.ini so VR3 cannot execute non-Python code</font>

Leo Log Window
Leo 6.4-devel, devel branch, build 949c37330a
2021-05-19 10:46:09 -0400
Python 3.9.5, PyQt version 5.15.2
darwin
setting leoID from os.getenv('USER'): 'swot'
current dir: /Users/swot/app/leo-editor
load dir: /Users/swot/app/leo-editor/leo/core
global config dir: /Users/swot/app/leo-editor/leo/config
home dir: /Users/swot
reading settings in /Users/swot/app/leo-editor/leo/config/leoSettings.leo
reading settings in /Users/swot/.leo/myLeoSettings.leo
duplicate, (not conflicting) key bindings in myLeoSettings.leo
all    BackSpace do-nothing
text   BackSpace backward-delete-char
duplicate, (not conflicting) key bindings in myLeoSettings.leo
all    Delete do-nothing
text   Delete delete-char
reading settings in /Users/swot/.leo/themes/EKRDark_use.leo
==== found asciidoc3 processor: None
Can't find /Users/swot/.leo/vr3/vr3_config.ini so VR3 cannot execute non-Python code
reading settings in /Users/swot/.leo/workbook_20191222.leo
redefining abbreviation date;; 
from '{|{x=time.asctime()}|}' to '{|{x=time.strftime("%Y/%m/%d")}|}'
redefining abbreviation html;; 
from '<html>\n<head>\n<title><|title|></title>\n<style>\n</style>\n</head>\n<body>\n<|content|>\n</body>\n</html>' to '@first <!DOCTYPE html>\n@language html\n@tabwidth -2\n\n<html>\n  <head>\n    <title><|title|></title>\n    <style>\n    </style>\n  </head>\n  <body>\n    <|content|>\n  </body>\n</html>'
redefining abbreviation ,, 
[...]

PR #1951, opened 2021-05-21
Fix no-markdown-rendering error (#1950).

tbpassin:
Change a g.es() to g.trace().  Remove several g.es().


edreamleo (approved):
If you like, you could also use g.es_print().

Issue #1907, opened 2021-04-25
Can't Save File with "..\" in @path directive


tbpassin:
Somewhere in the very recent past, I found that I can't save a plugin file that has "..\plugins" in the path.  Specifically, I have this construct (these paths are for testing, but my real plugin file isn't saving either:

```
-@path d:\temp\plugins
    -@file ..\plugins\plugin1.py
```

Note that this is on Windows. When I make a change to the file and try to save, I get a message that the file can't be saved because of errors. The directory _d:\temp\plugins_ gets created, but not the file.  I have tried with forward slashes as well with the same result.

One time the file did get created, but then it couldn't be changed.  Except ... erratically changes could be saved.

@files without the "..\" construct save normally.  Also, an @clean ../plugin/xxx file did not get created, but after I created it outside of Leo, changes did get saved **sometimes**.  _Refresh From Disk_ always works so far, so presumably Leo is constructing the path correctly.

I saw error messages like these:
```
initWriteIvars Did not create ../plugins for ../plugins/testplugin.py
```
I'm wondering if there isn't some kind of locking problem.  Maybe after reading an external file, Leo isn't giving up a lock somehow.

Version info:
```
Leo 6.4-devel, devel branch, build 3282f33cee
2021-04-22 15:25:27 -0500
Python 3.8.9, PyQt version 5.15.1
Windows 10 AMD64 (build 10.0.19041) SP0
```
The changeset is this evening's pull.

tbpassin:
I have a fix for this problem.  But I don't know if it will cause problems for some other edge cases.  The problem occurs in atfile.py, starting around line 208.  _g.makeAllNonExistentDirectories()_ gets called twice, the second time with a path expression that doesn't create new directories, and finally an incorrect path is constructed.

To recap the issue, Leo won't save my @file files when they have this path structure:

```
@path <absolute path>\plugins
      @file ..\plugins\<filename>
```
Leo's code base uses this path setup in many places.

In _atfile.py_, around line 208, we use two variables, _defaultDirectory_ and _targetFileName_.  Apparently either one can be missing (presumably None or an empty string).  The logic for handling these cases is a little messed up.  In pseudo code, with my comments:
```
if defaultDirectory:  # presumably might be None
    c.expand_path_expression(defaultDirectory)
    # make all directories in this path as needed

targetFileName = c.expand_path_expression(targetFileName)
# If targetFileName contained "..\", it still contains it
# after the expansion.  That's a problem
[...]

PR #1958, opened 2021-05-27
PR for #1907

edreamleo:
See #1907.

Aha. We can collapse the code by using g.os_path_realpath(g.fullPath(c, root)):

- targetFileName is now g.os_path_realpath(g.fullPath(c, root)).
- Replace defaultDirectory by root_dir = g.os_path_dirname(targetFileName).
- Create directories with g.makeAllNonExistentDirectories(root_dir).

There are some subtle complications arising to support unit tests.

### DIFF
--- a/leo/core/leoAtFile.py
+++ b/leo/core/leoAtFile.py
@@ -158,6 +158,7 @@ class AtFile:
         Compute default values of all write-related ivars.
         Return the finalized name of the output file.
         """
+        NEW = True
         at, c = self, self.c
         assert root
         self.initCommonIvars()
@@ -178,7 +179,10 @@ class AtFile:
             # For at.putBody only.
         at.outputList = []
             # For stream output.
-        at.targetFileName = targetFileName = root.anyAtFileNodeName() or ''  # #1914.
+        if NEW:
+            pass
+        else:
+            at.targetFileName = targetFileName = root.anyAtFileNodeName() or ''  # #1914.
             # For at.writeError only.
         d = at.scanAllDirectives(root)
             # Sets the following ivars:
@@ -202,6 +206,17 @@ class AtFile:
             if hasattr(at.root.v, 'tnodeList'):
                 delattr(at.root.v, 'tnodeList')
             at.root.v._p_changed = True
+        if NEW:
+            at.targetFileName = targetFileName = g.os_path_realpath(g.fullPath(c, root))
+            root_dir = g.os_path_dirname(targetFileName)
+            make_dirs = c and c.config and c.config.create_nonexistent_directories
+            if make_dirs and root_dir:
+                ok = g.makeAllNonExistentDirectories(root_dir)
+                if not ok:
+                    g.error(f"Did not create directory: {root_dir}")
+                    return None
+            return targetFileName
+        ### Legacy
         #
         # Return the finalized file name.
         # #1341 and #1450.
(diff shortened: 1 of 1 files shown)


tbpassin:
I briefly tested this change as I wrote in [#1907](https://github.com/leo-editor/leo-editor/issues/1907) with success where it had previously failed.

Here are some minor review remarks:
1. There is no need for the test for _c_ in 
~~~
make_dirs = c and c.config and c.config.create_nonexistent_directories
~~~
because _c_ is already used earlier in the file.  If _c_ were None, the earlier bit would have failed.  Perhaps a test for _c_ near the start of the file would be useful, since the file can't do anything without a _c_.
2. If _make_dirs_ is false, we don't make any directories.  That's fine, but in that case targetFileName still gets written somewhere.  It's location would probably be unexpected to a user. who may not have realized that directories were not going to be made.  I'm not sure what should happen in this case, but we should think about it.
3. At the end of the file there is a comment `# For at.writeError only.`  It is not clear if that comment refers to the line before, or is a stray comment left over by mistake.  If it refers to the previous line, I'd rather move it to be ahead of that line, of move it into an inline comment.
4. Pylint has some complaints although the overall score is 9.91/10.  Some of them are for too many branches or statements, which we can just disable, but some of them are for imports not at the top level.  Probably all of those could be moved to the top level.  Another complaint is for 
~~~
    if True:  # args:
        at.printError(*args)
~~~
Can we just remove the condition, or do you have some thoughts that it might be wanted in the future?

Otherwise, would you (@ekr) like me to deal with these pylint complaints?

edreamleo:
On Thu, May 27, 2021 at 8:28 AM tbpassin ***@***.***> wrote:

> I briefly tested this change as I wrote in #1907
> <https://github.com/leo-editor/leo-editor/issues/1907> with success where
> it had previously failed.
>
Thanks for all these detailed remarks and testing.

> Here are some minor review remarks:
>
>    1. There is no need for the test for *c* in
>
> make_dirs = c and c.config and c.config.create_nonexistent_directories
>
> Yeah, I noticed the seemingly extra guards.  I'll remove them.

2. If *make_dirs* is false, we don't make any directories. That's fine, but
> in that case targetFileName still gets written somewhere.


Good catch. A test for os.path.exists is needed.

> 4. Pylint has some complaints although the overall score is 9.91/10.
>
Not to worry. I usually run pylint before pushes. I'll take a look at the
complaint you show.

Edward

edreamleo:
@tbpassin Please retest the code. I have changed the code in subtle ways to make all unit tests pass, and for other reasons.  All unit tests pass here, but this is very tricky code.

tbpassin (commented):
This version succeeded in creating the ..\\  directories and the target file, even when Leo was started from my home directory.  I did not test any other functionality, including the self test code.

The fragment
~~~
        if not c and c.config:
            return None
~~~
could possibly be changed to an assert(), unless returning None is needed for unit testing.

tbpassin:
Hmm, I just now noticed an error message when I shut Leo down:
~~~
wrote recent file: C:/Users/tom/.leo/.leoRecentFiles.txt

  File "c:\tom\git\leo-editor\leo\core\leoApp.py", line 62, in on_idle
    callback()

  File "c:\tom\git\leo-editor\leo\core\leoExternalFiles.py", line 150, in on_idle
    assert len(self.unchecked_files) < n

AssertionError

removing callback: <bound method ExternalFilesController.on_idle of <leo.core.leoExternalFiles.ExternalFilesController object at 0x000001CA18A42A90>>
~~~

edreamleo:
@tbpassin

- About the guard on c. I'm happy with the code as it is.
- I'll create a new issue for the failed assert in leoExternalFiles.py.

tbpassin:
D'accord.

PR #1966, opened 2021-06-06
Fix #1965

### LINKED ISSUE #1965: Call LM.createAllImporterData() in leoBridge
See PR #1966. For leoInteg.  The fix is general, in bridge.initLeo.

edreamleo:
See #1965.


boltex (approved):
After looking briefly at the changes you made, I tested it by switching to the ekr-bridge branch - and I removed the call to createAllImporterData I had added in leointeg's server. I then ran leointeg and tried to import so files... and it worked! - so this fix seems good to me! :)

Also many thanks for cleaning up some other lines of code and moving back leoserver.py to the 'core' folder.

edreamleo:
@boltex Glad everything works.

PR #1977, opened 2021-06-13
Viewrendered3 Dark Theme Stylesheets

tbpassin:
Added new solarized dark themed stylesheet for RsT rendering.  VR3 attempts to use the dark stylesheet when Leo's theme is dark. Dark theme can be forced with new setting.  Defaults can be overridden by relative or absolute path to a custom stylesheet.

Version to 3.2.

Note that  only two files are involved with this PR: viewrendered3 and the new stylesheet in the plugins/viewrendered3 directory.


edreamleo (changes requested):
1. The PR must not have conflicts with devel.
2. The PR must not change either leoApp.py or leoAtFile.py.

tbpassin:
Those were unintended changes, probably happening while I was reading them and accidentally changed them. I'll start a new branch with hopefully only my intended changes.

tbpassin:
Let's just cancel this PR, and I'll issue a new one from a branch I just reset to Leo's devel branch.  Then I commited my VR3 changes.

PR #1981, opened 2021-06-15
Tbp freefloat editor

tbpassin:
Provides a free-floating view of a single node.  The view has an editor pane and a ReStructured Text rendering pane.

Pylint: 10.0/10.0.

I pulled from devel before committing and pushing this version.  There should be no other changes except for the one new file, freewin.py.


edreamleo (changes requested):
The PR changes 7 files. Something went wrong.

tbpassin:
Sometimes I think that I will never understand git!  The branch of my repo I used was created based on the devel branch from Leo's repo.  Before I committed this plugin file, I pulled from devel so it would be updated to the Leo repo.  I stashed a change to the VR3 stylesheet, Then I committed this version locally and pushed it up to my own remote on github.  How in the heck can there be other files that were changed?

And more important, how can I avoid this kind of problem in the future???  Do I have to hard reset my branch each time?

edreamleo:
On Wed, Jun 16, 2021 at 7:51 AM tbpassin ***@***.***> wrote:

> Sometimes I think that I will never understand git! The branch of my repo
> I used was created based on the devel branch from Leo's repo. Before I
> committed this plugin file, I pulled from devel so it would be updated to
> the Leo repo. I stashed a change to the VR3 stylesheet, Then I committed
> this version locally and pushed it up to my own remote on github. How in
> the heck can there be other files that were changed?
>

I have no idea.  What I would do is to make a copy of the new file in Leo
and then start a new branch, after pulling from devel.

Restore the new code, commit to the new branch and look at the diffs.  If
all seems well, create a new PR and look at the changed files. There should
be only one or two changes.  If not, close the PR.

Yes, this is probably what you did, or think you did :-)

Edward

tbpassin:
I thought that pulling from upstream/devel would accomplish what you say, but that seems to have been incorrect.  

The only thing about starting a new branch like this for each little change is that you end up with so many branches.  And if you go to delete them, you get warnings that you might lose some changesets which could affect other branches.  I get nervous at that point.

OK, cancel out this PR, and I'll start again.

PR #2002, opened 2021-06-24
Fix pyflakes and pylint complaints

edreamleo:
This PR keeps pyflakes and pylint happy.

**pyflakes changes**

1. In some cases, commenting out one unused assignment caused other assignments to become unused, and I commented out those assignments as well. Eventually, we'll want to delete all unused assignments.

2. In save_file, pyflakes complained that 'e', the exception type, was unused, so I added another print statement.

**pylint changes**

1. pylint seems to have found a real error.  In clone_find_tag, I changed _get_find_settings(c) to self.get_search_settings(c)

2. By default, pylint complains about constructs such as:

```
if x:
    return y
else:
    return z
```
preferring instead:
```
if x:
   return y
return z
```
I have changed one or two spots to use the convention used throughout Leo.

3. In get_body, I removed the if-then-else entirely:

`return self._make_minimal_response({"body": w_v.b or ""})`

Also in get_body, pylint wants all paths to return an explicit value, so the default returns None. Is this correct?

4. In set_selection, pylint complained about a test on type. It's more pythonic to use isinstance.

5. pylint complains if an f-string contains no {expression}, so I removed the f in one spot.

**Other changes**

I used f-strings in a few spots. Again, this follows Leo's typical usage.

### DIFF
--- a/leo/core/leoserver.py
+++ b/leo/core/leoserver.py
@@ -140,7 +140,6 @@ class LeoServer:
                 index: string;
             }[]
         """
-        tag = 'get_buttons'
         d = self._check_button_command('get_buttons')
         buttons = []
         # Some button keys are objects so we have to convert first
@@ -210,7 +209,6 @@ class LeoServer:
         Returns an object with total opened files
         and name of currently last opened & selected document.
         """
-        tag = 'open_files'
         files = param.get('files')  # Optional.
         if files:
             for i_file in files:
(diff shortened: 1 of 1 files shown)


boltex (approved):
Looks good, I'll try to keep those standards as I flesh out more of leoserver going forward. Go ahead and merge this 😄
Thanks again !

boltex:
I'll merge it now because I want to continue work on felix-server2 with those changes added ! Thanks again! :)

PR #2007, opened 2021-06-26
 Add capability to cause host outline to navigate to selected gnx. 

tbpassin:
…Automatically choose correct type of Qt widget for rendering view based on 1) if gnx navigation is desired in the rendered view; and 2) if QWebView is not available (e.g., with PyQt6).  Prevent zoom/unzoom keys from working if rendering widget is not a QWebView.

The new navigation capability is very convenient when Freewin windows are used with the "zettelkasten" system.  The navigation of the host to a new selected node does not change the view in the Freewin window because that view is always locked to the node it had when it was opened.

### DIFF
--- a/leo/plugins/freewin.py
+++ b/leo/plugins/freewin.py
@@ -3,15 +3,16 @@
 """
 #@+<< docstring >>
 #@+node:tom.20210603022210.1: ** << docstring >>
-Freewin - a plugin with a basic editor pane that tracks an outline node.
+Freewin - a plugin with a basic editor pane that tracks an
+outline node.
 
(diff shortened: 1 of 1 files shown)


edreamleo (approved):

edreamleo:
Thanks for this work.

In future, please be sure to run pylint on your code before issuing the PR.

I am about to issue my own PR to correct problems pylint found.

tbpassin:
Darn, forgot again.  I always remember for VR3 ...

edreamleo:
@tbpassin  Next time I'm sure you'll remember :-)

PR #2008, opened 2021-06-26
Fix pylint complaints. Simplify gotoHostGnx

edreamleo:
Thomas, I'll wait for your approval or comments before merging.

pylint complaints about anomalous backslashes are almost always fixed by using r-strings.

The revised gotoHostGnx function eliminates an erroneous pylint warning that p might be undefined. The revised function uses no flags.

### DIFF
--- a/leo/plugins/freewin.py
+++ b/leo/plugins/freewin.py
@@ -1,6 +1,6 @@
 #@+leo-ver=5-thin
 #@+node:tom.20210613135525.1: * @file ../plugins/freewin.py
-"""
+r"""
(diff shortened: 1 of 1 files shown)


tbpassin:
All right, the navigation using the regular expressions still work, so functionally this version is fine.  As to the simplification you introduced, the only reason I had the more complex code was to avoid selecting the target if it was already selected.  I suppose that wouldn't matter, really, and maybe Leo's selection code already checks that.  So OK, I'm fine with merging this version into devel.

PR #2042, opened 2021-07-10
Four new Leo themes:

tbpassin:
- True dark Solarized;
- True light Solarized;
- Light Solarized with calmer background color
- Dark theme with less garish and contrasty colors.

True solarized color themes for Leo, and more: see the thread [Four New Themes For Leo](https://groups.google.com/g/leo-editor/c/6E-C5SyeQUk) in Google Groups for more about these themes, including screen shots.


edreamleo (approved):
Thanks for these themes.

PR #2055, opened 2021-07-13
Felix server2

boltex:
on july 13th 2021 at 1:10 am I've finally covered support for all (or most) of Leo's functionality with the leointeg/leoserver combo!

The **leoserver.py** script is now stable and feels somewhat complete. 

I've also added a line to the **nodetags.py** plugin to make everytthing run smootly without an ui for this plugin.

The old leobridgeserver.py could and should be deleted entirely from devel after this is merged into devel. :)


edreamleo (approved):
I don't understand all the changes, but I agree that it is time to merge into devel.

I'll be in touch if I have questions.

Many, many, thanks for the stupendous amount of great work you have done.

edreamleo:
@boltex Rev 2e2fd17 deletes leobridgeserver.py

PR #2056, opened 2021-07-13
Preserve decluttering styles

ATikhonov2:
A suggested solution to issue#1786: Decluttering styles not preserved

The styles are lost when Leo redraws the outline.
Method 'declutter_node' of class LeoQtTree has a lazy implementation:
it keeps the decluttered header of processed nodes and icons in
the dictionary 'declutter_data'.

If a node's header is in that dictionary, Leo gets the decluttered text
and icons from that dictionary, without repeating the check against
the decluttering rules.

Styles are not saved in 'declutter_data'. It results in resetting the style
of nodes to the default each time the method finds the nodes' header
in the dictionary.

### DIFF
--- a/leo/plugins/qt_tree.py
+++ b/leo/plugins/qt_tree.py
@@ -336,61 +336,99 @@ class LeoQtTree(leoFrame.LeoTree):
         #@+node:ekr.20171122064635.1: *7* declutter_replace
         def declutter_replace(arg, cmd):
             """
-            Execute cmd and return True if cmd is any replace command.
+            Executes cmd if cmd is any replace command and returns
+            pair (commander, s), where 'commander' corresponds
+            to the executed replacement operation, 's' is the substituted string.
+            If cmd is not a replacement command return (None, None)
             """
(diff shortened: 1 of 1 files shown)


edreamleo (approved):
Looks good. Many thanks for this contribution.

ATikhonov2:
Thank you for your trust )

I've found a rare scenario when my update will make Leo close. Should I create a new issue on that error and create a new PR with the fix?

Repro steps - create a node with header = "@clean". Leo will fail to apply decluttering styling rules to the node and will close.

The fix - replace conditional statement at leo\plugins\qt_tree.py(357):
    if s:
        ...

to

    if isinstance(s,str):
        ...

I forgot that bool("") == False.

edreamleo:
On Tue, Jul 13, 2021 at 9:14 AM Gaurami ***@***.***> wrote:

> Thank you for your trust )
>
> I've found a rare scenario when my update will make Leo close. Should I
> create a new issue on that error and create a new PR with the fix?
>
You can just create a new PR. It can reference the original issue if you
like.

Edward

PR #2057, opened 2021-07-13
VR3 now works with either pyqt5 or pyqt6.

tbpassin:


edreamleo:
I am closing this PR. The problems with QAction and QActionGroup were not vr3's responsibility.

The problems have been fixed at rev 6a6b50d in ekr-qt6-2.

PR #2058, opened 2021-07-14
Tbp qt6 2

tbpassin:
Two files changed.  VR3 works with pyqt5, pyqt6.


edreamleo (approved):
We made similar changes. I've merged the two versions, preferring the change suggested by pylint.

PR #2083, opened 2021-07-17
1. Open external editor commands work with spaces in path;

tbpassin:
2. Removed openwith_rclick for external editor because Open With Editor already does same thing more flexibly.
3. Rename "Open In <editor>" to "Open With <editor>".
4. Only show editor filename, not entire path.

Pylint: 10.0/10.

### DIFF
--- a/leo/plugins/contextmenu.py
+++ b/leo/plugins/contextmenu.py
@@ -43,7 +43,6 @@ And call this in your plugin *once*::
 # Original version by Ville M. Vainio.
 
 import os
-import subprocess
 from leo.core import leoGlobals as g
 from leo.core.leoQt import QtCore
 from leo.core.leoGui import LeoKeyEvent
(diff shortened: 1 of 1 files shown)


edreamleo (approved):
Thanks for this. I did something similar, but your code looks better.

There is a merge conflict, but I think I can handle it.

tbpassin:
Thanks.  Funny about the merge conflict.  When I issued the PR, Github said it could be merged automatically.

edreamleo:
@tbpassin I think what happened is that I merged my branch into devel after you created your branch. My apologies for the confusion. I didn't realize that you were going to work on this issue.

tbpassin:
I thought that might be it.  I hadn't planned on working the issue, since I knew you had just been working on the same file and would be more familiar with the ends and outs.  But then you wrote how you didn't want to think about it any more - I heard some level of frustration - and I looked a little more at the code out of curiosity.  So I just went ahead.

BTW, somehow some lines that added quoting outside my new function got left in.  They aren't needed any more since that function does the same thing.  Sometime after 6.4b1 (if I remember to do it), I'll take them out with a PR.

edreamleo:
@tbpassin Thanks for these comments. There is no rush :-)

PR #2089, opened 2021-07-18
Viewrendered3 V 3.42

tbpassin:
Better svg handling.
    Handle missing QWebengineView more gracefully for
        html/xml display (body starts with "<").
    Export now works for html/xml display (body starts with "<").
    @jupyter rendering works.


tbpassin:
Please note that I have made one more little change since the PR was issued.  If possible, please bring in that change too, from tbp-transfer in by repo (Github isn't giving me a PR opportunity even though I can see the changed file in my repo).

edreamleo:
@tbpassin You can check the commits on this page to see which commits made it into this PR. If you pushed the commit you should see it here.

PR #2110, opened 2021-08-01
Tbp freefloat editor

tbpassin:
Render nodes starting with "<" as html/xml markup instead of as RsT.


tbpassin:
This PR is only for one file with only a few changes.  I had fetched/merged from devel just before.   I don't know why GitHub is saying there are 41 commits.  Maybe it's including the commits that brought my branch up to date with devel.

edreamleo (changes requested):
The PR resurrects leobridgeserver.py. Please delete.

Also, there is no need to change leoApp.py. Please revert leoApp.py.

tbpassin:
Why don't I just close this one, update my devel branch, and use that.  All those other changes were not anything I did myself.

PR #2122, opened 2021-08-09
felix-server4

boltex:
Made tk dialog to check for unsaved-leo-files upon closing. Also added '--persist' for server persistance after last disconnect, and '--limit n' connections for multi-user support.


edreamleo (approved):
Many thanks for this excellent work!

PR #2130, opened 2021-08-14
Revise setup.py, w/o importing leo and with new tests

edreamleo:
**Important**: The crucial define_entry_points function is unchanged, except for formatting.

**Summary of changes**

- Always use a **static** version number, defined in setup.py itself.
  Use neither the git version nor leo.core.leoVersion.version.
- Check the version with a regex matching only a limited number of valid version numbers:
  `pattern = re.compile(r'[0-9]+\.[0-9]+(\.[0-9]+)?((b|rc)[0-9]+)?')`
   That is: `N1.N2(.N3)?(bN|rcN)?`
- Besides the standard library, import only setuptools.
  - Don't use semvar library.
  - Import neither leoGlobals.py nor leoVersion.py.
- Add print_exception function and get_readme_contents functions.
- Add is_valid_version, test_is_valid_version, and dump_entry_point functions.
- Set `universal=0` in setup.cfg.  setup.py itself must be run with python 3.
- Use open, not codecs.open. The encoding of README.md is known!

Acknowledgement: I could not have begun this work without Matt's previous work.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -19,6 +19,8 @@ include pyproject.toml
 # same as scrub_datafiles in setup.py, prior to Github commit 9982d17e2470ac8313050b8a0288cd39d7ad4f5b
 prune leo/test
 prune leo/doc/html
+prune leo/core/.mypy_cache
+prune leo/core/mypy_stubs
 
 # Issue #603, https://github.com/leo-editor/leo-editor/issues/603
 include leo/doc/html/conf.py
--- a/leo/dist/leoDist.leo
+++ b/leo/dist/leoDist.leo
@@ -151,6 +151,13 @@
 <v t="ekr.20150425135153.2"><vh>brew install formula</vh></v>
 </v>
 <v t="mhw-nc.20190126223741.1"><vh>Pip</vh>
+<v t="ekr.20210813103123.1"><vh>--- no longer used, from setup.py</vh>
+<v t="maphew.20181010203342.385"><vh>get_version (not used)</vh></v>
+<v t="maphew.20181010203342.386"><vh>git_version (not used)</vh>
+<v t="maphew.20180224170257.1"><vh>clean_git_tag</vh></v>
+</v>
+<v t="maphew.20180224170149.1"><vh>get_semantic_version (not used)</vh></v>
+</v>
 <v t="maphew.20180224170853.1"><vh>@file ../../setup.py</vh></v>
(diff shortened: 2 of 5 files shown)


maphew:
Hi Edward! A fortuitous review request. I've been on holiday and this is first time I've looked at github notification in weeks. 

I've always been a little disquieted by the extensive scaffolding required for reading a version number from within the Leo depths, so I'm happy to see that removed in favour of some hard coded simplicity. 

_Starting my review now..._

maphew:
The `build` module is new to me, thanks for the heads up. It should be added to `pyproject.toml` so it's picked up as a build requirement and auto-installed by Travis et al. And also to setup.py `install_requires`, so it's ready at hand for Leo devs\*.

There are 3 batch files in `./leo/dist` - _build-check, build-exe, build-wheel_ that are orphaned from LeoDist.leo. Build-wheel could be updated to use `python -m build` (or maybe just removed).

---
\* Conventionally install_requires is supposed to be the _minimal_ package set needed to run the application and one is supposed to use different constructs for developers. However after spending many many days trying and failing to acheive the same install simplicity for developers without overloading install_requires I give up. I now personally redefine _minimal_ in the Leo context to be _recommended_, as in: all batteries included.

maphew (approved):

maphew:
I neglected to also say: the revised setup.py work fine on my machine (Python 3.6.10 :: Anaconda, Inc.) , according to _build-check_ and `python -m build` which prior to today had not seen a git pull from the Leo repository since the middle of winter.

edreamleo:
@maphew Many thanks for you helpful comments and code additions!

The [revised packaging guide](https://packaging.python.org/tutorials/packaging-projects/) seems much more comprehensible than before. After reading the guide and setup.py I think (at long last!) that I understand the basics. But I'm still a newbie. For example, I would not have thought to add build.py to the install_requires list.

maphew:
Thank you in turn for asking my opinion. :)  ...there's been a number of changes to the packaging guide since I last paid real attention. I knew setup.py is slowly moving towards deprecation in favour of pyproject.toml but was unaware of setup.cfg's upgraded status.

edreamleo:
@maphew You're welcome.

I am going to try pushing the b2 distro to testpypi, but I don't think there is a need to create a pypi distro for Leo 6.4b2.

Matt, are you still willing to do the pypi release for Leo 6.4 (final)?

PR #2131, opened 2021-08-16
felix-server4

boltex:
MAJOR FIX in leoGlobals.py PLEASE REVISE THIS PULL REQUEST
(Using the minibuffer in leointeg worked with commands such as 'mark' but not with rst3 and ecexute-script among others. This small fix using .get('c') instead of .c makes everything work!)
*BUT NOT SURE IF I DIDNT BREAK ANYTHING IMPORTANT*

### DIFF
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -296,7 +296,10 @@ def new_cmd_decorator(name, ivars):
     def _decorator(func):
 
         def new_cmd_wrapper(event):
-            c = event.get('c')
+            if isinstance(event, dict):
+                c = event.get('c')
+            else:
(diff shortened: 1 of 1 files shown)


boltex:
This fixes https://github.com/boltex/leointeg/issues/189 from https://groups.google.com/g/leo-editor/c/zuAKJ3lozFI/m/6PZ_s2J_AgAJ

edreamleo (approved):
Good catch. The new code,
```python
if isinstance(event, dict):
    c = event.get('c')
else:
    c = event.c
```
can not possibly fail in more cases than the old :-)

Proof: c will be None if the dict does not contain a 'c' key.

Issue #195, opened 2015-06-09
fcol when using @first directive is inaccurate


CandyAngel:
When using the @first directive, fcol and col (in the statusbar) are the same, but this is inaccurate as fcol should be 0 for the first character after "@first " (rather than 7.. or 8, depending how you count it).

edreamleo:
Fixed at rev c8adfe6.  Here is the checkin log:

```
Fixed bug 195: fcol when using @first directive is inaccurate.
1. p.textOffset returns None if no @<file> node is an ancestor.
2. p.textOffset never caches its result.
    In any case, it was quite wrong to cache results in positions.
3. Rewrote p.textOffset to simplify it.
4. QtStatusLineClass.update prints '' for fcol if p.textOffset returns None.
5. Changed unit tests accordingly.
```

edreamleo:
Rev 681169c changes the adjustments.  

p.textOffset no longer adjusts section definitions because @others and section references can appear in the same node.  Everything will be fine in typical cases.  For example:

```python
<< section ref >>
class myClass:
    @others
```
But fcol can be wrong in section definition nodes in contrived examples:

```python
if 1:
    << section ref >>
    @others
```
The fcol in the definition node will start at 0, not 4. I'm not going to fix this nit.

edreamleo:
As of rev aab6818, fcol appears to be accurate in all situations, including section definition nodes.

**Note**: The present code is not so forgiving about matching section references and definitions: they should match exactly, including whitespace and capitalization within << and >>.  Yes, it would be possible to use forgiving matches, but that might impact performance significantly.

PR #2141, opened 2021-08-18
Fix for server warning outputs and multi-user support.

boltex:
Fixes https://github.com/boltex/leointeg/issues/194 and also fixes connection refusal when user limit is hit and connection is refused. (it corrupted the user count to -1 of its real value now fixed)


edreamleo (approved):
Looks good. A print statement may be missing a `flush=True`, but I can fix that directly in devel.

edreamleo:
@boltex The fix for the connection limit may also fix #195 that I have just now gotten around to reporting.

PR #2142, opened 2021-08-18
Add flush=True as needed

edreamleo:
Hmm. I'm not sure any of these flush=True args are needed, but they can't hurt...


tbpassin:
But -

```
  File "c:\Tom\git\leo-editor\leo\core\runLeo.py", line 16, in leo_excepthook
    print('\nUncaught exception in Leo...\n', flush=True)

AttributeError: 'PyInterp' object has no attribute 'flush'
```

(Trying to run code in the Python Console panel).

edreamleo:
@tbpassin Good catch!

Issue #2143, opened 2021-08-19
Git-Diff broken for some cases


boltex:
See PR #2146.

Git-diff seems unreliable for me on various machines and OS complaining about 'no .git directory found in <directory>...'

Git-diff needs its cwd set properly in order to perform it's git operations. 
But strangely, (Screenshot below) the top window bar seems to indicate that Leo knows exactly what folder it's supposed to look for the .git folder because it's showing me exactly what branch I'm on... (when this is fixed I'll remove the work-around in leoInteg for 1,0.)

![Annotated-Screenshot from 2021-08-15 15-45-26](https://user-images.githubusercontent.com/847716/129993647-df44580c-ff33-4b4d-9de3-db93bf52874d.png)

Issue #2145, opened 2021-08-19
Add trace-related command-line args to leoserver.py


edreamleo:
- [x] Use argparse rather than optparse. argparse is much better in the long run.
- [x] Add --trace=requests, responses, verbose.
- [x] Support the traces! Imo, the traces look good regardless of the "verbose" argument.
- [x] Move unit tests into leo/unittests/core/test_leoserver.py. Simplifies the argument logic in leoserver.py!

PR #2146, opened 2021-08-19
PR for #2143

edreamleo:
See #2143.
- [x] Remove "directory" kwarg from all relevant methods & functions.
- [x] get_directory *only* uses c.fileName() to compute a base directory.

### DIFF
--- a/leo/commands/editFileCommands.py
+++ b/leo/commands/editFileCommands.py
@@ -724,13 +724,12 @@ class GitDiffController:
             self.file_node.b = (
                 f"{self.file_node.b.rstrip()}\n"
                 f"@language {c2.target_language}\n")
-    #@+node:ekr.20201208115447.1: *4* gdc.diff_pull_request (test)
+    #@+node:ekr.20201208115447.1: *4* gdc.diff_pull_request
     def diff_pull_request(self, base_branch_name='devel', directory=None):
         """
         Create a Leonine version of the diffs that would be
         produced by a pull request between two branches.
         """
-        g.trace(base_branch_name)  ###
         directory = self.get_directory(directory=directory, filename=None)
         if not directory:
             return
@@ -745,7 +744,7 @@ class GitDiffController:
             )
         else:
             g.es_print('FAIL: git rev-parse devel')
-    #@+node:ekr.20180506064102.10: *4* gdc.diff_two_branches (test)
+    #@+node:ekr.20180506064102.10: *4* gdc.diff_two_branches
     def diff_two_branches(self, branch1, branch2, fn, directory=None):
         """Create an outline describing the git diffs for fn."""
         c = self.c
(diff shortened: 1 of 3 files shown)


boltex:
all right - I see you've committed a couple times- imma switch to this branch and try it on my linux and other oses for a little while to see if it works :)

boltex:
doesn't seem to work, a single 'git diff HEAD' gets added to the outline, the leo log pane shows no messages  - but the server printed out :

```
server: User Connected, Total: 1, Limit: 1
git-diff: outline has no directory. filename: 'index.html'
git-diff: no .git directory: '' filename: 'index.html'
git-diff: outline has no directory. filename: 'style.css'
git-diff: no .git directory: '' filename: 'style.css'
git-diff: outline has no directory. filename: 'user-page.leo'
git-diff: no .git directory: '' filename: 'user-page.leo'
```

Screenshot from 2021-08-19 23-06-29

edreamleo:
@boltex The code is *cleaner* than before, but that's all I can say for sure.

The get_directory method contains the essence of the changes. As you can see, this method looks for the .git folder based on the given filename, or c.fileName() if no filename is given.

It's late now and I'm going to be. We may have to work together on this issue.

boltex:
I compared with before, to make sure i was on the ekr-git-diff branch and had some behavior changes, indeed, before i didnt get the 'git diff head' node alone, just a no **".git directory found in '/home/felix'**" message. 

So yeah it's improved, but it looks like its finding a directory, and starting something, but its not using the directory it found when it loops for the git operations. (is see you removed the directory parameters for some git diff methods, were they not useful?) or maybe is the 'directory' global being overwritten with an empty string or something after its found, and the git operations loop starts going with an empty or corrupted directory string? 

Just some ideas... as a start maybe I'll put so g.es here and there to make some more experiments with it tomorrow, i'm going to bed too :)

edreamleo:
@boltex Thanks for these comments.

I agree that the crux of this issue is correctly computing a "starting" directory. The starting directory must have a .git subdirectory to be useful.

Several cff's in leoPy.leo showed that the "directory" kwargs were never being used anywhere. Removing those kwargs might have been hasty, but doing so clarifies the code so I'll restore the "directory" kwargs only if absolutely necessary.

Thanks for reminding me that Leo's signon logic somehow does better than the diff commands. That may be an important clue.

As a sideways move, I'm going to work on #2145 today. Running leoserver.py with --trace=git may be useful.

edreamleo:
**Why Leo signs on properly**

LeoApp.computeSignon contains this code:
```
branch, junk_commit = g.gitInfo()
author, commit, date = g.getGitVersion()
```
Without a "path" kwarg g.gitInfo uses `__file__` as the starting point for the directory search. Using `__file__` this way *is* valid for Leo itself, but not in any other context.

In short, Leo's signon code is the easy case. Sigh.

edreamleo:
**Update**: at present git-diff is failing in the ekr-git-diff branch just as it is for Félix. This is *good* news! Mostly :-)

edreamleo:
@boltex Now get_directory *only* uses c.fileName() to compute a base directory. If that doesn't work, we'll probably have to rely on the cwd or a system setting.

Please let me know how the new code works for you.

boltex (approved):
I don't understand completely everything, but it looks clean, and most importantly I tried it with your newest commits : Success! It works!

PR #2152, opened 2021-08-22
Fixed arguments handling and help-string(s) output

boltex:
Fixed arguments that were gathered, but not (re)assigned properly - I tried all arguments to make sure they work.
Also revised documentation and description strings.

### DIFF
--- a/leo/core/leoserver.py
+++ b/leo/core/leoserver.py
@@ -3083,7 +3083,7 @@ class LeoServer:
             pass
         else:
             param = {}
-            
+
         # Handle traces.
         if trace and verbose:  # pragma: no cover
             g.printObj(d, tag=f"request {id_}")
@@ -3352,7 +3352,7 @@ class LeoServer:
             # - All the *cheap* redraw data for p.
             redraw_d = self._get_position_d(p)
             package ["node"] = redraw_d
-        
+
         # Handle traces.
         if trace and verbose:  # pragma: no cover
             g.printObj(package, tag=f"response {self.current_id}")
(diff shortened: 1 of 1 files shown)


boltex:
@edreamleo DONT MERGE YET  problems with this new arguments getter that does not support arguments with double dash sent from a process spawn in nodejs! (but works if copy pasted on command line as-is!)

I reviewed the code 5 times now - everything matches and i cannot find why this happens!! (might revert to the old arguments system getter and remove the specific sub parameters of the 'trace' argument.)

boltex:
argparse.ArgumentParser totally breaks leo integration :/

boltex:
unrecognized arguments: --limit 2 --port 32125

But if vscode uses the single-dash ones it works :  -l 2 -p 32125

the double dashed arguments that still work with argparse.ArgumentParser are the boolean ones

boltex:
Ok I found a solution in leoInteg : I have to pass the arguments parameters as separate 'arguments'. Meaning that Instead of this (where w_args is an array for argument strings):
   **w_args.push("--limit " + this._leoIntegration.config.limitUsers);**
to set an argument as **--limit 2**, with this new cli arg parsing librairy I have to do:
    **w_args.push("--limit");
    w_args.push(this._leoIntegration.config.limitUsers.toString());**

to then, finally, spawn the server process with the usual:
**this._serverProcess = child.spawn(w_pythonPath, w_args, w_options);**
Is this expected/normal?

boltex:
I looked online at many examples and indeed, the argument parameters are usually passed as separate entries in the arguments array when calling 'spawn process' in nodejs.

So this pull request is good - except it will break the --limit x parameter with the current 0.1.7 beta of leointeg. So I'll release a 0.1.18 beta tomorrow (probably the last beta version until release)

edreamleo (approved):
Thanks for all these careful changes. They look good to me.

edreamleo:
On Sun, Aug 22, 2021 at 7:39 PM Felix ***@***.***> wrote:

> So this pull request is good - except it will break the --limit x
> parameter with the current 0.1.7 beta of leointeg. So I'll release a 0.1.18
> beta tomorrow (probably the last beta version until release)
>

Just merged the PR. As usual, I've left the felix-server4 branch.

Many thanks for all this work!

Edward

PR #2153, opened 2021-08-23
Fix check-settings complaints in leoSettings.leo. Fix misspelling in VR3

edreamleo:
Fix problems reported by `@button check-settings' in leoSettings.leo.

Also removed top-level clone in leoDocs.leo.

@tbpassin There was an apparent misspelling in VR3.


tbpassin:
This change to VR3 seems correct.  The code wanted _vr3-ext-editor_, and the docstring called it _vr3-external-editor_.  All of the usages should agree with the code, as is done in this PR.

PR #2155, opened 2021-08-25
Felix server4

boltex:
Now call c.recreateGnxDict() when finishing opening a Leo file 
(fixes opening leoDocs.leo and outlines with @auto-md nodes)


edreamleo (approved):
The new call to c.recreateGnxDict() seems reasonable.

I'll review the `@persistance` logic to see whether Leo itself should call  c.recreateGnxDict() in more places.

Btw, "persistence" is misspelled, but it's too late to change that now :-)

PR #2157, opened 2021-08-25
Add Qt dialog to server

edreamleo:
I alphabetized helper functions, which increases the diffs here. Use Leo's git-diff command to see the real changes. To summarize the changes:

- Add general_yes_no_dialog and qt_runAskYesNoCancelDialog function.
   The qt_runAskYesNoCancelDialog creates a QApplication object and adds Leo's icon.
- The qt/tk_runAskYesNoCancelDialog functions now take a single "c" argument.
   They "inherit" the "message" var from the enclosing general_yes_no_dialog function.

I have not found a way to yank the focus away from the server console window to the dialog, with either Tk or Qt. So the user will have to click the dialog.

The code should be completely safe. The general_yes_no_dialog raises the Tk dialog if the Qt dialog raises any exception.

### DIFF
--- a/leo/core/leoserver.py
+++ b/leo/core/leoserver.py
@@ -3445,6 +3445,288 @@ def main():  # pragma: no cover (tested in client)
     global wsHost, wsPort, wsLimit, wsPersist, wsSkipDirty, argFile
 
     #@+others
+    #@+node:felix.20210807214524.1: *3* function: cancel_tasks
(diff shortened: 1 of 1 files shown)


boltex:
@edreamleo I've had time to play a bit with it now: seems to work great in both windows and debian linux/ubuntu

edreamleo:
@boltex Thanks for your testing!

PR #2166, opened 2021-09-07
Tbp vr3

tbpassin:
…de redirect symbols. Update version to 3.44.  Add text about the new redirect symbols to docstring.

This change came from wanting to run scripts with MySQL by redirecting them, like this:

`mysql -u user -p < script.sql`

With this change, an sql script can be run by setting up the script's node like this:

```
@language sql
@param -u user -p <

select .....
```

Of course, the mysql processor must have been specified in the ini file for VR3 external processors.


edreamleo:
Please base this PR on "devel", not master. You can do that by selecting the "Edit" button at the top of this page.

edreamleo (approved):

PR #2178, opened 2021-09-14
Current Line Highlighting - Fix behavior When Color Setting is changed...

tbpassin:
... or removed, clarify valid css colors in Help-For menu command, remove quotes and the string "none" if present in the color setting.


tbpassin:
I'm going to close this and submit a new one based on the current state of devel.  Apparently I used a branch updated to ekr-unit-test instead of to devel.

PR #2179, opened 2021-09-14
Highlight Current Line: Fix behavior when highlight color setting is changed ...

tbpassin:
... or removed.  Clarified help-for menu text.  Changed color to lightgrey and deactivated the line-highlight-color setting in leoSettings.py.  Added explanation of valid highlight color values to body of setting node.

This branch is based on devel, so should merge OK.


edreamleo:
Something is still wrong. I suggest copying your changes in a .leo file, then creating a new branch. I'm not going to look at 182 diffs :-)

Issue #2171, opened 2021-09-13
leoAst.py crash


edreamleo:
Beautifying nodediff.py crashed leoAst.py. Here is the source code that caused the crash:
```
   if mode == 'git':
        cmd = [
            'git',
            '--work-tree=%s' % path,
            'show',
            'HEAD:%s' % g.os_path_join( *(relative_path + [filename]) ),  # <-----
        ]
```
Here is the traceback:
```
beautify-files: nodediff.py
Traceback (most recent call last):
  File "c:\leo.repo\leo-editor\leo\core\leoKeys.py", line 2457, in callAltXFunction
    func(event)
  File "c:\leo.repo\leo-editor\leo\core\leoBeautify.py", line 74, in orange_files
    changed = leoAst.Orange(settings=settings).beautify_file(filename)
  File "c:\leo.repo\leo-editor\leo\core\leoAst.py", line 2667, in beautify_file
    contents, encoding, tokens, tree = tog.init_from_file(filename)
  File "c:\leo.repo\leo-editor\leo\core\leoAst.py", line 1242, in init_from_file
    list(self.create_links(tokens, tree))
  File "c:\leo.repo\leo-editor\leo\core\leoAst.py", line 1221, in create_links
    next(self.visitor(tree))
  File "c:\leo.repo\leo-editor\leo\core\leoAst.py", line 1474, in visitor
    yield from method(node)
  File "c:\leo.repo\leo-editor\leo\core\leoAst.py", line 1656, in do_Module
    yield from self.gen(node.body)
[snip]
  File "c:\leo.repo\leo-editor\leo\core\leoAst.py", line 1983, in do_BinOp
    yield from self.gen(node.right)
  File "c:\leo.repo\leo-editor\leo\core\leoAst.py", line 1312, in gen
    yield from self.visitor(z)
  File "c:\leo.repo\leo-editor\leo\core\leoAst.py", line 1474, in visitor
    yield from method(node)
  File "c:\leo.repo\leo-editor\leo\core\leoAst.py", line 2133, in do_Call
    yield from self.handle_call_arguments(node)
  File "c:\leo.repo\leo-editor\leo\core\leoAst.py", line 2197, in handle_call_arguments
    raise AttributeError(f"Invalid * expression: {ast.dump(z)}")  # pragma: no cover
AttributeError: Invalid * expression: Starred(value=BinOp(
    left=Name(id='relative_path', ctx=Load()),
    op=Add(),
[...]

PR #2181, opened 2021-09-14
PR for #2171

edreamleo:
See #2171.


edreamleo:
A new unit test passes. More work may be needed.

Issue #2248, opened 2021-10-05
Flake8 not imported by flake8 Command


tbpassin:
Trying to run flake8 on a node gave a message:

`can not import flake8`

I have not tried to track this down in the code as yet.

tbpassin:
I put the "Bug" label back on because even when flake8 is installed, the imports for the flake8 command don't work . Perhaps flake8 has changed since the command was written.  Here's the import error I get in a python interactive session:

```
>>> from flake8 import engine, main
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'engine' from 'flake8' (C:\Program Files\Python39\lib\site-packages\flake8\__init__.py)
```

tbpassin:
I think it's better to use _--extend-ignore_ instead of _--ignore_.  That way, if there is another _ignore_ that some gets into the command line (happens to me), or there is an ignore list somewhere else, your extra ignores get honored.  Otherwise they may not.

tbpassin:
You could also add flake8 to the File/Open With submenu.

tbpassin:
I had a little internal conversation about that before I tried it! But I already have _Open With Notepad++_, and that doesn't open anything in Leo either.  I suppose I could put it somewhere else.  Maybe a new _Check File_ submenu.

PR #2250, opened 2021-10-07
PR for #2248: rewrite flake8

edreamleo:
See #2248.

- [x] Rename flake8 command to flake8-files command, by analogy to the blacken-files command.
- [x] Rewrite flake8 command, similar to the blacken-files command.
- [x] Remove Flake8Command class.


tbpassin:
The new _flake8-files_ command does not use an ignore list ...

edreamleo:
@tbpassin I'm not wild about this idea, for several reasons.

1. Some people might not want all the "default" suppressions.
2. It would be more flexible, and simpler, to use flake8's config file to make flake8 work exactly as wanted.

PR #2255, opened 2021-10-10
Prevent Openwith menus from opening the same file for invocations aft…

tbpassin:
…er the first.


tbpassin:
I confirm that Openwith now behaves as expected.

edreamleo:
@tbpassin Thanks for the confirmation!

PR #2329, opened 2021-11-13
Added new section to docstring with explanation about how to 

tbpassin:
create the _plot-2d-clipboard_ command and how to bind it to a key, button, or menu item.


edreamleo (approved):

tbpassin:
With this being assigned to me,  since it has already been merged, is there something else I should do?

PR #2332, opened 2021-11-15
Felix-server4

boltex:
Added new methods to set/get ua's and body states now provide wrap and tabwidth to caller. (for next leoInteg release 1.0.4 features)
Ua's also optionaly sent back as boolean flag (with new  'uAsBoolean' flag).

### DIFF
--- a/leo/core/leoserver.py
+++ b/leo/core/leoserver.py
@@ -1236,6 +1236,9 @@ class LeoServer:
         )
         # Get the body wrap state
         wrap = g.scanAllAtWrapDirectives(c, p)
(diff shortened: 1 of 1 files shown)


edreamleo (changes requested):
The line: `if not type(tabWidth) == int:`

should be: `if not isinstance(tabWidth, int):`

In general, it's more pythonic to avoid type.

boltex:
Fixed :)

edreamleo:
@boltex Thanks.  It was a bit pedantic of me to ask for the fix, but I thought it would highlight the comment.

edreamleo (approved):

PR #2348, opened 2021-11-27
Fix two potential ReDoS issues

kevinbackhouse:
These two potential ReDoS issues were found by a [CodeQL query](https://lgtm.com/query/6017926600791434440/), written by my colleague @RasmusWL. We do not believe that these issues have any security impact in leo-editor, so I am posting these suggested fixes as a public pull request.

My fix for `xml_ws_pattern` is based on [this answer on stackoverflow](https://stackoverflow.com/a/1324838), which says that XML comments are not permitted to contain the string `--`.

In `leocursor.py`, I have modified the regex in a way which I believe fixes the potential ReDoS issue without changing its behavior.


edreamleo (approved):
Thank for this PR. The xml.py importer probably should get a new unit test :-)

PR #2349, opened 2021-11-27
PR: improve unit test framework

edreamleo:
Preliminary work before modifying python importers:

- [x] ic.createOutline returns None if perfect import test fails.
- [x] Use unittest.TestCase.id() instead of the string "file" for comparisons of the root node.
- [x] Remove calls to textwrap.dedent from all unit tests. run_test automatically adjusts and dedents the text.
- [x] BaseTestImporter.run_test returns the root of the generated nodes.

**Status**: All unit tests, including import tests, pass with *legacy* python importer code.


edreamleo:
The relevant code has been merged into the ekr-python-importer branch.

Issue #2327, opened 2021-11-11
Improve python importer


edreamleo:
The python importer needs a rewrite. See [this post](https://groups.google.com/g/leo-editor/c/_oYazfNO7EU/m/012FDvB6AwAJ).

Vitalije has provided the rewrite.  See PR [#2353.](https://github.com/leo-editor/leo-editor/pull/2353).

PR #2353, opened 2021-12-07
PR for #2327: Improve python importer

vitalije:
See #2327.

### DIFF
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -1444,47 +1444,14 @@ class LeoApp:
         d = g.app.atAutoDict
         for key in d:
             # pylint: disable=cell-var-from-loop
-            aClass = d.get(key)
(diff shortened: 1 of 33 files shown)


vitalije:
New Python importer is fully functional and 100% covered with unit tests. It is possible that we would encounter on some corner cases where it would be necessary to improve this importer. When we find such cases they can be used to add more tests.

I've made also changes in the way LeoApp uses importer modules. Now, g.app would collect and keep importer functions rather than importer classes. Anyway LeoApp used to turn importer classes in functions before use. Now, this is done once during loading application. In all importers, exported dictionaries have changed to export functions instead of classes.

I believe this schema is cleaner and importers are now better decoupled from LeoApp and vice versa.

edreamleo (approved):
Many thanks for this superb work.

PR #2356, opened 2021-12-09
PR for experiments reverting to EKR's importer

edreamleo:
This is likely not going anywhere, but we'll see.


edreamleo:
I am going to close this PR. Vitalije's work will be the basis of the new python importer.

PR #2361, opened 2021-12-23
PR for importer experiments

edreamleo:
Imo, the new python importer should be integrated into the existing importer framework.


edreamleo:
I am going to close this PR. There is no urgent need for changes.

Issue #2380, opened 2022-01-16
Contract-all de-hoists if the 'hoisted root' node is selected


boltex:
While trying to fix contract-all that de-hoists from any node in leointeg, I discovered that it does so in Leo but only if the 'hoisted root' node is selected. I wonder if those behaviors are related... (and if the behavior is intended? i guess not...?)

boltex:
I would suggest that doing contract-all while hoisted, should select the 'hoisted root node', and perform a 'contract-self-and-descendants'. 

Leaving the user in front of it's single, contracted, hoisted node. 

(leaving the invisible 'outside-of-hoist' tree in its original state.

PR #2384, opened 2022-01-19
PR for #2380

edreamleo:
See #2380.

### DIFF
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -3299,19 +3299,21 @@ class Commands:
         return ''
     #@+node:ekr.20031218072017.2909: *4* c.Expand/contract
     #@+node:ekr.20171124091426.1: *5* c.contractAllHeadlines
-    def contractAllHeadlines(self, event=None, redrawFlag=True):
+    def contractAllHeadlines(self):
         """Contract all nodes in the outline."""
         c = self
-        for p in c.all_positions():
-            p.contract()
-        # Select the topmost ancestor of the presently selected node.
-        p = c.p
-        while p and p.hasParent():
-            p.moveToParent()
-        if redrawFlag:
-            # Do a *full* redraw.
-            # c.redraw_after_contract(p) only contracts a single position.
-            c.redraw(p)
+        for v in c.all_nodes():
+            v.contract()
+        if c.hoistStack:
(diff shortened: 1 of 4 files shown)


leo/core/leoCommands.py
```diff
@@ -3299,19 +3299,21 @@ def navHelper(self, p, ch, extend):
         return ''
     #@+node:ekr.20031218072017.2909: *4* c.Expand/contract
     #@+node:ekr.20171124091426.1: *5* c.contractAllHeadlines
-    def contractAllHeadlines(self, event=None, redrawFlag=True):
```
boltex, line 3302: Not allowing an event argument breaks leoserver (see my other comment)

boltex:
Code looks clean and it works in Leo, but now leoserver complains that:  **_do_leo_command Recovered from Error contractAllHeadlines() got an unexpected keyword argument 'event' when issuing a contract all command.** 

That is because of a line you edited in leoserver, instead of calling it like so:
**value = func()** 
you added an 'event' parameter. (for some reason that im not good enough in python to understand... hehe)
**value = func(event={"c":c})**

boltex:
I pushed a tiny addition to this branch :) 

I brought back the event = none parameter in the contractAllHeadlines method declaration. 

Now that branch magically fixes all of Leo, Leoserver and Leointeg! (well all the hoist related issues, hehe!)
--
Félix

edreamleo (commented):
Removing the event kwarg was a blunder on my part.  Thanks for catching it.

edreamleo:
On Wed, Jan 19, 2022 at 8:50 PM Felix ***@***.***> wrote:

> I pushed a tiny addition to this branch :)
>

Thanks.

> Now that branch magically fixes all of Leo, Leoserver and Leointeg! (well
> all the hoist related issues, hehe!)


:-) I have just pushed some more changes to the PR.

1. c.contractAllHeadlines no longer calls c.redraw. contractAllHeadlines
used to have an evil 'redraw' kwarg. Now, the callers get to choose whether
to redraw or not.  More code, but clearer and more flexible.

2. I removed several redundant calls to c.selectPosition that were
immediately followed by c.redraw. c.redraw(p) also calls
c.selectPosition(p).

Imo, these changes are worth making in leoInteg and leojs.

Edward

PR #2417, opened 2022-02-14
Provide QtWebEngineWidgets imports to support QtWebEngineView. 

tbpassin:
Provides a new flag has_WebEngineWidgets that will be imported by leoQt from either leoQt5.py or leoQt6.py as appropriate.

Make VR3 use QtWebEngineWidgets.QtWebEngineView if available else the PyQt5 "WebView" (which was QtWebEngineView in disguise).

VR3 version to 3.8.

Corresponding changes to Freewin will be the subject of a separate PR.


edreamleo (approved):
I am unsure whether there should be messages if imports fail, but we can deal with this later.

edreamleo:
@tbpassin Thanks for this. After merging into devel I see no unwanted messages.

PR #2424, opened 2022-02-15
Prevent Leo from inserting cheat sheet into workbook.leo 

tbpassin:
unless workbook.leo does not exist.

Test for presence of "unl://" so Leo can open last session's files on startup.


edreamleo (approved):
Thanks for this. Don't worry about failing tests--they are an artifact of recent TravisCI changes.

Issue #2439, opened 2022-02-21
build_rclick_tree handles @others in a non-standard way


edreamleo:
In build_rclick_tree (mod_scripting.py) looks up `@others` in a non-standard way.

This may cause problems for any script that using `@others` for a purpose other than the usual, typical `@others` directive itself.

There's a method in g or in c to get the directives used in a node.

PR #2440, opened 2022-02-21
PR for #2439: mod_scripting.py

edreamleo:
See #2439.


edreamleo (commented):
The new has_at_others function is as close as I can come to an exact match to the code's stated purpose.

edreamleo (commented):
@boltex Please test the new code for yourself before reviewing :-)

boltex (approved):
I tried before and after with 'fake @others' and it now works as expected with your fix. 
The code itself seems  good now :) Thanks!

edreamleo:
@boltex Thanks for the review.

PR #2456, opened 2022-03-03
Automatic line highlight color based on style colors of most recently…

tbpassin:
… opened outline instead of original theme outline.

### DIFF
--- a/leo/plugins/qt_text.py
+++ b/leo/plugins/qt_text.py
@@ -846,8 +846,9 @@ if QtWidgets:
                 else:
                     hl_color = params['last_hl_color']
             else:
-                ssm = c.styleSheetManager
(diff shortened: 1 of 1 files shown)


edreamleo (approved):
Hmm. I think this will work, but I wouldn't bet my life that there are no unintended side effects.  I'll approve this change and we'll see what happens :-)

tbpassin:
It works in limited testing.  If you load a theme outline and the colors change, then the first time you select a new line in an outline, the highlight color changes to match.  I'm not sure the color will get reset correctly if you go reloading settings after closing that theme tab.  But it's better than what we had til now.

edreamleo:
@tbpassin Thanks for the testing. We'll see if anyone notices a change :-)

Issue #2430, opened 2022-02-17
CTRL-Click in <<named section>> Should Navigate To Its Node


tbpassin:
Much as a CTRL-click on a URL in the body will open the URL in a browser, and CTRL-click on a UNL will navigate to the corresponding  outline location, I would like to have a CTRL-click on a \<\<named section\>\> navigate to the node having that headline in the same outline (or more likely, in the same @file or @clean tree).

This capability would reduce the amount of scrolling and searching needed in the tree, which would remove another little bit of resistance in the programming process.

edreamleo:
@tbpassin I am not going to do this myself. I will consider a PR that implements this feature, but be warned: I will not accept a PR that significantly complicates the existing UNL code. I will not assign a milestone to the feature until I accept a PR.

PR #2457, opened 2022-03-04
CTRL-click on << named section >> navigates to the node having that …

tbpassin:
headline.  Implements issue [Ctrl-click On Named Section Navigates To That Section](https://github.com/leo-editor/leo-editor/issues/2430).  Does not affect CTRL-click on URL or UNL.

### DIFF
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -7860,11 +7860,10 @@ def openUrlHelper(event: Any, url: str=None) -> Optional[str]:
         if l_.startswith('<<') and l_.endswith('>>'):
             p = c.p
             px = None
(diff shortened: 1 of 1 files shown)


tbpassin:
Let's put this on hold for a bit.  It worked perfectly on the two test outlines I tried it on, but I just found one that it didn't.  I'll track down what the reason is and post redo the PR if needed.

tbpassin:
All right, got it fixed up.  p.nearest_roots() didn't always seem to work as expected, but just searching in p.subtree() works right, since the only allowed named sections are in the subtree.  This version works in the outline that failed earlier.

edreamleo (approved):

PR #2470, opened 2022-03-09
Fix Bad Winkey Symbol Insertion on Linux

tbpassin:
A simple hack to suppress the problematic Windows key behavior on some linux systems.


edreamleo (approved):
Thanks for this. Imo, this extremely limited code is exactly the right approach.

Also, the extended comment is fully warranted.

PR #2506, opened 2022-03-21
PR: Attempt to fix leoserver problems

edreamleo:
At present, test-all fails, but test-leoserver passes.

Note the disabled print statements in onSelectItem during unit testing.  I would not have expected those traces to fire.


edreamleo:
I am closing this issue.  It was only ever meant as an example for Félix.

PR #2509, opened 2022-03-22
Felix-server4

boltex:
Small additions/fixes. (set dirty / set changed on commands that should have dirtied the node/commander


edreamleo (approved):
Thanks for this work. I'll run the unit tests on my machine after merging into devel.

Issue #2519, opened 2022-03-25
Improve app.setIdFromDialog


edreamleo:
See PR #2520 and PR #2523.

app.setIdFromDialog should call sys.exit if the user refuses to supply a valid leoID.

vitalije:
> app.setIdFromDialog should call sys.exit if the user refuses to supply a valid leoID.

I don't think that is a very good idea. I would suggest that in this case Leo should generate random string. Most users, especially new ones, don't have any idea what the leoID is all about, and I believe that there is really any benefit in the ability to choose leoID. It would work perfectly well with any string.

If two or more people work together on some outline, then perhaps it is beneficial to have different leoIDs but even if they use same leoID,  probability to have a gnx clash is very small and possibility that two of them would create the same random string during the setup is almost zero.

If you think that this might happen, then perhaps leoID should look like three-four random letters followed by the number of milliseconds since the unix epoch.

I remember some new users have had problem that they couldn't install Leo because there were some issues with the showing the dialog. For example, users that want to use console GUI don't need to have installed neither Qt nor Tkinter, but they couldn't use Leo because it crashed every time trying to show the dialog. I would argue that for the most users it would be much easier if Leo hadn't asked them for leoID on the first launch.

We could say that setting the leoID is advanced feature for those who care about what their gnx will look like. Most users don't care at all, about that, and yet it is the very first (potentially confusing) experience they will have with Leo.

edreamleo:
@vitalije I considered, and rejected, these kinds of behaviors.​  The leoID is something that should be descriptive, as the dialog that prompts for them explains.

edreamleo:
@vitalije This issue exists because of a request from Félix. This is not the time or the place to re-imagine the role of leoID.

tbpassin:
As long as a user never has to later supply an ID that he may not remember, I suppose it's OK.  We never want to have a user locked out of their outlines because of forgetting the user ID.

edreamleo:
@tbpassin The worst that could happen is that Leo would again prompt for an id after the user mistakenly deletes leoID.txt (don't ask me how or why). There is never any need to remember anything: any id will do.

boltex:
@tbpassin @vitalije  

Just to reassure that this is a small change, and not a behavior breaking change: At very worst, this only changes behavior if used via leobridge: it would make Leo use an id string of  'None' for the current running session, without saving it to .leoId.txt file, if not found and not provided. So it's pretty safe and benign.

[...]

PR #2523, opened 2022-03-25
PR2 for #2519

edreamleo:
See #2519.

### DIFF
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -1118,7 +1118,8 @@ class LeoApp:
                 break
         #
         # Put result in g.app.leoID.
-        if not id_ and not g.app.inBridge:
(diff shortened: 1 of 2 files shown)


boltex (approved):
I tested it, along with leointeg and tried with https://github.com/boltex/leointeg/issues/248 in mind. 

This looks good to me! I also tried forcing the id to ot be found, this give a plain 'None' id which doesn't crash the bridge nor leointeg, so it's very nice !! :) thanks for this work!

 I'll take 5 minutes to add getLeoId and setLeoId to leoserver, to close up this issue which was sumbled upon by @ATikhonov2  and it will wrap up Leo 6.6 for me, along with wrapping up leointeg 1.0.7 !

Thanks again for this! :)

PR #2524, opened 2022-03-25
Update VR for Qt6

tbpassin:
VR:
1. Completed @pyplot nodes transition to qt6/qt5.
2. After every @pyplot rendering in VR, restore the default pyplot back end.
   (principle of least surprise for other matplotlib users).
3. Per pylint, fix two instances of local variable shadowing calling param.

VR3:
Abort processing @pyplot nodes with "not implemented" message.

test-all: passes
pylint: pyplot_backend: 9.67
           VR: 9.96 (there are some message for Qt6 imports about "c-extension-no-member)".  This commit only covers updating @pyplot capability.

### DIFF
--- a/leo/plugins/pyplot_backend.py
+++ b/leo/plugins/pyplot_backend.py
@@ -21,6 +21,8 @@ try:
     from matplotlib import pyplot as plt
 except ImportError:
     g.es_exception()
+
+import matplotlib
 #@-<< pyplot_backend imports >>
 #@+others
 #@+node:ekr.20160928073605.1: ** init
@@ -119,9 +121,8 @@ class LeoFigureManagerQT(FigureManager):
 
         self.canvas.figure.add_axobserver(notify_axes_change)
 
-        # close the figure so that we don't create too many figure instances
+        # Close the figure so that we don't create too many figure instances
         plt.close(canvas.figure)
-
     #@+node:ekr.20160929083114.1: *4* destroy
     def destroy(self, *args):
         # Causes problems.
--- a/leo/plugins/pyplot_backend.py
+++ b/leo/plugins/pyplot_backend.py
@@ -8,21 +8,17 @@
 #@+node:ekr.20160928074801.1: ** << pyplot_backend imports >>
 from leo.core import leoGlobals as g
 from leo.plugins import viewrendered as vr
-from leo.core.leoQt import isQt5, isQt6, QtWidgets
 from leo.core.leoQt import FocusPolicy
 try:
-    if isQt5 or isQt6:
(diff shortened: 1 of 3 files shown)


edreamleo (changes requested):
I have the following concerns
1. The call to g.es_exception() at the start of pyplot_backend.py.
2. The "from matplotlib import pyplot" statement outside a try/except block at line 1385 of viewrendered.py.
3. There is unreachable code starting at line 3590 of viewrendered3.py

tbpassin:
3). is the easiest.  I left the code there so if I decide to bring back @pyplot functionality in the future, it will be there to remind me.  I can comment it out or delete it, whichever you prefer.  I suppose deleting it would be the cleanest.  It would be available if needed from some earlier version in the repo.

1). The call to g.es_exception()  was like that in the previous code - I didn't change it.  If you want me to change it, let me know what you suggest.

2) I misremembered, thinking matplotlib had been imported earlier in the file so it wouldn't need to be tested.  I've added a try/except, but I won't push until I see your suggestions about the other two points.

edreamleo:
@tbpassin
- Please use `if 0:` To disable code. pylint is happier with that.
- About g.es_exception: I didn't realize it was there before. Never mind.
- I would prefer all imports be at the start of the file. If necessary, assign the imported symbol to None in an except block, then test for the imported symbol in later code.

tbpassin:
Additional changes to implement @ekr's sugggestions:
1. VR: import pyplot at module start;

2. VR: Test for pyplot existence at start of @pyplot processing;
3. Disable VR3 pyplot processing code with if not True;
4. VR: disable pylint warnings about inaccessible c imports;
5. VR: Disable pylint warning "too-many-public-methods";
6. VR: Implement two minor pylint suggestions.

VR pylint score: 9.99
VR3 pylint score: 10.0

tbpassin:
@edward, I didn't see an indication that this PR knows about the 2nd commit.  In the past when I pushed a new commit I have seen that the PR knew about it.  With the additional changes there should be an import in the VR import section:

```
got_pyplot = False
try:
    # pylint: disable=import-error
    from matplotlib import pyplot
    got_pyplot = True
except ImportError:
    pass
```

edreamleo:
@tbpassin It looks like you didn't push to origin. Please do so.

tbpassin:
I had pushed to origin, and when I did it again, I get a message that said origin was already up to date.  I have just added one blank line, committed and pushed that.  Now I will go to my github pages and see if I can submit a PR for the changes.  Hopefully that will be combined with this one.

tbpassin:
Github recognized the new commit but didn't offer a new PR - so you should be able to get the changes now.

edreamleo (approved):

Issue #2528, opened 2022-03-27
Review leoBackground.py


edreamleo:
See PR #2537.

The legacy code processed events only when pylint produced output.

The new code is much more responsive, but sometimes *all* output disappears.

Leo's mypy command doesn't use the BPM, which may be a problem.

edreamleo:
Running pylint on a new version of leoAst.py shows that the output can still block.

PR #2537, opened 2022-03-28
PR for #2528: background process manager mostly hangs

edreamleo:
See #2528.


edreamleo:
Merged into ekr-6.6.1-devel.

PR #2542, opened 2022-03-29
Get Qt version from Qt C++method instead of from the Pyqt sip wrapper…

tbpassin:
… constant PYQT_VERSION_STR.  The latter has been wrong at least once.


edreamleo (commented):
Looks good.  I'll merge this into devel after 6.6 is released.

edreamleo (approved):

Issue #2539, opened 2022-03-28
Fix new pylint warnings


tbpassin:
**EKR**: Upgrading to python 3.10 results in more warnings.  I'll fix them all.

pylint found two instances of this type of warning in LeoGlobals.py:

` W1113: Keyword argument before variable positional arguments list in the definition of __init__ function (keyword-arg-before-vararg)`

I am not sure if this could potentially cause a problem or not.  If not, we could clean it up for 6.6.1.  If it could, perhaps it should be done before 6.6 is released.

edreamleo:
@tbpassin 6.6 is frozen. It can wait until 6.6.1.

I don't see this pylint warning. What function is the complaint about?

tbpassin:
I just updated my pylint version yesterday.  Maybe that has something to do with it.

`C:\Tom\git\leo-editor\leo\core\leoGlobals.py:2266:4: W1113: Keyword argument before variable positional arguments list in the definition of __init__ function (keyword-arg-before-vararg)`

Hmm, I thought there were two, but now I only see this one case of it.  It's the __init__ for `class TracingNullObject`

edreamleo:
@tbpassin I just install Python 3.10, and I see many new warnings.  I'll fix them all.

tbpassin:
I'm still on 3.9x, but your new version probably got you the newest version of pylint, too.  Yes, I saw a lot more warnings that I didn't mention, but many or most of them just need to be disabled.

edreamleo:
@tbpassin I agree. I'll just disable the warnings.

PR #2544, opened 2022-03-29
PR for #2539: Suppress new pylint warnings

edreamleo:
See #2539


edreamleo (commented):
pylint passes all files on my machine with both python 3.9 and 3.10.

Issue #2546, opened 2022-03-29
Ctrl-click a x.py should find the corresponding @<file> node


edreamleo:
OMG: How did we ever live without this?!?

tbpassin:
Do you mean that if a body line contained the space-separated string _fancy_plugin.py_, then CTRL-Clicking it should cause the selected node to change to the _@file fancy_plugin.py_ or _@clean  fancy_plugin.py_ node in that outline?  Sounds good to me.

It would take a regex that wouldn't match on a url, gnx, unl, or << section reference >> by mistake.  That sounds doable.

edreamleo:
On Tue, Mar 29, 2022 at 11:10 AM Thomas Passin ***@***.***>
wrote:

> Do you mean that if a body line contained the space-separated string
> *fancy_plugin.py*, then CTRL-Clicking it should cause the selected node
> to change to the ***@***.*** <https://github.com/file> fancy_plugin.py* or ***@***.***
> <https://github.com/clean> fancy_plugin.py* node in that outline?
>
Yes.

> It would take a regex that wouldn't match on a url, gnx, unl, or <<
> section reference >> by mistake.
>
Happily, the task is easier than that. openUrlHelper does all the work, but
only if none of the other cases apply.

I've just now pushed the code. See PR #2549
<https://github.com/leo-editor/leo-editor/pull/2549>.

Edward

tbpassin:
I figured you'd pump this one out fast.

PR #2549, opened 2022-03-29
PR for #2546: ctrl-click filename searches for its node

edreamleo:


edreamleo:
Merged into ekr-6.6.1-devel.

Issue #2557, opened 2022-04-01
Running Leo's pylint command blocks pylint output


edreamleo:
That is, #2528 (merged into devel) has problems.

PR #2561, opened 2022-04-01
PR2 for #2557: pylint hangs, again

edreamleo:
See #2557.


tbpassin:
Still non-blocking but no output:

```
Leo 6.6.1-devel, devel branch, build b6fc9e4495
2022-04-01 14:32:01 -0500
Python 3.9.9, PyQt version 6.2.4
Windows 10 AMD64 (build 10.0.19043) 
```

PR #2563, opened 2022-04-02
PR: Fully annotate leoAst.py

edreamleo:
- [x] Fully annotate the TOG class.
   OMG: mypy found real errors. Wrappers need not (should not) call visit!
- [x] Require full annotations and return types in leoAst.py.
- [x] Fully annotate all of leoAst.py.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -44,7 +44,7 @@ exclude =
 
 # Settings for particular files...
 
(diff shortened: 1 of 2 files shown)


edreamleo (commented):
Issues #2564, #2565, and #2566 will continue this work.

PR #2583, opened 2022-04-06
Fix VR3 handling of missing @jupyter modules

tbpassin:
1. Fix missing error for nbconvert;
2. Improve error messages for missing imports for @jupyter nodes.
3. Version to 3.8.
test-all:

```
Ran 920 tests in 18.484s
OK (skipped=8)
```

### DIFF
--- a/leo/plugins/viewrendered3.py
+++ b/leo/plugins/viewrendered3.py
@@ -946,17 +946,25 @@ except ImportError:
     print('VR3: *** No numpy')
     np = None
 # nbformat (@jupyter) support, non-vital.
-jupyter_ok = False
+jupyter_ok = nbformat_ok = nbconvert_ok = False
+
 try:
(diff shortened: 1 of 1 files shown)


edreamleo (changes requested):
Doing AI on the exception is not necessary.  Please split the import into two pieces, with a clear explanation in each "except" block about how to install the missing package.

tbpassin:
All right, changes have been pushed.

edreamleo (approved):

edreamleo:
@tbpassin Good work. I got the message about `pip install nbconvert`, and doing so eliminated all messages from VR3.

tbpassin:
Good to hear!  I tested it by mis-spelling the names of the modules (like xnbconvert) in the source code rather than by actually uninstalling anything.

PR #2600, opened 2022-04-13
Annotate Leo's gui files

edreamleo:
There is no need to do this for 6.6.1, so I won't :-)

I'll use/improve Leo's add-mypy-annotations command.


edreamleo:
I am going to close this PR. It changed too many files, and the work on leoFind.py was done by hand. I'll create a new summary issue, and issue PR's for individual files.

PR #2632, opened 2022-05-01
mypy fixes/suppressions

edreamleo:
- [x] Define SolidLine in leoQt5 and leoQt6.
- [x] Define Widget = Any in four qt files.

### DIFF
--- a/leo/core/leoQt5.py
+++ b/leo/core/leoQt5.py
@@ -112,6 +112,7 @@ Shadow = QtWidgets.QFrame
 Shape = QtWidgets.QFrame
 SizeAdjustPolicy = QtWidgets.QComboBox
 SliderAction = QtWidgets.QAbstractSlider
+SolidLine = QtCore.Qt
 StandardButton = QtWidgets.QDialogButtonBox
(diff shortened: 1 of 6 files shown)


tbpassin:
When a name, say Editor, is set to `typing.Any`, its truthiness is True.  That might be unexpected if someone uses the name for some other purpose. Are these variables going to be used only as types for mypy?  If not, their type will change when they are assigned.  If yes, their names should reflect their special status so they won't be reused by mistake.

In fact, these names - if they are going to be used for typing - could all be put into their own module. that would remove any accidental  name clash.  Perhaps _leoTyping.py_.

if these remarks are off target, please explain the intended use of the new statements like these:

`Editor = Any`

the `SolidLine` changes look right.

edreamleo:
@tbpassin Yes, these assignments are workarounds for problems with mypy type aliases. Imo, there are serious problems with mypy in this regard. For example, it seems to be impossible to define QtWidgets properly for both pyQt5 and pyQt6. Details omitted. I'm done with mypy for now.

PR #2644, opened 2022-05-08
This fixes bug that was introduced in 'dehoist' command in the 'Mass update using regex-repl script' changes.

### LINKED ISSUE #2643: Bug was introduced in 'dehoist' command in the 'Mass update using regex-repl script' changes.
The dehoist command was altered by accident in the 'Mass update using regex-repl script' commit 7 months ago. Comparing with all other changes made in this file, (all mass-updates of 'p' testing'), at the same time it is obvious that it was an accident.

The bunch flag 'isExpanded' is ignored and the contract/expand operation is made on an arbitrary boolean test of 'p'. 

(compared to the old behavior which made sense and was in place for many years/decades).

dehoist operation implementation should be reverted to the old and correct way.

boltex:
Fixes #2643

I made sure to run tests on Windows too just in case: all tests pass.

### DIFF
--- a/leo/commands/commanderOutlineCommands.py
+++ b/leo/commands/commanderOutlineCommands.py
@@ -926,7 +926,8 @@ def dehoist(self, event=None):
         return
     bunch = c.hoistStack.pop()
     p = bunch.p
-    if not p:
+    # Checks 'expanded' property, which was preserved by 'hoist' method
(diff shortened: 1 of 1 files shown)


edreamleo (approved):
Thanks again!

PR #2656, opened 2022-05-19
PR: create c.copy_tree

edreamleo:


edreamleo:
This original algorithm has no chance of working.

PR #2665, opened 2022-05-21
Tbp find pane

tbpassin:
Add "Files Only" capability to Find tab.


edreamleo (commented):
There are two methods with the same decorator:  @cmd('set-find-file-only'). What's going on?

It's too late for this PR to be part of Leo 6.6.2.  I'll approve it after 6.6.2 is released.

tbpassin:
It must have been the result of doing the first part of the work a month or more before the second part.  I will see if there really is a need for the extra code in the second version.

tbpassin:
All fixed.  The second block with the decorator is needed but without the @cmd decorator.

```
pylint: leoFind.py: 9.95/10
        qt_frame: 9.91/10 
```

edreamleo (commented):
Looks good. I'll approve after making the 6.6.2 branch. Coming soon.

tbpassin:
: )

edreamleo (approved):

Issue #2679, opened 2022-05-28
Remove c.frame.startupWindow ivar


edreamleo:
See PR #2690.

c.frame.startupWindow is never set. In consequence, the closeFlag var is always False in the open_outline and openRecentFile methods in leo/commands/commanderFileCommands.py.

- [x] Remove the startupWindow ivar.
- [x] Remove closeFlag and related logic.
- [x] Change documentation of recentfiles1/2 events.

boltex:
In my opinion, i would guess that this logic was good when the initial opened document was an empty untitled document, and not the leocheatsheet. At that time, that untitled document's c.frame.startupWindow was set to true. and the 'closeFlag' also made sense in order to close the very first untitled document if it was not dirty.

I personally preferred it that way, but its not a big deal. I wonder if it should be made optional, like to 'show tips at startup'.  

That way we could keep and continue to use the c.frame.startupWindow and closeFlag flags. 

I'll wait for your comments on that matter :)

boltex:
Also, to be more precise: the way the logic is setup right now is not broken. 

The startupWindow would still be needed, and set to 'true', if an empty & untitled document was set up as the first document when invoking Leo without any recent files to show. (e.g. instead of the cheatsheet)

edreamleo:
On Mon, Jun 20, 2022 at 12:51 AM Felix ***@***.***> wrote:

> Also, to be more precise: the way the logic is setup right now is not
> broken.
>

I'm having trouble understanding this issue. Could we zoom about it?

Edward

edreamleo:
@boltex Per our zoom conversation, the fix will be to remove c.frame.startupWindow and all code related to the "closeFlag" far in c_file.open_outline and c_file.openRecentFile.

PR #2680, opened 2022-05-29
PR for #2679: LeoFrame.startupWindow

edreamleo:
See #2679.


edreamleo:
Félix tells me that removing the startupWindow ivar is the wrong approach. I am going to close this PR and wait for a PR from Félix.

PR #2681, opened 2022-06-07
PR: retire gcm.inited and gcm.setIvarsFromSettings

edreamleo:
Félix reports that g.app.config.inited (gcm.inited) is never cleared. But this means that gcm.setIvarsFromSettings is a no-op.

New traces seem to confirm this suspicion. I'll wait for Félix's review before retiring gcm.inited and gcm.setIvarsFromSettings.


boltex:
Looking at c.initSettings in leoCommands.py, in devel branch, build 71ce553a40 before those fixes...

I'm trying to understand why it would be a no-op... is it because in the method where **setIvarsFromSettings** is called, (the last line of **initSettings**),  the line right before it is called would have set all those ivars first? and thus the call to g.app.config.setIvarsFromSettings(c) is redundant?

(I'll investigate more tomorrow when i'll have more time)

boltex:
oh nevermind, you only retired gcm.inited, not gcm.setIvarsFromSettings ! :)  (The title of the pull request led me to beleive you removed the method and i didnt even looked at the changed files!)

edreamleo:
@boltex Please let me know if you think this PR is ready to mere.

boltex:
I also retired those from leojs and what do you know? You were right everything still runs fine and the settings from leoSettings, myLeoSettings, or @settings nodes all still work fine and apply in order, when loading , or using reload settings, just fine. :)

edreamleo:
On Thu, Jun 23, 2022 at 9:25 PM Felix ***@***.***> wrote:

> I also retired those from leojs and what do you know? You were right
> everything still runs fine and the settings from leoSettings,
> myLeoSettings, or @settings nodes all still work fine and apply in order,
> when loading , or using reload settings, just fine. :)
>
Thanks for the confirmation. I'm starting to think that most of
g.app.config (GlobalConfigManager class) can go away!

Indeed, all those default settings lists and dicts could be replaced by GCM
ivars, or maybe be moved into other code entirely. I must study the code
more, but the GCM settings look like emergency fallback values to me. If
only one place (outside the GCM) uses that fallback, then that place should
control the "policy" about what the fallback should be.

As I read the code, user settings do not (and probably should not) set
these fallback values.  In that case, most of the GCM can disappear, and
good riddance imo :-) Only a few utilities would remain. The
simplifications will ripple throughout Leo. I'm keeping my fingers crossed.

I'll study the code again before doing an ENB on this topic. I hope to
reach a conclusion later today.

Edward

edreamleo:
On Fri, Jun 24, 2022 at 6:05 AM Edward K. Ream ***@***.***> wrote:

Most (all?) uses of g.app.config ivars can and should be replaced by calls
> to c.config.get*.
>

PR #2692 <https://github.com/leo-editor/leo-editor/pull/2692> contains a
grand simplification of Leo's config code! It's still a work in progress.

Edward

PR #2692, opened 2022-06-24
Ekr config

edreamleo:
A grand simplification of the GlobalConfigManager (GCM) class:

The GCM no longer sets ivars or specifies defaults. This eliminates a lot of code and data!  More importantly, the GCM no longer sets policy. Rather than using g.app.config ivars, code throughout Leo calls c.config.get* to get settings.


boltex:
in leoGlobals, in the getBaseDirectory method, this should be a getString, not a getBool

base = c.config.getBool('relative-path-base-directory')

edreamleo:
@boltex Thanks for checking the code. Fixed in devel and felix-server4.

Issue #2718, opened 2022-07-13
Using chapter-next to get into a chapter instead of the menu needs a 'tree-refresh' to be performed


boltex:
See PR #2718.

The chapter commands and selection from the chapters dropdown menu seem slightly broken.

When using the chapter-select dropdown menu in an outline that has chapters, that gets me hoisted into that chapter. So far this is the only chapter related action that seems to work ok in Leo... :/

Using chapter-next and chapter-prev command (via alt+x) just select the node without visually hoisting it. (But then, changing the tab to another opened outline, and coming back forcing a refresh, gets it hoisted properly)

Also, while hoisted in a chapter, doing chapter-next sometimes just yields the  'can not hoist an @chapter node.' message and/or does nothing. 

Maybe this is also related to a missing 'tree-refresh' action that is missing after one of those commands has been issued.

Lastly I feel that doing a de-hoist command while in a chapter should do the equivalent of select-chapter-main... (is it already the case and it's just not refreshing / updating properly?)

(can anyone else confirm those behaviors?)

tbpassin:
I haven't looked at your items, but I just noticed that selecting a chapter in the chapter drop down list does not include the top-level node of the chapter - that is, the node with the "@chapter" headline.  IMO, the top-level should be included in the displayed tree.

tbpassin:
"Lastly I feel that doing a de-hoist command while in a chapter should do the equivalent of select-chapter-main... (is it already the case and it's just not refreshing / updating properly?)"

On my system chapter-select-main does not work the same as the Dehoist menu item.

boltex:
@tbpassin oh i see: perhaps you're refering to the fact that "chapter-select-main" is akin to a 'de-hoist-all-levels' instead of a single 'de-hoist' if hoisted more than once.

On the other hand, I've looked at the code and there seems to be missing c.redraw calls a the end of many chapter related commands. 

Lastly, about the @chapter headline being omitted, it probably was intended... but I also personally feel that it would be nice to have it throne at the top like a regular hoist, (when choosing a chapter with the chapter-related commands).

I'll wait for @edreamleo  to chime in, and I'll propose to fix all the chapter-related issues in one go (shouldn't be long nor hard) upon Edward's say on the matter.

boltex:
oh! 

I noticed that this documentation page is very much deprecated (header says "Leo 4.11") : https://leoeditor.com/outlines.html#using-chapters This contains references to commands that do not exist.

[...]

PR #2721, opened 2022-07-14
PR for #2718: chapters

edreamleo:
See #2718.

- [x] selectChapterByNameHelper calls c.redraw.
- [x] Add special case to dehoist command.
- [x] Fix the docstring for chapter-select command.
- [x] Fix ordering bugs in the chapter-back/next commands.
- [x] cc.selectChapterByNameHelper changes *no* expansion bits.

### DIFF
--- a/leo/core/leoChapters.py
+++ b/leo/core/leoChapters.py
@@ -370,7 +370,7 @@ class Chapter:
     #@+node:ekr.20070423102603.1: *4* chapter.chapterSelectHelper
     def chapterSelectHelper(self, w=None):
 
-        c, cc = self.c, self.cc
+        c, cc, u = self.c, self.cc, self.c.undoer
         cc.selectedChapter = self
         if self.name == 'main':
             return  # 2016/04/20
@@ -397,9 +397,11 @@ class Chapter:
             if p.hasChildren():
                 self.p = p = p.firstChild()
             else:
-                # 2016/04/20: Create a dummy first child.
+                bunch = u.beforeInsertNode(p)
+                # Create a dummy first child.
                 self.p = p = p.insertAsLastChild()
                 p.h = 'New Headline'
+                u.afterInsertNode(self.p, 'Insert Node', bunch)
         c.hoistStack.append(g.Bunch(p=root.copy(), expanded=True))
         # Careful: c.selectPosition would pop the hoist stack.
         c.setCurrentPosition(p)


leo/core/leoChapters.py
```diff
@@ -164,6 +164,7 @@ def selectChapterByNameHelper(self, chapter, collapse=True):
         c.contractAllHeadlines()
         chapter.p.v.expand()
         c.selectPosition(chapter.p)
```
boltex, line 166: I think **c.selectPosition(chapter.p)** is wrong, this should be:
```
if chapter.name == 'main': 
    c.selectPosition(chapter.p)  # as normal
else:
    c.selectPosition(chapter.p.firstChild())  # select first node inside
```

edreamleo:
@boltex I'll make the change you suggest soon. Thanks for the thorough review!

edreamleo:
@boltex Hmm. The change caused a unit test to fail. It's time for a break. I'll resume tomorrow.

boltex:
@edreamleo Thanks! 

i've taken the liberty to investigate why the test fails. I think i've found out what the real problem, and solution, are !!  

the failing test is setup to check exactly what i'm trying to solve by selecting the child of the chapter.p node. and when running the test, strangely, it is the child that is selected. ... So everything seems to be perfect and my proposed fix somewhat dubious to say the least... 

But! I've found out that the behavior of the chapter getting it's 'chapter.p' is different when running the test-all command,  compared to actually using it in Leo! e.g.: with alt+x and chapter-next 

hoho! so that led me to investigate the setting of chapter.p upon creation of a chapter, and, the behavior of chapter.select.... They all seemed good so I was even more confused than anything else at that point !!!

Turns out that the seemingly inoffensive  **c.contractAllHeadlines()** in the middle of **selectChapterByNameHelper** is the main suspect!! i found that out by observing that chapter.p was different after running this line! 

commenting out this line makes everything work fine and my proposed fix is not needed at all. (the first child of the chapter node will be the chapter.p, and the final, selected node)

I've tried protecting chapter.p by doing chapter.p = chapter.p.copy() but it didn't change anything 

So as a last  'clean' solution, i propose changing the `c.contractAllHeadlines()` by : 
```
for v in c.all_nodes():
        v.contract()
        v.expandedPositions = [] 
```

which is in essence, what contractAllHeadlines wants to do. 


this really fixes everything, all tetsts pass , and my proposed fix of selecting the .firstChild() if not 'main' was not the proper solution after considering the real fix I just found. 

:) 

I'll annotate the PR like i did yesterday to reflect this !

Glad to finally cleanup the chapters code!! (I'm exposing chapter commands with icons in Leo 1.0.8 -> coming days after Leo 6.6.3)

boltex:
btw i dont know why contractAllHeadlines() changes chapter.p.... but i'm sure it does because I just found out where it happened by doing :
`print('chapter.p is ' + chapter.p.h + '\n')`
before and after each line in selectChapterByNameHelper ! hahah! yeah , the poor man's debugger tool!

edreamleo:
@boltex Thanks for your investigations :-) I use g.trace all the time (instead of print) because it tells me what function/method created the traces.

I'll investigate why c.contractAllHeadlines changes chapter.p. Leo's selection logic extremely complex, for (mostly) good reasons, but I'm not surprised that there are surprises :-)

edreamleo:
@boltex As shown in d7775d0, the best solution seems to be for the chapter selection logic not to change *any* expansion bits:

- No more surprises re c.contractAllHeadlines().
- No faux-helpful (confusing!) calls to chapter.p.v.expand()

edreamleo:
@boltex I'm ready to merge this when you give the go-ahead.

boltex:
Works fine now, and leaving the expand/collapse states intact feels more natural to me when going into, or out of, chapters, to see the state of the tree as I left it when coming back to it. less surprises is good :)

One tiny defect still remains : While playing with it I observed that when going into an empty chapter creates a new headline inside of it so that it has at least one child to be present in the chapter. **in chapterSelectHelper method of the Chapter class** (makes sense) but, this child creation does not create an undo bead! so this 'silently breaks' the undo chain of operations if I may put it this way. 

I'll annotate the pr with a code suggestion :)

boltex:
Oh well seems i cannot annotate the code below a a maximum point of change in the github interface, so ill put it here: 

the lines 
```
            # 2016/04/20: Create a dummy first child.
            self.p = p = p.insertAsLastChild()
            p.h = 'New Headline'
```
should be prepended and appended with the usual undoer before and after calls, something like :


```
            u = c.undoer
            u_d = u.beforeInsertNode(p)
            # 2016/04/20: Create a dummy first child.
            self.p = p = p.insertAsLastChild()
            p.h = 'New Headline'
            u.afterInsertNode(self.p, 'Insert Node in ' + self.name, u_d)

```
Tried it in Leo itself, (i thought maybe it would crash leo if i did the undo while in the chapter) but it just removes the new node and gets me 'out' of the chapter without crashing or errors , so this is great - and i think would close the lid on the Chapters modifications for 6.6.3. hehe...

boltex:
btw undo conundrums are so easy to spot now in the soon-to-be-released leointeg 1.0.8! 

All thanks to the new 'undo history' panel that lets you visualize the undo beads, with the last highlighted, and right-clickable to change state to any bead instantly!  Works just like in photoshop! :)

edreamleo:
@boltex Thanks for the additional fix.  I'll add it later today.

edreamleo:
@boltex Rev 8166124 does as you suggest.

edreamleo:
@boltex Please let me know when you are ready to have this PR merged.

boltex (approved):
This is now working perfectly in Leo, leoBridge, leoserver and leointeg.

PR #2723, opened 2022-07-18
Felix-server4

boltex:
Fixes uA's set from reading external files in .leojs files. (Makes the .leojs file format read / write all nodes and external files and ua's properly) 

In leoserver.py, I also added chapters commands  and chapters 'rendering', ua setting commands,  and some small fixes to common commands.

### DIFF
--- a/leo/core/leoFileCommands.py
+++ b/leo/core/leoFileCommands.py
@@ -572,6 +572,9 @@ class FastRead:
         #
         # Traverse the tree of v elements.
         v_element_visitor(v_elements, hidden_v)
+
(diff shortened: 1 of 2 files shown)


edreamleo (approved):
The change to FastRead.scanVnodes is surprising, but it looks safe enough.  Good work.

boltex:
you mean FastRead.**scanJsonVnodes** :) ? 

This makes the JSON .leojs file format able to read/write all aspects of a leo document properly!... 100% equivalent to the XML .leo format! :)

Issue #2727, opened 2022-07-18
Simplify Leo's importers using python importer as a template


edreamleo:
Vitalije's new python importer is contains a brilliantly simple algorithm that uses python's tokenizer module. A generalized version of this algorithm is now the basis for the code in the Importer class.

**Note**: leoJS probably can't use use Vitalije's python importer because python's tokenizer module doesn't exist in typescript. So leo.plugins.importers.python contains two versions of the python importer: one for Leo itself and one for leoJS.

[This post](https://groups.google.com/g/leo-editor/c/doUUtmabvcw) discusses the new importers. To summarize, the generalized importer algorithm in the Importer class separates:
- the detection of **blocks** (functions, classes, and methods) from,
- the allocation of blocks to nodes.

**Importer.gen_lines** is the main line of the Importer class (and subclasses).

**i.get_block** computes the text range for one block and returns a named tuple describing the block's text range (and other data). A prepass in i.gen_lines calls this method repeatedly to detect all blocks.

**i.make_node** allocate blocks to child nodes using *only* text ranges.  This method calculates the indentation of `@others` statements using only the indentation of the newly-allocated nodes.

tbpassin:
Unless the importers don't work right, I'm inclined to go with "If it ain't broken don't fix it".

edreamleo:
@tbpassin The python importer will probably need to be replaced for leoJS. We'll see about the other importers later. It may be that only gen_lines will need to change to make everything simpler.

Remember that strong unit tests exist for most importers, so there is little risk.

PR #2729, opened 2022-07-20
PR for #2727: simplify importers

edreamleo:
See #2727. 

**Overview**

- [x] Replace the iterative logic in i.gen_lines by recursion, generalizing the ideas in Vitalije's python importer.
   All faux helpers (functions that had easily understood purpose) are gone.
- [x] Strengthen most unit tests using the check_outline function.
- [x] Require mypy annotations for all importers.
- [x] Delete importers, writers, commands and tests for .json and .ipynb files.
    No change to .leojs code.
- [x] Replace all scanner tables and related logic by i.scan_all_lines. A huge step forward.
    See #2741.
- [x] Full coverage tests for all importers and writers.

**What I did**:

- [x] Remove "parse_body" kwarg ctors.
- [x] Remove add_key function (in get_new_dict) from most importers. Use known comments strings instead.
- [x] Restore legacy python importer as a **prototype** for leoJS. It will never be part of Leo itself.
- [x] Add @language and @tabwidth directives in i.gen_lines.
- [x] Make Cython_Importer a subclass of Python_Importer.
- [x] Remove g.app.suppressImportChecks.
- [x] Importers never add underindented escape string.
- [x] Pass c, not importCommands, to all importers.
- [x] Eliminate `**kwargs` from the ctor of most importers.
- [x] Eliminate `atAuto` kwarg.
- [x] Require full annotations for all importers and writers.
- [x] Generate `@language html`, not `@language xml`: use self.name, not self.language.
- [x] Add check_flag (round-trip) and strict_flag to run_test.
- [x] Most tests pass round-trip tests. Some tests can't possibly pass round-trip tests.
- [x] Most python tests pass with strict_flag == True.
- [x] cython tests with strict_flag = True.
- [x] Replace class methods with top-level do_import callback functions.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -40,9 +40,8 @@ exclude =
 disallow_untyped_defs = True
 disallow_incomplete_defs = False
 
-# Importer plugins should be fully annotated...
-# [mypy-leo.plugins.importers.linescanner,leo.plugins.importers.python]
-[mypy-leo.plugins.importers.*]
+# Importer and writer plugins should be fully annotated...
+[mypy-leo.plugins.importers.*,leo.plugins.writers.*]
 disallow_untyped_defs = True
 disallow_incomplete_defs = True
 
--- a/leo/core/LeoPyRef.leo
+++ b/leo/core/LeoPyRef.leo
@@ -96,8 +96,6 @@
 <v t="ekr.20220213142735.1"><vh>function: full_match</vh></v>
 </v>
 </v>
-<v t="ekr.20201013034742.25"><vh>@@button test-jup-export</vh></v>
-<v t="ekr.20201013034742.26"><vh>@@button test-jup-import</vh></v>
 <v t="ekr.20220211154839.1"><vh>@@button test-show-clones</vh>
 <v t="ekr.20191010112910.1"><vh>@g.command('show-clone-ancestors')</vh></v>
 </v>
@@ -161,6 +159,9 @@
 </v>
 <v t="ekr.20210911191731.7"><vh>@data:@auto-otl ../test/external_files/at-auto-otl-line-number-test.otl</vh>
 <v t="ekr.20211107080227.7"><vh>@gnxs</vh></v>
+</v>
+<v t="ekr.20220811130331.1"><vh>@data:@auto-json c:\test\at-auto-json-test.py</vh>
+<v t="ekr.20220811162331.1"><vh>@gnxs</vh></v>
 </v>
 </v>
 <v t="ekr.20170427114412.1"><vh>scripts</vh>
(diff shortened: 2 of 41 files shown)


vitalije:
I urge you to reconsider your idea of turning helper functions into a methods, before it is too late.

I believe it is a bad idea. I suppose you wanted to make Cython importer a subclass of Python importer. And yes, that way, it would be less code in Cython importer. But the code will become uglier. So it is a trade off. On the other hand, Cython importer well may be the only importer that would ever extend the Python importer, and it is written only once. I would rather choose nicer code using helper functions and just copy-paste-edit(as necessary)  all helper functions.

The advantage of inner helper functions over methods is that:
  -  they can easily share data,
  -  they don't pollute the namespace,
  -  they have no attachments outside the main function/method,
  -  they produce no visible effect on the rest of the code.

Methods however:
  - have to access shared data through self parameter,
  -  therefore each partial result must have some external name,
  -  methods can be called from other modules, other functions, other methods,
  -  they all need some external names too.

All of that makes a big contribution to overall complexity of the code. All those bad consequences for just one benefit: that one other importer can spare a few lines of code. I don't think it would be a good trade.

edreamleo:
@vitalije First, it is *never* too late to change code :-)

Classes are python's way of sharing common, *related* code. I see nothing wrong or dubious about the Cython importer. Imo, the present Cython importer clearly shows that the Cython and Python importers are almost identical. 

Imo, methods do *not* "pollute" the namespace of their classes. To the contrary, methods gain clear access to ivars in a more natural (imo) way than using non-local vars.

Yes, I understand that this is not your prefered style, but I'm doing the work and I think I'm entitled to my opinion.

vitalije:
@edreamleo Ok. I hope I didn't sound too demanding. My only intention was to offer you another point of view.

Please, don't take this as an offense or my stubbornness, but just for the sake of discussion, let me point out a few things. Using classes and methods for sharing data is certainly possible and allowed in Python. I am not saying it isn't. But using inner function definitions and non local vars is also perfectly natural in Python. It isn't any less pythonesque. So, if you don't use this advantage that Python offers, your code tends to be more verbose than it must be. In my opinion this is sub-optimal.

Once more, just for the sake of discussion, look at the following lines from the zen of Python:
  - Flat is better than nested
  - Sparse is better than dense
  - Readability counts

I think that in comparison with inner helper functions and non local vars, methods and instance vars are more like nested than flattened. In order to use them, one has to go through upper frame (through self parameter). `self.my_instance_var_a = self.my_method(self.my_instance_var_b)` is more dense than `a = f(b)`. Because those vars are used exclusively inside one simple function or method, they don't need to have too elaborate names. The same is true about inner helper functions. Although nothing is preventing you to have elaborate names anyway.

And finally if you ever wish to change something inside such function, you can be pretty sure, that you can make any change you like, and no one will notice the change. It isn't so if you declared class with methods and instance vars. You'll have to first look everywhere else to ensure that no code is relying on the old implementation. So it increases complexity of code, and *Simple is better than complex*.

Once again, I don't insist on this. Take this just as a suggestion for some future coding and think about it, then do whatever you feel is right.

edreamleo:
@vitalije The question of when (or if) to share code is tricky. The latest two importers are the .org and .otl importers. These importers are similar. Nevertheless (following your guidelines) I have chosen to make them independent. It's a judgment call. In the past I would have created faux helpers. This time I refrained.

But imo the unit tests are most important. I have improved the old (feeble) tests by using your check_outline method. Imo, this method is just about perfect.

vitalije:
@edreamleo I am glad if I've been of any help.

edreamleo:
@vitalije Yes, you have been of immense help :-)

PR #2744, opened 2022-08-17
PR: add annotations to leoserver.py

edreamleo:
mypy warnings (loosely) suggest the following changes (All methods are server methods):

- [x] _ap_to_p returns None, not False on error returns.
- [x] Remove "strict" kwarg to _get_p, returning a valid Position (or raise an error).
- [x] Add _get_optional_p, returning a valid Position or None (or raise an error).
    Called only from get_children.
- [x] d_to_childIndex_v returns None, not False, on error returns.

**Other changes**

- [x] Don't use the (deprecated) PosList class. It's better to create lists of positions!
- [x] Catch json serialization problems in the outline_to_json *function*.
- [x] test_most_public_server_methods strips (for the unit test only) all uA's from test/test.leo.
- [x] mypy passes with annotations required for leoserver.py.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -40,6 +40,11 @@ exclude =
 disallow_untyped_defs = True
 disallow_incomplete_defs = False
 
+# leoserver.py should be fully annotated...
+[mypy-leo.core.leoserver]
+disallow_untyped_defs = True
+disallow_incomplete_defs = True
+
 # Importer and writer plugins should be fully annotated...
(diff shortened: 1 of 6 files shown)


boltex (approved):
Full type annotations for the server! Wow! This is great!! 
Looks all good to me! I also appreciate the conversion to more pythonic code of some of the parts that you made!

edreamleo:
@boltex Glad you like the PR.  I'll be adding more to this PR today.

edreamleo:
Just created the second pull request: PR #2749.

PR #2752, opened 2022-08-20
Felix-server4

boltex:
Fixed all nav search related problem following the search-controller cleanup. 
(A bit too much cruft was removed, so only headlines were shown in results, This small fix now gets the body matches shown in the results)


edreamleo (commented):
Looks good. Let's not merge anything into devel now. We can do it later, all at once.  Sound like a plan?

edreamleo:
@boltex I am going to close this PR. I've "cherry picked" leoserver.py and three other files and merged those files into devel.  See PR #2755.

PR #2765, opened 2022-08-23
Boltex/issue2764

### LINKED ISSUE #2764: Some default commands (like diff marked nodes) do not create undo beads.
Some 'default' commands. (common commands enabled by default on a fresh leo install - not weird commands from 3rd party plugins) do not create undo beads. 

I'll try to fix the most I can before 6.6.4 release :)

boltex:
Fixes #2764 All exposed commands in leoCompare.py now generate proper undo beads when executed. (also cleaned up  minor typing issue in qt_frame.py and a parameter passed wrong in 'insertHeadlineHelper' insert-node command)


edreamleo (approved):
Thanks for this. I'll merge later today after I merge the ekr-attic branch.

PR #2773, opened 2022-08-25
Felix-server4

boltex:
Fixed leoserver "insert_named_node" method that broke during well-intended typing ;)


edreamleo (approved):
My bad. Thanks for the correction.

PR #2815, opened 2022-09-05
Fix user attributes

spikethehobbitmage:
Leo supports storing plain strings as user attributes if the key starts with 'str_', but there is a quoting bug that results in a corrupted file if the value contains the '"' character. This patch causes Leo to use correct quoting for such values.

Compatibility note: Affected files will be readable by unpatched versions of Leo but won't be writable. Users not using this feature are unaffected.

This patch also sorts user attributes by key name, otherwise the order isn't stable from one save to another and causes noise when using version control. This change is fully backward compatible.


edreamleo (approved):
Thanks for this fix.

PR #2816, opened 2022-09-05
fix Declutter REPLACE-REST

spikethehobbitmage:
Fixes a crash bug when using Declutter's REPLACE-REST feature.


spikethehobbitmage:
rebased to current devel

edreamleo (approved):

PR #2817, opened 2022-09-05
fix crash on invalid declutter rule

spikethehobbitmage:
Fix crash on reading a file with an invalid declutter rule.


edreamleo (approved):
Again, thanks for this fix.

PR #2818, opened 2022-09-05
Support JSON encoded user attributes

spikethehobbitmage:
This patch adds a feature to support JSON encoded user attributes by using a 'json_' prefix, similar to how plain strings can currently be stored using a 'str_' prefix.

Both read and write code fall back to the standard pickle routines if JSON encode/decode fail for any reason.

Compatibility: Older versions of Leo will not be able to read files that use this feature. Files not using this feature are not affected.

### DIFF
--- a/leo/core/leoFileCommands.py
+++ b/leo/core/leoFileCommands.py
@@ -203,6 +203,14 @@ class FastRead:
         if attr.startswith('str_'):
             if isinstance(val, (str, bytes)):
                 return g.toUnicode(val)
+        # Support JSON encoded attributes
+        if attr.startswith('json_'):
+            if isinstance(val, (str, bytes)):
+                try:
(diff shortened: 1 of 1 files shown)


edreamleo (approved):

edreamleo:
@spikethehobbitmage After merging, mypy makes the following complaint:
```console
leo\core\leoFileCommands.py:213: error: On Python 3 formatting "b'abc'" with "{}" produces "b'abc'", not "abc"; use "{!r}" if this is desired behavior  [str-bytes-safe]
```
Please fix this so that mypy is happy.

spikethehobbitmage:
Fixed, but it looks like I'll have to re-issue the pull request. Should be up momentarily.

edreamleo:
@spikethehobbitmage Yes. A new on-line PR is perfectly fine :-)

Issue #2827, opened 2022-09-09
Coverage tests for qt_text.py


edreamleo:
PR #2839 (closed) contains hopeless code.
PR #2840 is the second try. It will follow the Ahas and ground rules below:

**Summary**: Testing Qt code is easier than I ever imagined!

**Aha!** The TestQtGui class *already* creates a fully-functional LeoQtGui instance!
- TestQtGui.setUpClass calls leoTest2.create_app(gui_name='qt'), which creates the gui for *all* the tests.
- In other words, the heavy lifting (time cost) has already been paid!

**Ground rule**: No changes to LeoQtGui are allowed, including new tests on g.unitTesting.
In fact, several existing tests on g.unitTesting may disappear.

**Aha!**: Idle time will never happens during unit tests! Qt will (probably) *never* redraw the screen!

**Aha!**: Tests of interaction may *force* idle time (by calling a QApplication method) after creating a QKeyEvent.

PR #2839, opened 2022-09-12
PR for #2827: qt tests

edreamleo:
See #2827.


edreamleo:
I am closing this issue, for the reasons given in #2827. Hacks such as in the PR would do far more harm than good.

PR #2876, opened 2022-09-20
Added type annotations to freewin.py.

tbpassin:
pylint: 10/10
mypy: no issues


edreamleo (approved):

edreamleo:
@tbpassin Thanks for this work!

tbpassin:
My very first try at learning how to to it!

edreamleo:
@tbpassin All went well except that the `@file node` had your local path in it. I have fixed this in devel.

tbpassin:
Good catch!  I imported this file into another output to to the development to avoid those delays on saving with LeoPyRef.  Since it was the same file, I assumed it would be update in LeoPyRef with no problems.  Wrong again!

PR #2913, opened 2022-10-15
For auto-md files, insert space after headline introducer ('###', etc)

tbpassin:
Some MD processors require a space here.  This PR parallels a similar change for VR3.

### DIFF
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -269,28 +269,25 @@ def open_outline(self: Self, event: Event=None) -> None:
 
         c.bringToFront()
         c.init_error_dialogs()
-        ok = False
         if fileName:
             if g.app.loadManager.isLeoFile(fileName):
                 c2 = g.openWithFileName(fileName, old_c=c)
                 if c2:
+                    c = c2  # #2906: Switch c here!
+                    c.init_error_dialogs()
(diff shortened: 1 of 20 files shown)


edreamleo (changes requested):
Thanks Thomas. I'm a bit concerned about adding extra leading whitespace to headlines.  How about this?

`self.put(f"{'#' * level} {p.h.lstrip()}")  # Leo 6.6.4: preserve spacing.`

tbpassin:
I see what you are thinking of - just in case p.h has leading whitespace. Who knows what one or another MD processor would do?  Yes,  I agree.

Why don't you go ahead and make the same change in VR3's code?  I just checked it and it too could end up with an extra space.  The code is at line 3484:  Change

```
headline_str = '## ' + headline
```
to
```
headline_str = '## ' + headline.lstrip()
```

edreamleo:
@tbpassin Please add the call to lstrip to this PR. I'll then approve it. If you like you can make the change to the VR3 plugin in this same PR. Thanks.

tbpassin:
Superceded by PR https://github.com/leo-editor/leo-editor/pull/2918, which adds a space and lstrips() the headline for VR3 and @auto-md files.

edreamleo (commented):
devel already has these changes!

edreamleo:
@tbpassin No problem.

Issue #2911, opened 2022-10-15
Local-to-outline Menu Shows Up In Other Outlines


tbpassin:
I have some outlines that define a custom menu - intended to be available only for that one outline.  They are implemented by a _@menuat_  tree under the outlines's _@settings_ tree.  These menu items launch outline-specific scripts. 

This works nicely **except** that all outlines opened after this outline also display the custom menu in the menu bar.  I expected (and want) that these custom menus would only appear in their own outline.

Would it be feasible to prevent this undesired display of custom menus outside of their own outline?

Example local menu:
```
@settings
    @menuat Help before
        @menu Local
            @item special-action-1
            @item special-action-2
```

edreamleo:
@tbpassin The `@menuat` logic is wretchedly complex. I'll entertain a PR, but I am not interested in doing the work myself.

tbpassin:
I appreciate that!  I spent some time trying to find a starting point for the code.  I was hoping enough would come back to you that you would be able to at least know if it could be done without a complete rewrite.

How about we leave this as "Later".  I may make a run at it, but with the complexity I don't know if I will want to work it all out either.

edreamleo:
@tbpassin The starting point is easy to find: just search Leo's code for menuat. You will find pbc.doMenuat.

edreamleo:
@tbpassin Good luck :-)

tbpassin:
Some progress, and a conundrum.  In the course of collecting settings when an outline is opened, `lm.getPreviousSettings()` is called.   This calls 
``` 
d1, d2 = lm.computeLocalSettings(c,
    lm.globalSettingsDict,
    lm.globalBindingsDict,
     localFlag=True)  # d1 and d2 are copies.
```
**Before** the call, *lm.globalSettingsDict* does not contain any _menus_ setting. **After** the call, it does contain a _menus_ setting, and that setting contains the locally-set menu tree.  The next outline that opens uses an *lm.globalSettingsDict* that still has the (spurious) previous locally-set menu setting.

It looks like the call ```lm.computeLocalSettings()``` has a side effect of changing the 
`lm.globalSettingsDict`.  I don't suppose this was intended, and I don't as yet know how it happens.

edreamleo:
@tbpassin Many thanks for your investigations

It does indeed look suspicious that LM.computeLocalSettings updates LM.globalSettingsDict.

tbpassin:
[...]

PR #2937, opened 2022-10-26
Prevent outline-local menus from being re-created on outlines opened later

tbpassin:
Fixes  Issue https://github.com/leo-editor/leo-editor/issues/2911


edreamleo (approved):
Many thanks for this PR. It can't have been easy.

tbpassin:
No, it wasn't easy at all.  I won't get into the hours involved!  And even now, there are some things that happen that I never really understood.  But at least the changes themselves are fairly easy to understand.

This was a case where using clones all in one place for the nodes I worked with really paid off.  Yay Leo!

Issue #2938, opened 2022-10-27
Fix crasher in iSearch_helper


edreamleo:
In iSearch_helper, the second-to-last line has an error: w is used before being defined at the line:  `c, binding='BackSpace', char='\b', w=w)`

To see the bug, do <Alt+i> and then type a character not in the text:

```console
  File "/home/felix/leo-editor/leo/core/leoFind.py", line 2697, in iSearch_helper
    c, binding='BackSpace', char='\b', w=w)
UnboundLocalError: local variable 'w' referenced before assignment
```

tbpassin:
LeoDocs.Leo says 

"    Alt+S isearch-forward"

Bu that activates the "Search" menu instead.

edreamleo:
@tbpassin Right. Alt-I activates isearch-forward. Fixed in a recent ref.

tbpassin:
Not fixed in devel today so far.  This change appears to work:  

```
c, binding='BackSpace', char='\b', w=c.miniBufferWidget  # was w=w
```

edreamleo:
@tbpassin PR #2939 contains an even simpler fix. It is waiting for comments.

PR #2939, opened 2022-10-27
PR for #2938: crash in isearch helper

edreamleo:
See #2938.

### DIFF
--- a/leo/core/leoFind.py
+++ b/leo/core/leoFind.py
@@ -2694,7 +2694,7 @@ class LeoFind:
             g.es(f"not found: {pattern}")
             if not again:
                 event = g.app.gui.create_key_event(
-                    c, binding='BackSpace', char='\b', w=w)
+                    c, binding='BackSpace', char='\b', w=None)
                 k.updateLabel(event)
     #@+node:ekr.20131117164142.16950: *5* find.isearch_state_handler
     def isearch_state_handler(self, event: Event) -> None:  # pragma: no cover (cmd)
--- a/leo/core/leoKeys.py
+++ b/leo/core/leoKeys.py
@@ -3774,7 +3774,7 @@ class KeyHandlerClass:
         ch, stroke = event.char, event.stroke
         if ch in "\n\r":
             return
-        if stroke and not k.isPlainKey(stroke):
+        if ch != '\b' and stroke and not k.isPlainKey(stroke):
             return  # #2041.
         c.widgetWantsFocusNow(w)
         i, j = w.getSelectionRange()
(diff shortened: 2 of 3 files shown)


tbpassin:
When I made the default to be  `w=c.miniBufferWidget`, I found that typing another character made it show up in the minibuffer, which seemed like a good choice since the user can see what was typed.  If the default  is `w = None`, I don't think the user will see any keystroke, will they?  Just because no match was found shouldn't prevent the user from seeing the typed character.

My test was to type a single "!", which was matched somewhere, then a second "!" which was not.  The second "!" appeared in the minibuffer.

edreamleo:
@tbpassin Thanks for your testing. As you say, the PR's present code is not correct.

edreamleo:
@tbpassin Assigning `w=None` works. The culprit, surprisingly, was k.updateLabel.

boltex (approved):
I tried it and it no longer crashes, also, the code looks pretty good to me!

 (although i am not very familiar with the widgets)
  :)

edreamleo:
@boltex Thanks for your testing :-)

tbpassin (changes requested):
This does not fix the original problem.  In `find.iSearch_helper`, a keystroke in the minibuffer that is not matched now does indeed display in the minibuffer, but the helper code still fails with no widget at `c, binding='BackSpace', char='\b', w=w`:

```
  File "c:\Tom\git\leo-editor\leo\plugins\qt_events.py", line 138, in eventFilter
    k.masterKeyHandler(key_event)
  File "c:\Tom\git\leo-editor\leo\core\leoKeys.py", line 3126, in masterKeyHandler
    if k.doMode(event):
  File "c:\Tom\git\leo-editor\leo\core\leoKeys.py", line 3306, in doMode
    handler(event)
  File "c:\Tom\git\leo-editor\leo\core\leoFind.py", line 2725, in isearch_state_handler
    self.iSearch_helper()
  File "c:\Tom\git\leo-editor\leo\core\leoFind.py", line 2697, in iSearch_helper
    c, binding='BackSpace', char='\b', w=w)
UnboundLocalError: local variable 'w' referenced before assignment
```

edreamleo:
@tbpassin The code you show does not exist in devel rev cdecbd4.

boltex:
@tbpassin you're probably testing devel or some other branch, because I confirm what @edreamleo just said.

tbpassin:
Hmm, I had patched in the change to `k.updateLabel` as in the PR.  In the PR when I looked, I only saw listed that one change to the one file, so I didn't change anything else.  Now I see more changes in #2938.  Sorry for causing confusion.

Oh, yes, and changing to `w = None` did fix the crash problem, I just tried it.

Issue #2954, opened 2022-11-10
sort-children Command Jumps To Unpredictable Node After The Sort


tbpassin:
Using the _sort-children_ command, the selected node after the sort sometimes is at some unexpected location, even going so far as the root node of the entire tree at times.  Sometimes it stays on the original parent node, as expected.

The code (for _sort-siblings_) has a comment 

`# Sorting destroys position p, and possibly the root position.`

This would perhaps explain why sorting an already-sorted node didn't change the selected parent.

I suggest that a remedy would be to store the parent node's gnx before the sort, and then use it to find the correct position after the sort.  Is there some reason why this wouldn't be a good idea?

Desired behavior: after the sort, the selected node is the same parent node as before the sort.

tbpassin:
Looking at the code more closely, it uses `p = c.setPositionAfterSort(sortChildren)` to set the final position.  But I don't see how that method can work, because it uses the position **after** the sort whereas we want to get the right node **before** the sort started.

tbpassin:
Trying out using the gnx idea, it does end up with the right node selected after the sort.  Here's what I tested:

Early in the method, before there is any sorting:

`start_gnx = p.v.gnx`

After the children are sorted:

```
    found_gnx = False
    for p1 in c.all_unique_positions():
        if p1.v.gnx == start_gnx:
            found_gnx = True
            break
    if found_gnx:
        if sortChildren:
            p = p1.parent()
        else:
            p = p1
    else:
        p = c.setPositionAfterSort(sortChildren)
```

tbpassin:
I would put in a PR but I'm hoping for some discussion in case there would be some undesired effect I haven't thought of.

edreamleo:
@tbpassin I trust you. Please submit a PR.

tbpassin:
Fixed in PR https://github.com/leo-editor/leo-editor/pull/2955

PR #2955, opened 2022-11-11
For sort-children and sort-siblings commands, make sure after-sort 

tbpassin:
selected node is the same as before-sort.

Fixes Issue https://github.com/leo-editor/leo-editor/issues/2954


edreamleo (approved):
Thanks for the PR.

Issue #2949, opened 2022-11-07
UNL String In Status Bar Does Not Properly Escape Apostrophe


tbpassin:
I copied a node's UNL from the status bar at the bottom of Leo's window.  The UNL had an apostrophe in it.  I pasted it into a node's body.  Only the part of the UNL up to the apostrophe highlighted, and when I CNTL-clicked on the UNL, the correct outline opened but at the top, not at the target node.

Replacing the apostrophe by its URL-encoded value (_%27%_) made everything work, including the navigation to the correct node.

Here is the UNL:

`unl://C:/Tom/git/leo-editor/leo/doc/LeoDocs.leo#Leo's Documentation-->Advanced Topics-->Leo University-->@rst html\leo-university.html-->Lesson 5: c.recursiveImport imports files recursively`

I'm not sure whether the RegEx for identifying a UNL in text should allow for an apostrophe, whether it needs to be escaped for Qt to return the correct anchor string when CTRL-clicked, or whether the downstream processing code doesn't work with the apostrophe.  But there is a disconnect in there somewhere.  I am inclined to think that the code that calculates the UNL should URL-encode the character, but I'm open to other suggestions.

Expectation: A UNL copied from the status bar should navigate correctly when CNTL-clicked in a node body.

It's also possible that some other special character(s) have the same issue.

edreamleo:
@tbpassin Thanks for this report. I'll fix it in a new branch, to be merged into devel after releasing 6.7.1.

tbpassin:
The regex that recognizes the unl needs to have single and double quotes escaped:

`unl_regex = re.compile(r"""\bunl:[^`'")]+""")`

So do the url and gnx regexes. We probably need to use escapes for the quotes consistently.

edreamleo:
@tbpassin Thanks for these researches :-) Please issue a PR when you are ready.

edreamleo:
Closed via PR #2964.

PR #2956, opened 2022-11-12
URL encode UNLs, except do not change "-->" to "--%3E" 

tbpassin:
(other code may expect "-->", e.g., persistence domain controller).  This will eliminate problems with `"'"` characters in node headlines.

Fixes Issue https://github.com/leo-editor/leo-editor/issues/2949.


tbpassin:
Hold off on merging this.  Leo seems to be forgetting its sizes and open files after it has closed.  Would that be in the persistent data controller?  That's one of the few places that uses UNLs.

tbpassin:
We should scrap this PR.  I will submit a simpler one that only escapes apostrophes and does not seem to have any side effects like this one has.

Issue #2957, opened 2022-11-12
Rewrite the sort commands and related undo code


edreamleo:
See PR#2958. PR #2955 introduces new problems into the sort commands.

The sort-children and sort-siblings commands and their related undo/redo logic should be completely rewritten.  I alone am responsible for the present mess.

edreamleo:
Closed via PR #2958.

PR #2958, opened 2022-11-12
PR for #2957: Rewrite sort-children

edreamleo:
See #2957.

- [x] Split the two sort commands into entirely separate pieces, eliminating the evil 'sortChildren' kwarg.
- [x] Rewrite sort-children.
- [x] Rewrite sort-siblings.
- [x] Retire beforeSort* methods.
- [x] Revise undo/redo/Sort* methods.
- [x] Retire setPositionAfterSort.
- [x] Move dump methods into LeoUnitTest class.
- [x] Create leo/unittests/commands/test_outlineCommands.py.
- [x] Remove strange unit-test-related assertion in undo code.
- [x] Create unit tests for sort-children and sort-siblings.

### DIFF
--- a/leo/commands/commanderOutlineCommands.py
+++ b/leo/commands/commanderOutlineCommands.py
@@ -1767,7 +1767,7 @@ def sortChildren(
     reverse: bool=False,  # Not used in Leo's core.
 ) -> None:
 
-    """Sort the siblings of p."""
+    """Sort the children of p."""
     c, u = self, self.undoer
     if not p:
         p = c.p
--- a/leo/commands/commanderOutlineCommands.py
+++ b/leo/commands/commanderOutlineCommands.py
@@ -1759,21 +1759,42 @@ def toggleSparseMove(self: Self, event: Event=None) -> None:
 #@+node:ekr.20080425060424.1: ** c_oc.Sort commands
 #@+node:ekr.20050415134809: *3* c_oc.sortChildren
 @g.commander_command('sort-children')
-def sortChildren(self: Self, event: Event=None, key: str=None, reverse: bool=False) -> None:
-    """Sort the children of a node."""
-    # This method no longer supports the 'cmp' keyword arg.
-    c, p = self, self.p
-    if p and p.hasChildren():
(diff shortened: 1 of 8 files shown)


tbpassin:
sort-children and sort-siblings both leave the after-sort focus on the initial before-sort node.  I'm happy with that.

tbpassin:
There is a `u.afterSortSiblings`, but I don't see a matching `u.beforeSortSiblings`.  Same for sorting children.

Is that intentional, or an oversight?

edreamleo:
@tbpassin Yes, retiring u.beforeSort* was intentional, as indicated in the PR's checklist.

If I were writing the Undoer class now I would do the following:

- Eliminate most u.before* methods.
  The u.after* methods can contain the "before" and "after" snapshots.
- Eliminate most Undoer ivars.
   Instead, the undo/redo methods could get their data from the g.Bunches the after methods create.

tbpassin:
Thanks.  Also, I noticed in `u.afterChangeBody` an admonition not to use a ternary operator:

```
    # Careful: don't use ternary operator.
    if w:
        bunch.newSel = w.getSelectionRange()
    else:
        bunch.newSel = 0, 0  # pragma: no cover
```

Just for my education, why is that?  I thought that the first part of the ternary expression would not be executed if the condition were False, so why would there be a problem with the following ternary?

```
bunch.newSel = w.getSelectionRange() if w else 0, 0
```

boltex:
@tbpassin maybe because the last comma at the end of the line, **to mean the set 0,0**  makes it indeterminate if the bunch._newSel (which is a set of 2 numbers) is first the ternary, then the ",0"  as the last number of the set. or both at once as the 'else' of the ternary " 0 , 0" 

(sorry if this is unclear, i didnt have cofee today yet)

edreamleo:
@tbpassin Félix is correct: it is a bit tricky to use the ternary operator with tuples. But this should work:

    bunch.newSel = w.getSelectionRange() if w else (0, 0)

Perhaps the comment is a holdover from an ancient confusion of mine, which I finally got clear about. For *one-element tuples only*, a trailing comma is required to denote the tuple, like this `(0,)` or this `tuple(0)`.

Issue #2959, opened 2022-11-13
Fix problems with ftm.set_widgets and sftm.init_widgets


edreamleo:
From an email from Félix. There are several problems with FindtabManager.set_widgets_from_dict and StringFindTabManager.init_widgets:

-  sftm.init_widgets does not handle the "File Only" option.
- Neither method defines "val" after the comment `# Ensure one radio button is set.`
- Both methods should fall back to "Entire Outline" instead of "File Only"!

edreamleo:
Closed via PR #2961.

PR #2961, opened 2022-11-13
PR 2 for #2959

edreamleo:
See #2959.

### DIFF
--- a/leo/plugins/qt_frame.py
+++ b/leo/plugins/qt_frame.py
@@ -1194,7 +1194,7 @@ class FindTabManager:
         if not find.node_only and not find.suboutline_only and not find.file_only:
             w = self.radio_button_entire_outline
             w.toggle()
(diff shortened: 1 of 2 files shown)


leo/plugins/qt_frame.py
```diff
@@ -1194,7 +1194,7 @@ def radio_button_callback(
         if not find.node_only and not find.suboutline_only and not find.file_only:
             w = self.radio_button_entire_outline
             w.toggle()
-    #@+node:ekr.20210923060904.1: *3* ftm.init_widgets_from_dict (new)
+    #@+node:ekr.20210923060904.1: *3* ftm.init_widgets_from_dict
```
boltex, line 1197: For consistency: Node is called 'ftm.init_widgets_from_dict' while the method is called 'set_widgets_from_dict' .

PR #2982, opened 2022-11-29
PR: Fix minor flake8 complaints

edreamleo:
- [x] Fix straightforward flake8 warnings.
- [x] Add flake8 suppressions to setup.cfg for difficult/dangerous flake8 warnings.

### DIFF
--- a/leo/commands/checkerCommands.py
+++ b/leo/commands/checkerCommands.py
@@ -341,9 +341,6 @@ class Flake8Command:
         """Run flake8 on all files in paths."""
         c, tag = self.c, 'flake8'
         for root in roots:
-            # @noflake8
-            if any(z.strip().startswith('#noflake8') for z in g.splitLines(root.b)):
-                continue
             path = g.fullPath(c, root)
             if path and os.path.exists(path):
                 g.es_print(f"{tag}: {path}")
--- a/leo/commands/commanderEditCommands.py
+++ b/leo/commands/commanderEditCommands.py
@@ -877,13 +877,14 @@ def find_bound_paragraph(c: Cmdr) -> Tuple[str, List[str], str]:
     # scan backward, adding trailing lines of head to ins.
     if insert_lines and not startsParagraph(insert_lines[0]):
         n = 0  # number of moved lines.
-        for i, s in enumerate(reversed(head_lines)):
+        for s in reversed(head_lines):
             if ends_paragraph(s) or single_line_paragraph(s):
                 break
             elif startsParagraph(s):
                 n += 1
                 break
-            else: n += 1
+            else:
+                n += 1
         if n > 0:
             para_lines = head_lines[-n :] + para_lines
             head_lines = head_lines[: -n]
--- a/leo/commands/convertCommands.py
+++ b/leo/commands/convertCommands.py
@@ -177,7 +177,8 @@ class To_Python:  # pragma: no cover
             elif self.match(lines, i, '\n'):
                 i += 1
                 i = self.removeExcessWsFromLine(lines, i)
-            else: i += 1
+            else:
+                i += 1
     #@+node:ekr.20150514063305.141: *5* removeExessWsFromLine
     def removeExcessWsFromLine(self, lines: List[str], i: int) -> int:
         assert(i == 0 or lines[i - 1] == '\n')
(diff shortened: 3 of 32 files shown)


edreamleo:
@vitalije I am unsure of the proper annotation for the newly created defs in leoFileCommands.py.

mypy complains about the 'x' arg to the new findNode and pv functions. Do you have any suggestions?

vitalije:
I can't find the definition of `pv`. And `findNode`' takes an argument of type `str` and returns `VNode` instance.

vitalije:
Oh, I've found it. It is the same as `findNode`.

edreamleo:
Changing the annotation of 'x' to 'str' in the two functions generates these complaints:
```console
leo\core\leoFileCommands.py:1090: error: Argument 1 to "findNode" has incompatible type "VNode"; expected "str"  [arg-type]
leo\core\leoFileCommands.py:1091: error: Argument 1 to "findNode" has incompatible type "VNode"; expected "str"  [arg-type]
leo\core\leoFileCommands.py:1199: error: Argument 1 to "pv" has incompatible type "VNode"; expected "str"  [arg-type]
leo\core\leoFileCommands.py:1200: error: Argument 1 to "pv" has incompatible type "VNode"; expected "str"  [arg-type]
```
Changing the annotations of 'x' to 'VNode' generates these complaints:
```
leo\core\leoFileCommands.py:1086: error: No overload variant of "get" of "dict" matches argument types "VNode", "VNode"  [call-overload]
leo\core\leoFileCommands.py:1086: note: Possible overload variants:
leo\core\leoFileCommands.py:1086: note:     def get(self, str) -> VNode
leo\core\leoFileCommands.py:1086: note:     def [_T] get(self, str, Union[VNode, _T]) -> Union[VNode, _T]
leo\core\leoFileCommands.py:1196: error: No overload variant of "get" of "dict" matches argument types "VNode", "VNode"  [call-overload]
leo\core\leoFileCommands.py:1196: note: Possible overload variants:
leo\core\leoFileCommands.py:1196: note:     def get(self, str) -> VNode
leo\core\leoFileCommands.py:1196: note:     def [_T] get(self, str, Union[VNode, _T]) -> Union[VNode, _T]
```

vitalije:
Correct annotation for argument is `str` and return value is of type `VNode`. The type checker complains due to the fact that the algorithm uses Python ability to change the type of the members in the list. Initially, all `vnodes` restored from the database have their `parents` and `children` fields set to list of strings (only gnx is stored instead of the `vnode` instances). In the above mentioned lines, those fields ( `children` and `parents`) are transformed from the lists of strings into the list of `VNodes`. However, type checker can't understand this transformation and it warns that we're trying to use `vnode` instances as arguments to `findNode` which expects only strings. 

The code can be changed to keep the track of children and parent `gnx` keys during loading in the separate dictionary and not directly in the fields of `VNode` instances, but IMO this would unnecessarily complicate the code just to make type checker happy. I don't know if it is possible to silence these few warnings using some comments in the above lines or not. If not I suggest just to leave the annotation to `Any`. 

Assuming that no other thread is reading from the `gnxDict` during the execution of these methods, the code is safe. At the end of the method, there will be no `VNode` instances with `children` and `parents` fields that contain anything else but `VNode` instances. Only during the execution of these methods, temporarily, they will contain strings in those fields.

vitalije:
Just to illustrate what is happening in the `retrieveVnodesFromDb` method:

https://github.com/leo-editor/leo-editor/blob/01dc721d2680ce4b3aef045532c63b5ff73dabd8/leo/core/leoFileCommands.py#L1058-L1071

Variables `children` and `parents` are retrieved from the database as a strings (line 1059). They contain just `gnx` fields of each child/parent separated by space character. At lines 1067 and 1068, fields of newly created `VNode` instance `v.parents` and `v.children` are set to list of strings. Then, after all `Vnode` instances are recreated, hence all of them are now values of the `fc.gnxDict`, we can replace strings with the corresponding `VNode` instances in the lines 1090-1091

https://github.com/leo-editor/leo-editor/blob/01dc721d2680ce4b3aef045532c63b5ff73dabd8/leo/core/leoFileCommands.py#L1090-L1091

The flow of the operation consists of two separated but tightly connected phases. In the first phase, lines 1058-1071 we're just creating vnodes and after we have created all of the nodes, in the second phase we restore the links between them as they were when the outline was saved. It is necessary for the first phase to be completed before the second phase starts. Otherwise, we could bump into the problem trying to link not yet existing nodes.

Although the type checker rightly complains, the code is perfectly legal Python code.

edreamleo:
@vitalije Thanks for your comments. I'll add the correct annotation, with a `# type:ignore` telling mypy not to worry.

PR #2986, opened 2022-12-04
PR: Fix problems with read-ref-file

edreamleo:
The read-ref-file does not work for me. I do not understand the code in detail.

- [x] Prevent a crash in read-ref-file.
- [x] Suppress mypy complaints.


vitalije:
In order to use this command one must have reference file set beforehand. 
Good catch.  It could be quite dangerous to run it when no reference file has been set because it deletes all of the top level nodes above the reference file marker node (in case there is no such node, it means the entire outline will be deleted).
It seems I'm likely the only one using these commands.

PR #2988, opened 2022-12-05
Update viewrendered3.py

wangzhaohe:
Add asciidoctor_icons config `@string vr3-asciidoctor-icons=''`

Valid values for the icons attribute are as follows:

image (or empty)

- Icons resolve to image files in the directory specified by the iconsdir attribute.

font

- Icons are loaded from an icon font (like Font Awesome). Not all backends support this mode, such as DocBook.

### DIFF
--- a/leo/plugins/viewrendered3.py
+++ b/leo/plugins/viewrendered3.py
@@ -63,6 +63,7 @@ Previous Recent Changes
 ========================
 For Asciidoctor only, new setting to suppress the footer inserted by the
 Asciidoctor by default (``@bool vr3-asciidoctor-nofooter = True``).
+Asciidoctor by default (``@string vr3-asciidoctor-icons=''``).
 
 New minibuffer commands *vr3-freeze* and *vr3-unfreeze*.
 Improved detection of the notebook URL in *@jupyter* nodes.  The URL no longer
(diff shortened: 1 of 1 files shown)


tbpassin:
1. Please correct line 188 to read:
```
   "@string vr3-asciidoctor-icons", "''", "font, image", "Whether to use icons (``asciidoctor`` only)"
```
2. If you specify _image_, don't you need to be able to specify an optional image directory as well?

wangzhaohe:
@tbpassin Your are right. I found that you have corrected line 188 like bellow. Thanks!

```python
"@string vr3-asciidoctor-icons", "", "'', font, image", "Whether to use icons (``asciidoctor`` only)"
```

wangzhaohe:
> If you specify image, don't you need to be able to specify an optional image directory as well?

@tbpassin Yes. An `:imagesdir: image_path_here` should be added.

But there is a little puzzle that different  .leo file will have different `:imagesdir:`. So perhaps a base imagesdir should be added. Then all image files should be in the same directory tree. Otherwise the user should still assign `:image_path:` or full image file path in each leo body text (for vr3 render view).

I wonder whether the `:imagesdir:` setting in MyLeoSettings.leo will break down the default `./` relative path. For example:

```
image:: ./an_image.png[]
```

So I think it should be tested then get the answer.

tbpassin:
I did talk about an images directory, though I understood that it meant the directory that icon images will be found.  Since you did not propose a setting for a directory, I thought you intended for asciidoctor to use its own default one.  I did see in the docs you linked to that the directory could be a relative one, but I don't know whether that would be relative to the file being processed or to the asciidoctor directory.

Once you become certain what is needed here, you can propose something, if you would.  My own experience with Asciidoctor is basically none, so it's hard for me to offer specific help.

wangzhaohe:
@tbpassin Add imagesdir attribute support.

<https://github.com/leo-editor/leo-editor/pull/2993>

PR #2993, opened 2022-12-06
Update viewrendered3.py

wangzhaohe:
Add `imagesdir` attribute support.

<https://docs.asciidoctor.org/asciidoc/latest/macros/images-directory/#imagesdir-attribute-syntax>

>The value of imagesdir can be an absolute path, relative path or URL. By default, the imagesdir value is empty. That means the images are resolved relative to the document. If an image macro’s target is an absolute path or URL, the value of imagesdir is not added to the target path.


tbpassin (commented):
Looks all right to me.

edreamleo:
@tbpassin *Please* assign a milestone for all issues and PRs. Otherwise I can go days without noticing, as in this case.

edreamleo (approved):

tbpassin:
I would but this one wasn't mine. It comes from @wangzhaohe.

Issue #2991, opened 2022-12-05
Check for ".pyw" files when checking for ".py" ones.


tbpassin:
Leo's code base has a number of cases where it checks to see if a string ends in ".py".  Most of these ought also to check for ".pyw", since the intent is probably to see if a python file is involved.

These cases should be individually checked and corrected if needed..

edreamleo:
Closed via PR #2999.

PR #2999, opened 2022-12-09
Include "pyw" in checks for python file.

tbpassin:
pytest:

```940 passed, 1 warning in 14.68s```

Resolves Issue https://github.com/leo-editor/leo-editor/issues/2991.


edreamleo (changes requested):
Please use `s.strip().endswith(('.py', '.pyw'))` instead of `s.strip().split('.')[-1] in ('py', 'pyw')`

tbpassin:
I've not seen that construction, it's new to me.  Very convenient, though! OK, I'll go through and make the changes.

PR #3018, opened 2022-12-30
Add asciidoctor_diagram showing support in vr3 panel

wangzhaohe:
A support to write some diagrams with leo then show in vr3 panel.

image

Save vr3_adoc.adoc & vr3_adoc.html in getNodePath

Advantage 1: Can show asciidoctor_diagram in vr3 panel

Advantage 2: Can run asciidoctor command manually without copy vr3_adoc.adoc from home directory to getNodePath. Because of some relative image path need getNodePath.

      `asciidoctor -r asciidoctor-diagram -r asciidoctor-pdf -b pdf xxx.adoc`
  
      diagram example:
          [plantuml, target=diagram-classes, format=png]
          ....
          class BlockProcessor
          class DiagramBlock
          class DitaaBlock
          class PlantUmlBlock

          BlockProcessor <|-- DiagramBlock
          DiagramBlock <|-- DitaaBlock
          DiagramBlock <|-- PlantUmlBlock
          ....

Disadvantage: Will Save vr3_adoc.adoc & vr3_adoc.html and dynamic image in each getNodePath.

@tbpassin Would you please consider to save  vr3_adoc.adoc & vr3_adoc.html in each getNodePath? It destroy your design which save them in home directory.

### DIFF
--- a/leo/core/LeoPyRef.leo
+++ b/leo/core/LeoPyRef.leo
@@ -3956,6 +3956,7 @@ def batch_rule8a(colorer, s, i):
 <t tx="ekr.20221201043917.1">@language batch
 echo off
 cd C:\Repos\leo-editor
+echo test-leo
 python -m unittest
 </t>
 <t tx="ekr.20221201080228.1">@language python
@@ -3983,6 +3984,7 @@ black.main()
 <t tx="ekr.20221204070905.1"></t>
 <t tx="ekr.20221204071056.1">@language batch
 echo off
+echo pylint-leo
 time /T
 call python -m pylint leo --extension-pkg-allow-list=PyQt6.QtCore,PyQt6.QtGui,PyQt6.QtWidgets %*
 time /T
@@ -3994,38 +3996,33 @@ cd c:\Repos\leo-editor
 rem See leo-editor/.mypy.ini for exclusions!
 rem Always use the fast (official) version of mypy.
 
+echo mypy-leo
 python -m mypy --debug-cache leo %*
 </t>
 <t tx="ekr.20221204071220.1">@language batch
 echo off
+cls
 cd c:\Repos\leo-editor
-python -m unittest ^
-    leo.unittests.core.test_leoAtFile.TestFastAtRead.test_doc_parts ^
-    leo.unittests.core.test_leoAtFile.TestFastAtRead.test_html_doc_part ^
-    leo.unittests.core.test_leoAtFile.TestFastAtRead.test_verbatim ^
-    leo.unittests.core.test_leoAst.TestOrange.test_verbatim ^
-    %*
+echo test-one-leo: test_cursesGui2
+call python -m unittest leo.unittests.test_plugins.TestPlugins.test_cursesGui2 %*
 </t>
 <t tx="ekr.20221204071554.1">@language batch
 echo off
 cls
 cd C:\Repos\leo-editor
-echo reindent
-call reindent leo
-echo beautify
+
+echo full-test-leo
+call reindent-leo.cmd
 call beautify-leo.cmd
-echo unittests
-call python -m unittest
+call test-leo.cmd
 echo.
-echo mypy
 call mypy-leo.cmd
 echo.
-echo flake8
-call python -m flake8
-echo pylint
+call flake8-leo.cmd
 call pylint-leo.cmd
 echo.
-echo Done!</t>
+echo Done!
+</t>
 <t tx="ekr.20221204072154.1">@language batch
 echo off
 cd c:\Repos\leo-editor
@@ -4034,6 +4031,8 @@ reindent leo
 <t tx="ekr.20221204072456.1">@language batch
 echo off
 cd C:\Repos\leo-editor
+
+echo beautify-leo
 call python -m leo.core.leoAst --orange --recursive leo\core
 call python -m leo.core.leoAst --orange --recursive leo\commands
 call python -m leo.core.leoAst --orange --recursive leo\plugins\importers
@@ -4043,6 +4042,8 @@ call python -m leo.core.leoAst --orange --recursive leo\plugins\writers
 echo off
 cd c:\Repos\leo-editor
 rem: See leo-editor/setup.cfg for defaults.
+
+echo flake8-leo
 python -m flake8 %*
 </t>
 </tnodes>
--- a/leo/plugins/viewrendered3.py
+++ b/leo/plugins/viewrendered3.py
@@ -2304,6 +2304,7 @@ class ViewRenderedController3(QtWidgets.QWidget):
         self.asciidoctor_suppress_footer = c.config.getBool('vr3-asciidoctor-nofooter', default=False)
         self.asciidoctor_icons = c.config.getString('vr3-asciidoctor-icons') or ''
         self.asciidoctor_imagesdir = c.config.getString('vr3-asciidoctor-imagesdir') or ''
+        self.asciidoctor_diagram = c.config.getBool('vr3-asciidoctor-diagram', default=False)
 
         self.external_editor = c.config.getString('vr3-ext-editor') or ''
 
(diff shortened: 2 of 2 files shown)


wangzhaohe:
@tbpassin I am not familiar with git. So I did not changed the LeoPyRef.leo contents although I change the viewrender3.py by LeoPyRef.leo edited. So I think you can ignore the LeoPyRef.leo file.

edreamleo (changes requested):
@wangzhaohe Thanks for this PR. I don't want to change LeoPyRef.leo this way.

@tbpassin Thomas, I would prefer that the PR comes from you. Are you willing?

tbpassin:
I'm willing to make the changes and submit a PR once I understand them and am sure it's a solution I'm agreeable to.  I haven't had a chance to look yet.  I do know that I recently had to change where temporary files were stored because of permission problems on Linux, and I _think_ that was in connection with VR3/asciidoc.  So I will have to check that, and also how asciidoctor is invoked for these diagrams, and what to do if other asciidoc engines do not support this kind of diagram.

It may be some little time before I can get that all done.

@wangzhaohe, where can I get a look at your proposed code changes?

wangzhaohe:
@tbpassin The fork is here.

<https://github.com/wangzhaohe/leo-editor/commit/9f587bf68d5f5abccb72a94a69238b6c03f6186c>

And with another plantuml example:

```
[plantuml, target=diagram-formula, format=png]
....
:<math>int_0^1f(x)dx</math>;
:<math>x^2+y_1+z_12^34</math>;
note right
Try also
<math>d/dxf(x)=lim_(h->0)(f(x+h)-f(x))/h</math>
<math>P(y|bb"x") or f(bb"x")+epsilon</math>
end note
....
```

image

tbpassin:
I created a branch for this work in my own repo and copied the revised vr3.py file to it.  I set the variable _vr3-asciidoctor-diagram_ to `True`.  I installed the plantuml gem.  The diagram did not display an image.  Here is the error message:

```
asciidoctor: FAILED: 'asciidoctor-diagram' could not be loaded
```

I see from the asciidoctor-plantuml site that a Tomcat server is supposed to be running.  I have not set that up.  Without a local server, where is asciidoctor-plantum supposed to find a server?

wangzhaohe:
@tbpassin My environment is bellow, please have a reference. Will it do some help?

```
> asciidoctor -v
Asciidoctor 2.0.18 [https://asciidoctor.org]
Runtime Environment (ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin21]) (lc:UTF-8 fs:UTF-8 in:UTF-8 ex:UTF-8)
```

***

<https://github.com/asciidoctor/asciidoctor-diagram/issues/238> says:

> asciidoctor-diagram requires asciidoctor >= 1.5.7. The gem spec for diagram has been updated to reflect this, but I haven't made a release with that change yet. Sorry for the inconvenience. Installing a more recent version of asciidoctor should fix the problem for you.

***

I only install asciidoctor-diagram, not install asciidoctor-plantuml

```
https://docs.asciidoctor.org/diagram-extension/latest/

[source,console,]
$ gem install asciidoctor-diagram

....
Fetching asciidoctor-diagram-2.2.3.gem
Fetching asciidoctor-diagram-ditaamini-1.0.3.gem
Fetching asciidoctor-diagram-plantuml-1.2022.12.gem
Successfully installed asciidoctor-diagram-ditaamini-1.0.3
Successfully installed asciidoctor-diagram-plantuml-1.2022.12
Successfully installed asciidoctor-diagram-2.2.3
Parsing documentation for asciidoctor-diagram-ditaamini-1.0.3
Installing ri documentation for asciidoctor-diagram-ditaamini-1.0.3
Parsing documentation for asciidoctor-diagram-plantuml-1.2022.12
Installing ri documentation for asciidoctor-diagram-plantuml-1.2022.12
Parsing documentation for asciidoctor-diagram-2.2.3
Installing ri documentation for asciidoctor-diagram-2.2.3
Done installing documentation for asciidoctor-diagram-ditaamini, asciidoctor-diagram-plantuml, asciidoctor-diagram after 0 seconds
3 gems installed
....
```

***

This is my install history

image

tbpassin:
Oh, I see.  You had (or I mistakenly thought you had) a link to plantuml, so that's what I installed.  OK, with the correct gem installed, your two example diagrams do get displayed.  That's a good start.

Next, I will need to:

1. Review the saved file locations;
2. Remove the explanatory text you put in triple quotes and add some material about this capability in the docstring;
3. Add the new settings to the docstring;
4. Figure out if `' -r asciidoctor-diagram'` should still be added to the asciidoctor invocation if the node(s) do not have any diagrams.  If not, we will need to check whether there are any diagrams and only use the extension when there are;
5. See if there is a way to check asciidoctor for the presence of the extension in case it is not installed but the setting asks for it to be used.

tbpassin:
By the way, in the future if you want to make changes to an external @file that is in PyLeoRef - like viewrendered3.py - please commit only the external file but not PyLeoRef.  PyLeoRef does not actually contain your changes - it loads the content of the external file when it is opened in Leo, so there should be nothing to commit.

wangzhaohe:
> By the way, in the future if you want to make changes to an external @file that is in PyLeoRef - like viewrendered3.py - please commit only the external file but not PyLeoRef. PyLeoRef does not actually contain your changes - it loads the content of the external file when it is opened in Leo, so there should be nothing to commit.

I got it, thanks for your guidance.

edreamleo:
@tbpassin @wangzhaohe I am going to close this PR. I'll wait for Thomas to issue his own PR.

Feel free to continue discussions here if you like.

tbpassin:
I will continue to work on this.   @wangzhaohe, thank you for making me aware of this capability for constructing diagrams.  It's also a way to render equations, for which I hadn't found a standard way with asciidoc.

tbpassin:
There is an extension for Sphinx that allows something similar, but with Mermaid diagrams instead of PlantUML: [Sphinxcontrib-mermaid](https://sphinxcontrib-mermaid-demo.readthedocs.io/en/latest/).  I have not tried it, and since it is specific to Sphinx it will probably not work with VR3, which uses docutils.

tbpassin:
Trying this out on Linux, asciidoctor failed to render the diagrams with an error message that GraphViz is needed.  I happened to have that on my Windows machine, so everything worked there.  On my Linux VM, I then installed Graphviz, but asciidoctor thinks that the dot executable (part of GraphViz) is at a different location than where it is actually installed.  

I will have to work out how to configure asciidoctor to use the right location.  If that cannot be done, we may need to file a bug with asciidoctor.

wangzhaohe:
I just learn asciidoc for a few days, and don't learn asciidoc deeply. So I am willing to do some tests if I can. And there must have many work waiting to do for you. I can do something, please tell me.

tbpassin:
Here is how we can check to see if asciidoctor-diagram is installed:

```
from subprocess import run

def check_gems(gem:str, encoding:str = 'utf-8') -> bool:
    """Check if a particular ruby gem is installed."""
    cmd = f'gem list {gem}' 
    proc = run(cmd, shell = True, capture_output = True)
    installed = gem.encode(encoding) in proc.stdout
    return installed

has_adoc_diagram = check_gems('asciidoctor-diagram')
g.es(has_adoc_diagram)   # or print()
```

This works on Linux and Windows.

wangzhaohe:
Tested! check_gems() works on macOS Monterey Version 12.2.1 too.

```
Leo Log Window
Leo 6.7.2-devel, devel branch, build f1b522eb79
2022-12-29 12:52:34 -0600
Python 3.9.5, PyQt version 5.15.2
darwin
```

tbpassin:
Good to know, thanks!

Earlier today I submitted a PR for the plantuml diagram capability, and it includes the `check_gems()` function.

PR #3027, opened 2023-01-03
Fix command adoc output path changed

wangzhaohe:
The command `adoc` output path is not in a fixed position, perhaps in some condition. Bellow is one output position used to be.

```
adoc: wrote /Users/swot/app/leo-editor/leo/core/xxx.adoc

adoc: wrote /Users/swot/app/leo-editor/xxx.adoc

adoc: wrote /Users/swot/xxx.adoc

adoc: wrote /Users/swot/learning/html/xxx.adoc (in my project directory, wanted)
```

So the xxx.adoc should be found manually, then copy to the project directory because of the asciidoc directive like `include` or `image` needing relative path.

### DIFF
--- a/leo/core/leoMarkup.py
+++ b/leo/core/leoMarkup.py
@@ -260,7 +260,8 @@ class MarkupCommands:
                 i_path = self.filename(p)
                 # #1398.
                 i_path = c.expand_path_expression(i_path)
-                i_path = g.os_path_finalize(i_path)
+                n_path = c.getNodePath(c.p)  # node path
+                i_path = g.os_path_finalize_join(n_path, i_path)
                 with open(i_path, 'w', encoding='utf-8', errors='replace') as self.output_file:
                     self.write_root(p)
                     i_paths.append(i_path)


wangzhaohe:
@tbpassin Did you have meet this phenomenon?

tbpassin:
No, I never tried to insert images or includes in an asciidoc file.  In the last PR, I simply kept the change in the file location that you had put in. If that is not the best place to put those files, you should explain better what is needed.  Usually a relative location for an image file is relative to the directory of the including file.

I don't want to make any changes here until I understand the problem better, and also make sure that any proposed solution will also work with the Asciidoc and Asciidoc3 processors.

Next time, before asking for a PR, would you please raise an issue so we can discuss it and make sure we have a suitable solution?  

In fact, please create an issue for this and please include an example with it so I can reproduce and understand the problem.

wangzhaohe:
All right, I will raise an issue soon.

edreamleo (approved):

wangzhaohe:
@tbpassin I have raise an issue with an example https://github.com/leo-editor/leo-editor/issues/3032

The issue is about an adoc command, not rended in vr3 panel. Please have a check about it. Thanks!

Issue #3045, opened 2023-01-11
Error During Startup - "can not import leo.plugins.writers.plugins"


tbpassin:
I have started getting the following error message during startup:
```
Traceback (most recent call last):

  File "c:\Tom\git\leo-editor\leo\core\leoApp.py", line 2480, in createWritersData
    m = importlib.import_module(f"leo.plugins.writers.{sfn[:-3]}")

  File "C:\Users\tom\AppData\Local\Programs\Python\Python310\lib\importlib\__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)

  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import

  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load

  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked

ModuleNotFoundError: No module named 'leo.plugins.writers.plug'

can not import leo.plugins.writers.plugins
can not import leo.plugins.importers.plug
[
         'C:/Users/tom/.leo/plugins'
]
```

I am able to successfully import these writers plugins in a Python REPL session outside of Leo.  I added a print statement to look at the import paths, and they look normal to me.

I have not noticed any operational faults with saving outlines with external files, but I don't have any `@auto` files.  The readme file in the plugins/writers directory says they apply to `@auto` files.

I'm not sure when this started.  I just noticed it this morning, but it seems to have been in devel before now.

edreamleo:
@tbpassin Thanks for this report. The first and last entries in the traceback shows the probable cause:
```console
File "c:\Tom\git\leo-editor\leo\core\leoApp.py", line 2480, in createWritersData
    m = importlib.import_module(f"leo.plugins.writers.{sfn[:-3]}")
...
ModuleNotFoundError: No module named 'leo.plugins.writers.plug'

can not import leo.plugins.writers.plugins
can not import leo.plugins.importers.plug
[
         'C:/Users/tom/.leo/plugins'
]
```
Presumably the last three characters of the *module* name `leo.plugins.writers.plugins` are being truncated.

This is surprisingly difficult code, but ensuring that sfn endswith ".py" would probably be a good enough guard.

edreamleo:
@tbpassin This issue is my responsibility. I'll submit a PR asap.

tbpassin:
"Presumably the last three characters of the module name leo.plugins.writers.plugins are being truncated."

[...]

PR #3046, opened 2023-01-11
PR for #3045: import crash

edreamleo:
See #3045.

### DIFF
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -2473,7 +2473,7 @@ class LoadManager:
         ):
             for filename in g.glob_glob(pattern):
                 sfn = g.shortFileName(filename)
-                if sfn != '__init__.py':
+                if sfn.endswith('.py') and sfn != '__init__.py':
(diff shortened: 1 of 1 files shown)


tbpassin:
I'm still getting one error message (in Windows, anyway) but no stack trace:

```
can not import leo.plugins.importers.plug
[
         'C:/Users/tom/.leo/plugins'
]
```

Should it be looking there, in my own .leo/plugins directory?

edreamleo:
@tbpassin It's hard to know without seeing the .leo file and your .leo directory.

The message doesn't concern me as long as Leo starts up.

tbpassin:
I did have a _plugins_ directory in the .leo directory, for reasons I barely remember, and when deleted it I stopped getting that message.  When it was present but empty the message was still emitted.

edreamleo:
@tbpassin Thanks for the update. Imo, the remaining message was likely correct. In any case, it was good to add the (necessary) guard.

Issue #3058, opened 2023-01-15
Right-click Context Open With Editor Should Prefer "external-editor" Setting


tbpassin:
Leo has a tree context menu item to edit the selected node in an external editor.  Although there is an _external-editor_ setting, the method to decide what editor to use for the context menu rclick prefers other possibilities.

The method `guessExternalEditor()` resolution order  is

```
        os.environ.get("LEO_EDITOR") or
        os.environ.get("EDITOR") or
        g.app.db and g.app.db.get("LEO_EDITOR") or
        c and c.config.getString('external-editor')
```

So even if you have told Leo to use a particular editor, it will not do if there is an environmental variable _EDITOR_ or _LEO-EDITOR_.  On many Linux systems, _EDITOR_ will be _nano_ by default.  This could be changed but that would require the user to edit some file, like _bashrc_, which is not desirable.  Nano is not what I want to use, and it "opens" in the terminal that opened Leo, where it may not even be seen if the terminal is hidden behind the Leo window as it often is.

I would like `guessExternalEditor()` to first use the _external-editor_ setting.  That is what it's for.  If there is agreement, I will put in a PR for this little change.

edreamleo:
@tbpassin Good catch. Please do submit a PR.

edreamleo:
Closed via PR #3059.

PR #3059, opened 2023-01-16
g.guessExternalEditor() now prefers to use the "external-editor" setting

tbpassin:
Addesses https://github.com/leo-editor/leo-editor/issues/3058


edreamleo (approved):
@tbpassin Thanks for your attention to details like this. I appreciate it.

Issue #3073, opened 2023-01-20
Recreate sources (@rst nodes) for slideshows


edreamleo:
See PR #3083.

The `@rst` nodes are missing for the slides in `leo/doc/html/slides`.

The `_sources` folder in *some* of the slideshow folders contain the intermediate files.

- [x] Delete the "installation" slideshow. It is out of date.
- [x] Review all other slideshows.  They *are* worth keeping!
- [x] Revise `@rst html/slides.html`.
- [x] Use the top-level `_static` folder for all slides.
- [x] Add `@rst` nodes for all slideshows.

tbpassin:
Here is a package (on PyPi) to create slides with RsT:

[rst2html5slides](https://pypi.org/project/rst2html5slides/)

I have not tried it as yet.

edreamleo:
@tbpassin hanks for the link. The problem I have is different: it is recreating rST from the existing .html files.

​There are several possible ways forward, none of them palatable:

A: Forget about creating `@rst` nodes. Just hack the .html files,
  either editing the files externally to Leo or after importing the .html into LeoDocs.leo:
- Change the links to the logo​.
- Add references to custom.css.
- Use a static change date.
- Remove mention of sphinx (since it won't be used in future!).

B: Recreate `@rst` nodes by hand. This might be better, but it might take more work initially.

I am tempted to just trash the slideshows.

tbpassin:
I'd redo them  by hand.  Although trashing them might be the best idea.  OTOH, I'm not sure if the `@slide` production process works with rst3.  That's how I found those use-before-assign bugs, by trying to run rst3 on the slide tree.  That makes me think that the file hasn't been used for some time.  So even if it used to work for making slides it might not any more (although after I fixed the bugs, it did produce an rst output file, so that's good).

edreamleo:
@tbpassin :-) The titles in the slides indicate that they are very old. I've looked through git's history and found no `@rst` nodes in LeoDocs.leo.

The slides aren't great, and the installation slides are obsolete, but I'm not quite ready to give up yet.

Meh. Leo's html importer doesn't work on these slides. See #3079. So that's next.

Once the html importer works I can use search/replace to revise the slides. Yes, it will be a halfway measure, but it might suffice.

edreamleo:
I have found a straightforward way of recreating `@rst` nodes from old .html files.

edreamleo:
Closed vi PR #3083.

PR #3080, opened 2023-01-21
Clean html slides by hand.

edreamleo:
I am closing this PR. It will be much easier to create the missing `@rst` nodes.  See #3073.


edreamleo:
This PR taught me the value of rST. Editing .html files by hand is a bad idea.

PR #3082, opened 2023-01-21
Fixes #3074, exception on startup with --theme= for some themes.

### LINKED ISSUE #3074: New Startup Exception
After updating the to current changeset in devel, I started getting a new error on startup:

```
C:\Users\tom>py-leo-git --theme=tbp_dark %userprofile%\.leo\workbook.leo
Using PYTHONPATH: ;c:\Tom\git\leo-editor

Leo 6.7.2-devel, devel branch, build 1f74c18d50
2023-01-20 09:26:19 -0600
Python 3.10.9, PyQt version 6.4.1
Windows 10 AMD64 (build 10.0.19044) SP0

Uncaught exception in Leo...
  File "c:\Tom\git\leo-editor\leo\plugins\qt_text.py", line 977, in paintEvent
    if w == c.frame.body.widget and \
AttributeError: 'NoneType' object has no attribute 'widget'
```

Leo continues to start up and work, but this is a change.  At the same time, during startup when the session will have many open outlines, parts of Leo panes flash on and off in unusual locations compared with previous versions, where by and large the Leo window was blocked from painting until all outlines were loaded.

tbpassin:
Fixes #3074.

### DIFF
--- a/leo/plugins/qt_text.py
+++ b/leo/plugins/qt_text.py
@@ -974,7 +974,7 @@ if QtWidgets:
                     self.leo_cursor_width = width
                     w.setCursorWidth(width)
 
-            if w == c.frame.body.widget and \
+            if hasattr(c.frame.body, 'widget') and  w == c.frame.body.widget and \
                     c.config.getBool('show-rmargin-guide'):
                 #@+<< paint margin guides >>
                 #@+node:tom.20220423204906.1: *4* << paint margin guides  >>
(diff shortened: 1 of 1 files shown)


edreamleo (changes requested):
Please delete the backslash-newline and simplify the code.

Something like this (not tested):
```python
            if (w == getattr(c.frame.body, 'widget', None)
                and c.config.getBool('show-rmargin-guide')
            ):
```

tbpassin:
I think that the check for ` w == c.frame.body.widget` needs to be there in some form, because this class, _LeoQTextBrowser_, is also used in other places that are not the frame's body widget.  We don't want to try to paint a right hand margin in those cases (or maybe some of them would like it, but not all).

I'll see if the expression can be made more graceful.

tbpassin:
OK, I made the requested change and also took care of most pylint suggestions.

edreamleo (approved):

PR #3094, opened 2023-01-24
Added a short sleep, which allows the splash screen to display visibly…

tbpassin:
… on many Linux systems.


edreamleo (approved):
Short PR, backed with excellent work.

Issue #3088, opened 2023-01-23
LeoFind.py's word-search-forward and word-search-backward are bugged because of small typo.


boltex:
See PR #3098.

Change request_whole_**world**  to request_whole_**word**.

boltex:
While around the same method, the leoFind property **findAllUniqueFlag** is useful, but it's neighbor, **changeAllFlag**, is never used throughout Leo's codebase. :)

boltex:
Other remarks:  
_(sorry in advance if it's me who do not understand the expected behavior. )_

The **findAllUniqueFlag** property which is set to true when starting a find-all-unique-regex command, is not reset to 'False' when starting a regular 'find-all' command...

 (and i'm not sure how it's supposed to playout with the regular **pattern_match** flag from the checkbox in the find panel. I've made a few experiments with and without while using the **find-all-unique-regex** command, but it's still not clear to me what the expected output should be. )

Lastly, after doing a **find-all-unique-regex**, the regular find-all command does not fill up the generated summary node.

edreamleo:
Fixed via PR #3098.

PR #3098, opened 2023-01-24
PR for #3088: find typo

edreamleo:
See #3088.


edreamleo:
@boltex It would be difficult to create a unit test that have found these typos because the typos appear in interactive code.

PR #3111, opened 2023-01-30
Made the SetEncoder support vnodes and positions, and checked json validity for responses.

boltex:
Note : I tested it with leoInteg but I am unable to run the Leo tests suites since the unit-testing system was changed.


boltex:
I finally figured out how to run unittests, (had to switch to windows, it didn't work for me in linux, but i'm a total newb with those things) 

All test pass.

edreamleo (approved):

PR #3123, opened 2023-02-02
Boltex/issue3100

### LINKED ISSUE #3100: Find operations that 'mark' nodes of found or replaced instances should produce undo beads that also undo the marks.
For find/replace/find all/replace all commands

If the 'mark finds/mark changes'  options are checked, then the commands should make undo beads that effectively revert the marks properly. 

A little bit of if/then logic might be needed:
 - make undo groups if it changes AND marks the node (otherwise the simple 'body-change undo' already in place is ok)
 - an instance may stay marked if it was already marked before the find / match, 
 - etc...

I'll gladly make a fix and a PR for this once the find-all command is fixed with PR #3099

boltex:
Fixes undo support for find/replace operations so that they take marking into account.

Also makes find-all and replace-all respect the 'Mark Finds' and 'Mark Changes' options

Fixes #3100

### DIFF
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -1242,7 +1242,8 @@ class Commands:
         return None, None
     #@+node:tbrown.20091206142842.10296: *5* c.vnode2allPositions
     def vnode2allPositions(self, v: VNode) -> List[Position]:
-        """Given a VNode v, find all valid positions p such that p.v = v.
+        """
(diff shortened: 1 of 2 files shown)


boltex:
I ran unit tests and tried out the affected commands also in LeoInteg.

edreamleo (approved):

edreamleo:
Many thanks for this work!

Running `reindent-leo.cmd` fixed on line. No big deal. See ca06cc672.

Issue #3124, opened 2023-02-02
Improve searches when control-clicking a name


edreamleo:
- Control-clicking a class name starting with `_` searches for `def` instead of `class`.
- Always search for the other kind of def if the first search fails.
- Search for `name = ` if first two searches fail.

edreamleo:
Closed via PR #3128.

PR #3128, opened 2023-02-03
PR for #3124: improve ctrl-click searches

edreamleo:
See #3124.

**Summary**

The `find-def` command (and ctrl-clicks):
- searches for `class`, `def` and assignments using whole-word searches.
- *restores* previous search settings only if the search fails.
- *retains* the whole-word search settings that resulted in the match.
- Experiments show that using regex searches would not be worthwhile.
  The gain in accuracy is not worth the extra complexity of the patterns shown to the user.

**Details**

- [x] `find.find_def` searches for `class`, `def` and assignments, using alternative names if the original searches fail.
- [x] Add comment in `find.find_def` about calling sequence.
- [x] Add comment in `g.openUrlOnClick` about calling sequence.
- [x] Remove evil 'do_def' and 'strict' kwargs.
- [x] Remove `find.find_def_strict`. `g.openUrlOnClick now calls `find.find_def`.
- [x] Revise the unit test.

### DIFF
--- a/leo/commands/spellCommands.py
+++ b/leo/commands/spellCommands.py
@@ -484,7 +484,7 @@ class SpellCommandsClass(BaseEditCommandsClass):
             self.handler.add()
         else:
             self.openSpellTab()
-    #@+node:ekr.20150514063305.486: *4* find
+    #@+node:ekr.20150514063305.486: *4* spell-find
     @cmd('spell-find')
     def find(self, event: Event = None) -> None:
         """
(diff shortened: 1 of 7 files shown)


edreamleo:
@boltex This PR has turned into quite a project. I am happy with the results, but it needs more testing before I'll merge it.

I would welcome your comments.

boltex (approved):
I've played with the CTRL+click on URL'S a bunch. I've tried alt+click on classes mentioned here and there in leo's codebase to test that too.

Also tried with complicated regex. (a regex to find urls to test ctrl+click hehe!)

`(http|ftp|https):\/\/([\w_-]+(?:(?:\.[\w_-]+)+))([\w.,@?^=%&:\/~+#-]*[\w@?^=%&\/~+#-])`

I could not get it to bug or give any errors.

edreamleo:
@boltex Many thanks for your testing. The new code deserves to be tested more widely, so I'll merge it now. We should be able to fix any new buglets easily.

PR #3131, opened 2023-02-04
Leo scripts: in "Display function call hierarchy in Leo" move @others…

tbpassin:
… so that it can be called. Use triple quotes for docstring.

Somewhere along the line, the @others line got moved to the bottom, but it needs to be near the top for the script to work.


edreamleo:
@tbpassin *Please* remember to assign a milestone. Otherwise I am likely to overlook the PR.

tbpassin:
You need to use _-dy_, then it works.

```
    try:
        self.scroll_lockout = True
        hbar.setValue(dx)
        vbar.setValue(-dy)  # <=======
    finally:
        self.scroll_lockout = False
```

But what do you have against having scrollbars?

PR #3140, opened 2023-02-09
Pylint - Run Against Current Subtree If No External Files Found

tbpassin:
Now searches for external files up and down from the current node.  If found, runs pylint on all external Python files found.  If not, runs pylint on the tree headed by the current node.

### DIFF
--- a/leo/commands/checkerCommands.py
+++ b/leo/commands/checkerCommands.py
@@ -8,6 +8,7 @@ import os
 import re
 import shlex
 import sys
+import tempfile
 import time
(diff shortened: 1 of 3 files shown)


edreamleo (approved):
@tbpassin Excellent PR. Well done.

edreamleo:
@tbpassin rev a9965df in devel fixes a mypy complaint and removes trailing whitespace in a few lines. No big deal.

In the future, please run the equivalent of `full-test-leo.cmd` before committing.

PR #3142, opened 2023-02-11
leoAst.py: fix print function args; allows beautify-files-diff to succeed

tbpassin:
Also fixes undefined variable bb that prevented exception from being handled corrrectly.

This fixes a misplaced parens in the `print_obj()` method.

### DIFF
--- a/leo/core/leoAst.py
+++ b/leo/core/leoAst.py
@@ -259,7 +259,7 @@ class LeoGlobals:  # pragma: no cover
     #@+node:ekr.20191226175441.1: *3* LeoGlobals.print_obj
     def print_obj(self, obj: Any, tag: str = None) -> None:
         """Print an object."""
-        print(self.to_string(obj, tag))
+        print(self.to_string(obj), tag)
     #@+node:ekr.20220327120618.1: *3* LeoGlobals.short_file_name
     def short_file_name(self, fileName: str) -> str:
         """Return the base name of a path."""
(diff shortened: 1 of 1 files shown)


edreamleo (approved):
@tbpassin Good catch.

edreamleo:
@tbpassin Imo, the correct fix is as follows:
```
    def print_obj(self, obj: Any, tag: str = None) -> None:
        """Print an object."""
        print(self.to_string(obj, tag=tag))
```

edreamleo:
@tbpassin See rev 6f3e9a7 in devel for the new fix.

tbpassin:
OK; I just fixed what I saw, but I don't object.

edreamleo:
@tbpassin No problem. Thanks for finding this bug.

tbpassin:
"Imo, the correct fix is as follows:

    def print_obj(self, obj: Any, tag: str = None) -> None:
        """Print an object."""
        print(self.to_string(obj, tag=tag))
In other words, the call should use a kwarg."

Actually, _obj_ and _tag_  are both positional and kw arguments - see
[PEP-570](https://peps.python.org/pep-0570/):

"If / is not specified in the function definition, that function does not accept any positional-only arguments."

Trying it out in the REPL:
```
>>> def f(msg, tag = ""):
...    print(msg)
...    if tag: print(tag)
...
>>> msg = 'this is a test'
>>> tag = 'the tag'
>>> f(msg)
this is a test
>>> f(msg, tag)
this is a test
the tag
>>> f(msg, tag = tag)
this is a test
the tag
>>>
```

edreamleo:
@tbpassin I agree that `tag` should be a keyword-only arg. The `LeoGlobals` class is a good place to transition to pep8 names and keyword-only args. I'll submit a new PR soon.

PR #3145, opened 2023-02-12
Excessive Save Time (issue 3143): when saving an outline, only clear …

tbpassin:
…node dirty bit if node is actually dirty.


edreamleo (approved):
Thanks, Thomas. Good work.

Issue #3161, opened 2023-02-21
g.handleScriptException crashes


edreamleo:
See PR #3163.

Previously `g.es_exception` returned `Tuple[str, int]`, the filename and index within the file.  Now it doesn't, so `g.handleScriptException` crashes.

This issue must be fixed for Leo 6.7.2.

Imo, `g.handleScriptException` should never have relied on `g.es_exception` this way.

edreamleo:
Closed via PR #3163.

PR #3163, opened 2023-02-21
PR for # 3161: exception in g.handleScriptException

edreamleo:
See #3161.

- [x] Call `g.getLastTracebackFileAndLineNumber` in `g.handleScriptException`.
- [x] Convert `script` and `script1` args to optional kwargs: they are no longer used.
- [x] Remove 'script' and 'script1' args from all calls to `g.handleScriptException`.
- [x] Create a strong new unit test.

### DIFF
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -716,7 +716,7 @@ class Commands:
         except KeyboardInterrupt:
             g.es('interrupted')
         except Exception:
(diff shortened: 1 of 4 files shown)


tbpassin (approved):
These changes look all right to me.  You changed the arguments of g.handleScriptException(() in the only two places that I could find them used, so that change should not cause surprises.

edreamleo:

Issue #3170, opened 2023-02-23
picture_viewer plugin doesn't restore saved scrolling positions properly


edreamleo:
See PR #3171 for my (futile!) experiments.

I can see what is happening: scrolling a picture can somehow throw off the scrolling of the next slide (when moving forward) or the previous slide (when moving backward).

This is only an issue when `--use-db` is in effect. In that case the picture viewer uses an internal dict and saves that dict to a `.json` file on exit. I'm pretty sure the dict itself is updated correctly. The problem is in moving the viewrect (viewport?) just after creating the image.

@tbpassin I wonder whether you would consider helping with this issue. I've tried everything I can think of with Qt scrolling, but I'm a putz when it comes to graphics.

tbpassin:
I could look at it, though I'm mostly in the dark about scrolling.  I do know that the QtWebEngine doesn't give you an API for scrollbar positioning, as best as I have been able to find out.  That's been  maddening in the VR3 context.  The older WebKit widget did.

Can you point me to some example cases so I don't have to roll my own?

edreamleo:
@tbpassin I'm not sure I know what you mean by example cases. 

`Slides.scrollContentsBy` gets called when the viewRect has _already_ scrolled.  See the docs for [QAbstractScrollArea.scrollContentsBy](https://doc.qt.io/qt-6/qabstractscrollarea.html#scrollContentsBy) which says:

"Calling this function in order to scroll programmatically is an error, use the scroll bars instead (e.g. by calling [QScrollBar::setValue](https://doc.qt.io/qt-6/qabstractslider.html#value-prop)() directly)."

The main thing is that `Slides.do_scroll` isn't working :-)

tbpassin:
After looking at the docstring I saw I just had to make a directory with some images.  I'm running the pictureviewer from a node in my workbook.

As for scrolling, I'm not sure at this point what the problem is.  If I zoom a picture so it's too large for the window, I can pan it using the arrow keys.  But I don't see any scrollbars.  How is panning different from scrolling?

I do have an unrelated problem, though.  Two of my three test pictures display rotated 90 deg.  They don't show rotated in other viewers I tried, including Irfanview, Paint, and the Windows picture viewer.  At a minimum I think we'll want to add a command to rotate the pictures.

tbpassin:
Finally I think I'm getting synced up with you.  When I specify _use_db_ to `run()`, then close and re-open the viewer, the zoom factor was restored but the panned location was not.

[...]

PR #3171, opened 2023-02-23
PR for #3170: scrolling in the picture_viewer plugin

edreamleo:
See #3170.  I finally realized that I need `Slides.do_scroll`.  But it's not working.

### DIFF
--- a/leo/plugins/picture_viewer.py
+++ b/leo/plugins/picture_viewer.py
@@ -217,6 +217,7 @@ if QtWidgets:
         scale: float = 1.0
         starting_directory: str = None
         use_db: bool = False
+        trace: bool = False
         verbose: bool = False
         wrap_flag: bool = True
 
@@ -242,7 +243,8 @@ if QtWidgets:
             # Change the window's title.
             self.setWindowTitle(file_name)
             # Set self.scale, self.dx and self.dy.
-            print('')  # For trace in load_data.
+            if self.trace:
+                print('')  # For trace in load_data.
             if self.reset_zoom:
                 self.load_data()
             else:
@@ -254,7 +256,7 @@ if QtWidgets:
                     QtGui.QPixmapCache.clear()
                 if 0:  # Doesn't work.
                     self.scroll_area.setupViewport(self.view_port)
-                if 1:  # Doesn't work.
+                if 1:  # Doesn't work?
                     # Reset the viewport.
                     self.reset_scroll()
 
@@ -262,7 +264,7 @@ if QtWidgets:
                 pixmap = QtGui.QPixmap(file_name)
                 TransformationMode = QtCore.Qt if isQt5 else QtCore.Qt.TransformationMode
                 transform = TransformationMode.SmoothTransformation  # pylint: disable=no-member
-                if 1:  # Take the smaller immage.
+                if 1:  # Take the smaller image.
                     image1 = pixmap.scaledToHeight(int(self.height() * self.scale), transform)
                     image2 = pixmap.scaledToWidth(int(self.width() * self.scale), transform)
                     image = image1 if image1.height() <= image2.height() else image2
@@ -486,7 +488,7 @@ if QtWidgets:
             else:
                 self.scale = 1.0
                 self.dx = self.dy = 0
-            if 1:  # Don't remove.
+            if self.trace:  # Don't remove.
                 print(
                     f"load_data: {g.caller():<20} {self.slide_number} scale: {self.scale:9.8} x: "
                     f"{self.dx} y: {self.dy}")
@@ -495,7 +497,7 @@ if QtWidgets:
 
             if 0 <= self.slide_number < len(self.files_list):
                 # Don't remove this trace.
-                if 1 and g.caller() != 'scrollContentsBy':
+                if self.trace and g.caller() != 'scrollContentsBy':
 
                     print(
                         f"save_data: {g.caller():<20} {self.slide_number} scale: {self.scale:9.8} "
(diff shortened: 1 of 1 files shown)


tbpassin:
You need to use _-dy_, then it works.

```
    try:
        self.scroll_lockout = True
        hbar.setValue(dx)
        vbar.setValue(-dy)  # <=======
    finally:
        self.scroll_lockout = False
```

But what do you have against having scrollbars?

tbpassin:
I would never have expected this:

`ScrollBarPolicy.ScrollBarAlwaysOff` but you scroll by telling the scrollbar to do it anyway.

edreamleo:
@tbpassin I tried this kind of code several time. Imo it has no chance of working, regardless of the sign of dy :-)

**Scrollbars**

I dislike scrollbars because:
1. They waste real estate (I usually use full-screen mode) and
2. It's better for me to avoid using the mouse. The arrow keys scroll properly.

As for scrollbar policy, it affects only the visibility of the scrollbars, not their functionality. Two asserts in the `do_scroll` method never fail:
```python
hbar, vbar = area.horizontalScrollBar(), area.verticalScrollBar()
...
assert hbar.isEnabled()
assert vbar.isEnabled()
```
**Traces**

The latest code in this PR significantly improves traces. For example, I opened a test folder, zoomed the first slide (slide 0), moved to the next slide and returned to slide 0 and then quit. The resulting traces shows that that the `x/y (dx/dy)` values are correct when returning to slide 0:
```console
Found 10 pictures in C:/Users/Edward Ream/Edward4/test

load_data: show_slide           0 scale:       1.0 x: 860 y: -332
save_data: zoom_in              0 scale:      1.05 x: 860 y: -332

load_data: show_slide           0 scale:      1.05 x: 860 y: -332
save_data: zoom_in              0 scale:    1.1025 x: 860 y: -332

load_data: show_slide           0 scale:    1.1025 x: 860 y: -332
save_data: next_slide           0 scale:    1.1025 x: 860 y: -332

load_data: show_slide           1 scale:  1.157625 x: 1020 y: -366
save_data: prev_slide           1 scale:  1.157625 x: 1020 y: -366

load_data: show_slide           0 scale:    1.1025 x: 860 y: -332
Writing C:\Users\Dev\.leo\picture_viewer.json
picture_viewer: done
```
*Note*: The traces in `load/save_data` show their calling methods,  `show_slide`, `zoom_in` or `zoom_out`, etc.

**Summary**

`show_slide` is calling `do_scroll` with the proper values of `self.dx` and `self.dy`! This is a new development :-)

Imo, the culprit in *drawing* the next slide is likely that "pre-existing" values for `self.scale`, `self.dx`, and `self.dy` are somehow interfering with the *creation* of the next slide.

My next idea: attempt to re-init the scroll area *before* creating the pixmap for the next slide.

edreamleo:
@tbpassin Despite my comments about the scrollbars being enabled, I'm not convinced that calling `QScrollbar.setValue` is the right approach for this application. But first I'll try helping `setValue` by restoring the scroll area to a pristine state before creating the pixmap.

edreamleo:
@tbpassin Looking at the trace above, I wonder whether the problem might involve altering scaling the `dx/dy` values.

Indeed, `show_slide` contains several lines like
```python
image1 = pixmap.scaledToHeight(int(self.height() * self.scale), transform)
```
At present, neither `show_slide` nor the `zoom_in/out` methods scale the `dx/dy` values. And maybe a good thing too, because the `dx/dy` values are *ints* not *floats*. It's a can of worms.

edreamleo:
@tbpassin I'm stuck. After 36b25d0 I'm out of ideas. It's time for a break. This is a minor issue :-)

tbpassin:
Maybe I have not tried enough variations on opening pictures, so I'll give you more details on what I did.  I created a folder with three jpeg images in it.  I opened it and observed that the program would display all three, although two were rotated.  Then I zoomed in on one of them so that it would be too large for the window, and then I panned it with arrow keys.

I closed the program and re-opened it.  I observed that this one picture was still zoomed in but not panned to the position it had when saved.  I did not zoom the other two pictures and so did not have any panning to check, though they did seem to be displayed the same size as before.

After I changed self.dy to -self.dy in the calls to the scrollbar, after restarting the program, the first image appeared zoomed to the same amount, but now it was panned to the position it had had before.

You may have noticed that I called the scrollbars after the pixmap was scaled and displayed, not before.

I wrote that "it works" because the zoom and pan states of the one image seemed to have been restored after restarting.

As an aside, I would have changed the code that decides whether to use image1 or image2, i.e., at

`image1 = pixmap.scaledToHeight(int(self.height() * self.scale), transform)`

I would have chosen based on the unscaled height and width.  Why do the transform on both when only the chosen one needs it?  Unscaled x and y values will both be scaled linearly, so will still have the same ordering before and after transform.

edreamleo:
@tbpassin Great comments:
- You have probably solved the main problem: restoring the panned values.  I'll use your code.
- I have been fixated on what is likely a harder problem: restoring panned values when revisiting a slide during the same slideshow. Happily, this problem is much less important!
- I'll also take your advice and scale only the winning image.

tbpassin:
With the code I posted, the zoom and pan do not change when I change slides and then return to them within the same session.  So I'm not seeing the problem that you see.

Remember, I'm using the code from a few days ago, before your recent changes, and before the advent of `do_xxx()`.  I just changed the code where the stored values are applied.

What's important but not addressed in this thread is preventing an image from being rotated 90 deg when it should not be.  The two of mine that are wrongly rotated are both portrait orientations.   IOW, they have the height larger than the width.  The landscape image isn't rotated being, and that is correct.

PR #3187, opened 2023-03-02
PR: simplify signatures of three related routines

edreamleo:


edreamleo:
@tbpassin My motto is: confusing code should be clarified :-)

tbpassin (changes requested):
I really object to changing the signature of `find_gnx()`.  Several of my custom in-Leo apps depend on it, and we cannot know about other people's code.  Even within Leo's own code, it must be used many places, and all those will all break.  One obvious one will be CTRL-clicking on a gnx.

When Microsoft decided they had to handle Unicode strings everywhere, they did not change their existing string methods.  Instead they added Unicode versions of the old ones, with "W" appended to their names.  

PLEASE follow their example and define a new method that has the signature you want for this PR, e.g., `find_gnx2()`.  New code can start using it instead, and the current method can be marked as depecated.

edreamleo:
@tbpassin rev 990d3e5 restores find_gnx. Does that work for you?

tbpassin:
> @tbpassin rev [990d3e5](https://github.com/leo-editor/leo-editor/commit/990d3e5bdd441bdae728a32531857397ddfbe2db) restores find_gnx. Does that work for you?

Yes, thank you!

leo/commands/gotoCommands.py
```diff
@@ -25,36 +25,38 @@ def __init__(self, c: Cmdr) -> None:
 
     #@+others
     #@+node:ekr.20100216141722.5622: *3* goto.find_file_line
-    def find_file_line(self, n: int, p: Position = None) -> Tuple[Position, int, bool]:
+    def find_file_line(self, n: int, p: Position = None) -> Tuple[Position, int]:
         """
         Place the cursor on the n'th line (one-based) of an external file.
-        Return (p, offset, found) for unit testing.
+
```
boltex, line 31: This doc string should point out the return values for a failure to find the line is a (None, -1) tuple.

leo/commands/gotoCommands.py
```diff
@@ -81,24 +83,25 @@ def find_node_start(self, p: Position, s: str = None) -> Optional[int]:
                 return i + 1
         return None
     #@+node:ekr.20150622140140.1: *3* goto.find_script_line
-    def find_script_line(self, n: int, root: Position) -> Tuple[Position, int, bool]:
+    def find_script_line(self, n: int, root: Position) -> Tuple[Position, int]:
         """
         Go to line n (zero based) of the script with the given root.
```
boltex, line 88: Same here, This doc string should point out the return values for a failure to find the line is a (None, -1) tuple.

leo/core/leoserver.py
```diff
@@ -2110,11 +2110,11 @@ def goto_global_line(self, param: Param) -> Response:
         gc = c.gotoCommands
         line = param.get('line', 1)
         try:
-            junk_p, junk_offset, found = gc.find_file_line(n=int(line))
+            p, junk_offset = gc.find_file_line(n=int(line))
+            focus = self._get_focus()
```
boltex, line 2114: Great fix for the server too, thanks! But _get_focus should remain outside the try-except block as it's method body is a try-except block itself already.

boltex (changes requested):
noted 3 small comments about those changes. nothing big :)

leo/core/leoserver.py
edreamleo: I agree. `_get_focus` gives a more specific error.

Issue #3190, opened 2023-03-03
show-settings Command Fails


tbpassin:
PR #3191 fixes the crash by replacing the `assert` with a comment. In general such `assert`s can do more harm than good.

Some recent change seems to have caused this.  Here is the error traceback:
```
Traceback (most recent call last):

  File "c:\Tom\git\leo-editor\leo\core\leoGlobals.py", line 304, in new_cmd_wrapper
    func(self, event=event)

  File "c:\Tom\git\leo-editor\leo\commands\helpCommands.py", line 1152, in showSettings
    self.c.config.printSettings()

  File "c:\Tom\git\leo-editor\leo\core\leoConfig.py", line 1941, in printSettings
    for name, val, _c, letter in g.app.config.config_iter(c):

  File "c:\Tom\git\leo-editor\leo\core\leoConfig.py", line 1240, in config_iter
    assert isinstance(gs, g.GeneralSetting), repr(gs)

AssertionError: None
```

```
Leo 6.7.2-devel, devel branch, build 5f2578c58b
2023-02-27 13:53:28 -0500
Python 3.10.9, PyQt version 6.4.1
Windows 10 AMD64 (build 10.0.19044) SP0
```

tbpassin:
This isn't as simple as it seems.  I got this failure repeatedly, but not when I opened another Leo session and tried with one or two other outlines.

The code at the line pointed to in the error message is:
```
    for key in sorted(list(d.keys())):
        gs = d.get(key)
        assert isinstance(gs, g.GeneralSetting), repr(gs)
        if gs and gs.kind:
```
I'm thinking that `d.get(key)` probably returned None or an empty string.  This probably doesn't happen very often, and could be handled by adding a line:

```    
for key in sorted(list(d.keys())):
        gs = d.get(key)
        if not gs: continue   # <----------
        assert isinstance(gs, g.GeneralSetting), repr(gs)
```

edreamleo:
@tbpassin Thanks for this report. Testing for `None` should suffice. I'll fix this next.

edreamleo:
@tbpassin PR #3191 fixes the original crash.

I'll leave this issue for your further work. Please base that work on `devel`. Or feel free to close this issue :-)

tbpassin:
I think it's OK to close it at this time.  Thanks!

PR #3191, opened 2023-03-03
PR for #3190: crash in show-settings

edreamleo:
See #3190.

### DIFF
--- a/leo/core/leoConfig.py
+++ b/leo/core/leoConfig.py
@@ -1234,10 +1234,9 @@ class GlobalConfigManager:
         limit = c.config.getInt('print-settings-at-data-limit')
         if limit is None:
             limit = 20  # A reasonable default.
-        # pylint: disable=len-as-condition
         for key in sorted(list(d.keys())):
             gs = d.get(key)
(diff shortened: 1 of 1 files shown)


tbpassin (changes requested):
I don't think this quite gets at my particular issue.  Instead of quitting if a key does not have a settings value, I would like that key to be skipped but the loop should continue.  In my case, somehow several outlines seem to have (non-reproducibly) loaded blank settings.  They do not need to fail an assertion, since they will not cause any harm.  Thus a continue statement is appropriate.  The assertion should still protect against incorrect types as it currently does.

edreamleo:
@tbpassin Please change this PR or issue a new PR as you like.

tbpassin:
> @tbpassin Please change this PR or issue a new PR as you like.

OK. How can I amend your PR?

edreamleo:
I think you can check out my branch and then push your changes.  If not, just create your own branch, delete the offending assert and go from there.

edreamleo:
@tbpassin I have resolved the original issue by replacing the `assert` with a comment.

In general, such `assert`s can do more harm than good.

Issue #3194, opened 2023-03-07
Add c.find_h() and c.find_b()


SegundoBob:
See PR #3196.

**Original Report**

I have many scripts that use `c.find_h()` and `c.find_b()`.  They have simple calls and are adequate to my needs.  Yesterday I updated to the latest devel Leo-Editor and I find that they no longer exist.  Instead there seems to be various much more complicated search functions in QuickSearchController in the leoserver.py module.

How about restoring the simple functions cmdr.find_h() and cmdr.find_b()?  Or telling me how best to replace my calls to cmdr.find_h() and cmdr.find_b()?

I tried posting this question on the Leo-Editor forum, but it was repeatedly deleted immediately.   This question is probably more appropriate for the forum.  I don't consider the problem a bug.

Regards,
SegundoBob

edreamleo:
@SegundoBob I apologize for the missing methods. I don't remember deleting them on purpose, though perhaps I did.

I'll put them back later today.

edreamleo:
@SegundoBob Upon further study, I'd like to amend my suggestion. Instead of using functions it would be good to define `c.find_h` and `c.find_b` exactly as shown above.

At present, the `stickynotes.py` plugin and the `leo_babel/tests/lib_test.py` test file call `c.find_h` and `c.find_b`. Apparently Leo's tests (especially pylint) are not covering these files properly.

SegundoBob:
Edward,

c.find_b() and c.find_h() would be great.

I had to poke around leoPyRef.leo awhile to discover how to call the new find_b() and find_h():

`lx = c.frame.nav.scon.find_h('rfd2302', c.all_positions())`

So c.find_h() would be a great improvement.

The old c.find_h() and c.find_b() only had one required argument.  The new c.find_b() and c.find_h() have a second required argument.  If you gave this second argument the default value `c.all_positions()`, then all the existing calls to c.find_b() and c.find_h() would continue to work as before without having to change the calls.  Unfortunately, this would require some change to the existing function definitions or some kludging since `c.all_positions()` must be called when each call is made, not when the function is defined.

Since the two new arguments (positions and options)  are useful, I think they should be retained.

Regards,
SegundoBob

tbpassin:
> ```python
> def find_b(self,
>     regex: str,
>     positions: List[Position],
> ```

Wouldn't that `List[Position]` better be any kind of a sequence?  Then generators (and tuples) could be used as-is too.

edreamleo:
[...]

PR #3196, opened 2023-03-08
PR for #3194: Add c.find_h/b commands

edreamleo:
See #3194.

### DIFF
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -13,7 +13,7 @@ import tabnanny
 import tempfile
 import time
 import tokenize
-from typing import Any, Dict, Callable, Generator, List, Optional, Set, Tuple, Union
+from typing import Any, Dict, Callable, Generator, Iterable, List, Optional, Set, Tuple, Union
 from typing import TYPE_CHECKING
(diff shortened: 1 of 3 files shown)


tbpassin (changes requested):
leoCommands.py
-------------
line 4515
    `for v in list(d.keys()):`
  simplify to 
    `for v in d:`

Line 4521
   `def checkBatchOperationsList(self, aList: List) -> Tuple[bool, Dict]:`

   Please add a docstring that explains what the contents of `aList` should be.
   From its use, it appears to expect a triple op, p, n.  Please say what
   those should represent.  I can see later in the code that p is expected to
   be a position, but please explain a little more.

Line 4536
```
if not ok:
       break
```

Shouldn't this be continue instead of break?  You want to process
all z's even if some of them aren't ok, don't you?

Line 4525
    ` try:`

What is this try/except block trying to protect?  It seems to be surrounding too much code. If op, p, n = z might fail, then the rest of the code should not be in the try block. If it is not being guarded, it can come before the try block.

edreamleo:
@tbpassin The changes to `checkBatchOperationsList` would have nothing to do with this PR, so I won't make them.

PR #3209, opened 2023-03-19
1. Adjust leo test and utility scripts in the scripts directory to automatically change to the leo-editor directory

tbpassin:
Previously, the directory was hard-coded to c:\Repos\leo-editor, which will be wrong for most devs.

2. Added new cmd file (set-repo-dir.cmd) to perform the cd to leo-editor.

3. Changed "python" to "py" because the two can run different versions of Python 3, and the one launched by "py"  is usually what is wanted.  (E.g., on my system "python" launches 3.9.9, while "py" launches 3.10.9).   (in a venv, "py" launches the version of Python that was used to set up the venv).

4.  Corrected the path in make-leo.cmd.  It could not have worked as it was.  (Perhaps the doc directory used to be in leo-editor, but now it is in leo-editor/leo).

NOTE - these changes do not address the case where the reindent script is not on the system path (as is the case for my system).

NOTE - this commit includes the changes to LeoPyRef because the new @clean set-repo-dir.cmd node had to be added to it.

### DIFF
--- a/leo/core/LeoPyRef.leo
+++ b/leo/core/LeoPyRef.leo
@@ -205,6 +205,7 @@
 <v t="ekr.20170302123956.1"><vh>@file ../doc/leoAttic.txt</vh></v>
 </v>
 <v t="ekr.20221204070905.1"><vh>Script files</vh>
+<v t="tom.20230319122643.1"><vh>@clean ../scripts/set-repo-dir.cmd</vh></v>
 <v t="ekr.20221204072456.1"><vh>@clean ../scripts/beautify-leo.cmd</vh></v>
 <v t="ekr.20230115020533.1"><vh>@clean ../scripts/blacken-leo.cmd</vh></v>
 <v t="ekr.20221204074235.1"><vh>@clean ../scripts/flake8-leo.cmd</vh></v>
@@ -3956,9 +3957,10 @@ def batch_rule8a(colorer, s, i):
 </t>
 <t tx="ekr.20221201043917.1">@language batch
 echo off
-cd C:\Repos\leo-editor
+call set-repo-dir
+
 echo test-leo
-python -m unittest %*
+py -m unittest %*
 </t>
 <t tx="ekr.20221201080228.1">@language python
 #!/usr/bin/env python3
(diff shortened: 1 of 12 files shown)


edreamleo (approved):
Good work, Thomas. It's great to have a second pair of eyes and second test environment.

tbpassin:
: ) You kept at me to use test-all, etc, but it couldn't work as it was.

I still need to solve the reindent path problem.

edreamleo:
@tbpassin Thanks for reminding me about `reindent`&mdash;it's in `python/Tools/scripts/reindent.py`.

tbpassin:
I know where it is, but it's not on my path.  It won't run just by invoking its name.  It's not on Python's path either, so I can't just issue "py -m reindent".  I'm trying to come up with a relatively fool-proof method for the cmd file to find it without assistance.

edreamleo:
@tbpassin The contents of my `reindent.cmd` file is:

    python310 C:\Python\Python3.10\Tools\scripts\reindent.py -n -r %*

A hard-code python should be just fine. `reindent` probably hasn't changed in a decade. Just use any available version.

edreamleo:
Rev a62da5bc in devel restores `reindent-leo.cmd`.

`reindent-leo.cmd` *must* assume that reindent is on the path. Imo, the best (easiest) way to do this to define `reindent.cmd` as shown above.

tbpassin:
But it's **not** on the path or sys.path for my computer.  Why not?  Because which Python version should have its Tools/scripts directory on the path?  On my last computer, I needed to have Python 2.7 as well, and its installer put its directories on the path.  For each version of Python3.x I installed, I actually wrote a batch file for each that munged the Python2.7 path segments to point to a non-existent location, then added the Python 3.x directories.  Ugh!  I don't want to do that again.

I also don't want to hard-code the directories in the leo/scripts directory because they won't always work for anyone else, and if I instead modify the script for my own system each time I pull a new version from Github it will be something that I forget to do (and be a real pain).

Anyway, I know how to find the right directory with a Python script:
```
import sys, os.path
SUFFIX = r'Tools\Scripts\reindent.py'
path = None
for pth in sys.path:
    candidate = os.path.join(pth, SUFFIX)
    if os.path.exists(candidate):
        path = pth
        break
g.es(path)
```

So I plan to incorporate this into a new _leo-reindent.py,_ and call it from _reindent-leo.cmd_.  I just want to verify it works in a venv - pretty sure it will.  Then I'll put in a PR tonight or tomorrow.

And then there's Linux ...

edreamleo:
@tbpassin I tried something like your approach but couldn't make work. I wish you luck :-)

Please file a new PR for any new code. Thanks.

tbpassin:
Solved it! See PR https://github.com/leo-editor/leo-editor/pull/3211.

The solution is to write the full path of _reindent.py_ to a file in ~/.leo.  Then the _reindent-leo_ script can set a variable from the file and use it to launch reindent.py.

Writing the path to an environmental variable didn't work because that variable vanishes when the Python process that wrote it ends.

PR #3210, opened 2023-03-22
Boltex/issue3205

### LINKED ISSUE #3188: Add reverse-sort-children and reverse-sort-siblings commands
I recently wished I had reverse-sort-siblings.

### LINKED ISSUE #3205: Bug in redo sort-children
The redo sort-children does not work when trying to redo commands after undoing itself,  and undoing a prior sort-sibling on another node.

In other words, to reproduce: 
1: sort siblings somewhere 
2: sort Children somewhere else, 
3: undo multiple times, 
4: redo multiple times, and observe that the sort-children does not get redone.

boltex:
(Passes all tests & manually tested) Fixes #3205 and Fixes #3188
Fixes sort commands and their undo operations.
Adds reverse-sort commands.
Fixes the revert file command clearing of the undoer.
Fixed available commands in leo-server.
Changed Félix's old pseudonym to his real name.


boltex:
I reverted to the old sort commands architecture, but kept the clever 'Position fix' that @edreamleo  had put in place. 

I also found and fixed the source of the elusive sort commands bug, (i.e. not copying the saved order array when setting it back.) which was unrelated to the 'position fix'.

I also fixed other lesser bugs: Leaving the focus on the affected node after doing a sort command. That is: the parent node, if sort-children is used, and the original selected node, when doing sort-siblings.

Idem for their undo/redo counterparts.

Félix

PR #3211, opened 2023-03-22
Running Leo's utility scripts from any directory

tbpassin:
1. Add new python script find-reindent.py;
2. Make all scripts find set-repo-dir.cmd from where-ever they are launched;
3. reindent-leo.cmd now reads location of reindent.py from a file .leo\reindent-path.txt (written by #1 ).

Note - all scripts now use %~dp0 for the complete path to the script's directory (leo-editor-scripts).  So no matter where one of these scripts is launched from it will use the right directory paths.

The new script find-reindent.py finds reindent.py and writes its path to a file in ~/.leo.  The reindent-leo script reads this path into an environmental variable which it uses when it launches reindent.py.  If the path cannot be found, reindent.py is not run.

NOTE - reindent.py is now launched with the "-r" parameter so that it will recurse through leo and its subdirectories.

LeoPyRef had to be included in this commit because a new script was added.

### DIFF
--- a/leo/core/LeoPyRef.leo
+++ b/leo/core/LeoPyRef.leo
@@ -206,6 +206,7 @@
 </v>
 <v t="ekr.20221204070905.1"><vh>Script files</vh>
 <v t="tom.20230319122643.1"><vh>@clean ../scripts/set-repo-dir.cmd</vh></v>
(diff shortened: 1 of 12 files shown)


edreamleo (changes requested):
Good job finding `Tools/scripts/reindent.py`.

Please change `reindent.cmd` so that it reindents only the `leo/core` and `leo/commands` folders.

PR #3214, opened 2023-03-25
PR: qt_tree.py: expand/contract when mouseover with CTRL/SHIFT

tbpassin:
Add new event handler onItemEntered(): expand/contract node when mouse moves over it with <CTRL>/<SHIFT> pressed.  Connect to QTree "itemEntered" event.

Note: tree setMouseTracking(True) has been added to enable this event.

**EKR**: Thomas showed me yesterday that this allows a nifty UI improvement.

### DIFF
--- a/leo/plugins/qt_tree.py
+++ b/leo/plugins/qt_tree.py
@@ -103,6 +103,8 @@ class LeoQtTree(leoFrame.LeoTree):
         tw = self.treeWidget
         tw.itemDoubleClicked.connect(self.onItemDoubleClicked)
         tw.itemClicked.connect(self.onItemClicked)
+        tw.setMouseTracking(True)
+        tw.itemEntered.connect(self.onItemEntered)
         tw.itemSelectionChanged.connect(self.onTreeSelect)
         tw.itemCollapsed.connect(self.onItemCollapsed)
         tw.itemExpanded.connect(self.onItemExpanded)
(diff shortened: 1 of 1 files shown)


edreamleo (approved):

boltex:
@edreamleo @tbpassin  IMO This is really surprising and annoying somehow... 
e.g.: when navigating the tree with the mouse and going CTRL+C to copy, or CTRL+B to execute, etc, but because of this new behavior and because the mouse cursor was over the node, it opened the node and shifted the tree and/or selected the first child, so as to do the intended 'ctrl'+XXX command to the child instead... Could this be optional and not by default?

Félix

tbpassin:
OK, I'll take care of it.  Either with a setting or something else I'll come up with.  I'm sorry I overlooked it.  I must always have had the mouse over the body when I was testing.

boltex:
I'm on ubuntu 20.04.

it does it while i'm moving slightly also at the same time. it's inconsistent. (seems after a parent as opened, i have to mouse out of, and back into the headline of a child for it to open with this mouse hover+ctrl feature.) 

(after trying it out some more...)

oh i see, it's while it's moving and the cursor 'gets into the bounding box' while having ctrl+pressed. 

Well indeed it surprised me as to how and why this happened. I have not 100% mouse immobility as I'm starting to press CTRL+something, thus crossing over a node bounding box and activating this feature as i'm intending to use a CTRL+something command.  I would think this should be 'turned on' voluntarily by some config option.

tbpassin:
Yes, it's not really changing on "hover".  Qt will only emit the event when the mouse "enters" a node (there is no "leave" event). Qt does not have a "hover" event per se. I first implemented this feature in browsers (in Javascript), which do have a "hover" CSS pseudo-state - not quite the same as an event.

I'll add an `@bool` setting to enable the feature, and a guard against having CTRL and SHIFT down at the same time.  I'll also check it on my Ubuntu 22 VM.  In the meantime, if you want to disable these gestures, in qt_tree.py comment out the line (in `qtree.initAfterLoad()`)

```
tw.setMouseTracking(True)
```
(Or change its arg to be False)
and restart Leo.

edreamleo:
On Sat, Mar 25, 2023 at 11:07 PM Felix ***@***.***> wrote:

> @edreamleo <https://github.com/edreamleo> @tbpassin
> <https://github.com/tbpassin> IMO This is really surprising and annoying
> somehow...
>
I agree that there are some rough edges that should be cleaned up. I
approved the PR without testing.

There are several possible user options:

- A general on/off switch.
- Which modifiers trigger this behavior.
- Whether mousing *toggles* the expansion state.

For example, shift-control mousing might close expanded nodes and open
unexpanded nodes.

@Thomas Passin ***@***.***> I'll leave the details to you :-)

Edward

tbpassin:
I'll keep on it. Toggling the expansion state does not work well at all.  You are always opening and closing nodes you didn't expect to.

edreamleo:
@tbpassin Thanks for the heads up.

tbpassin:
I just opened PR https://github.com/leo-editor/leo-editor/pull/3218.  This adds a setting to allow or disallow the mouse gesture behavior.  The change also prevents any expansion or contraction if both the CTRL and SHIFT keys are depressed when the mouse moves over a node on the tree.

PR #3218, opened 2023-03-26
qt_tree.py Mouse Gestures

tbpassin:
qt_tree.py:

1. Add boolean setting _use-mouse-expand-gestures_ to control tree expansion by mouse gesture;
2. Do not expand or contract tree on mouse enter if both CTRL and SHIFT keys are depressed.
3. Remove temporary diagnostic line.


edreamleo (approved):
This new code should be safe because it is disabled by default.

PR #3220, opened 2023-03-26
Boltex/issue3216

### LINKED ISSUE #3216: Copy / paste as JSON is broken when nodes contain 'user attributes' (ua's)
Copy / paste as JSON is broken when nodes contain ua's. **But since the Save/Open with JSON (.leojs files) works fine,** and it's supposed to use the same (almost identical) methods & algorithms, then surely this bug is due to a small overlooked detail.

I'll try to find it and fix it.

boltex:
Fixes #3216, Makes the .leojs  file output stable, along with the copy-node-as-json command too.
Also makes the .leojs file format associated with Leo by the installation script utilities.
Also fixed EKRLight UI theme.


boltex:
I've made the unittest run 100% under windows too.

edreamleo (approved):

Issue #3223, opened 2023-03-27
Improve mod_autosave plugin


edreamleo:
`mod_scripting.onIdle` should not cause cause any UI changes.  See PR #3224.

edreamleo:
Closed via PR #3224.

PR #3224, opened 2023-03-27
PR for #3223: improve mod_autosave plugin

edreamleo:


tbpassin (commented):
1. It looks like all at-file save messages will be printed to the console.  I would be very annoyed if I periodically got a screen of messages every minute (or whatever auto the save interval is).

2. I think there could be a potential timing conflict if the autosave process starts while the user is doing something active (possibly navigating or editing) as the autosave starts.  I realize that it's called as an onIdle process, but I don't really know when onIdle handlers get called.  I'm also concerned if the idle period can end and the user start to do something while the autosave is in progress - as it might be if the outline save takes a relatively long time (e.g., LeoPyRef).  These may be non-issues as I don't understand onIdle.

3. Lines 79-80:

    d['last'] = time.time()
    gDict[c.hash()] = d

This will delete any other keys that may happen to be in gDict[c.hash()].  Potentially this dict might get used by someone else, and we don't want to delete their items.  I recommend instead:

    d_ = gDict[c.hash()]
    d_['last'] = time.time()

or even:

    gDict[c.hash()]['last'] = time.time()

4. I'm concerned about Leo freezing during a longer autosave, which will be really annoying for a user.

edreamleo:
@tbpassin Thanks for your comments:

- Idle time happens when no UI events are happening. In practice, this means that all of Leo's main thread has completed. I see no potential for problems. Any race conditions would have become apparent long ago.
- I think the console messages are important, but I suppose there could be a `silent` option.
  I plan to start using this plugin myself. Let's keep this item open.
- If people can't abide long saves then I don't see how they could possibly use this plugin.
  No way am I going to do work in a separate process.
- I like the `Dict[c.hash()]['last'] = time.time()` pattern. I'll make that change.

edreamleo:
@tbpassin After the user saves the outline the plugin won't generate a message again until the user changes the outline. Imo, this message is important.

There is an unavoidable tradeoff here: If the plugin doesn't change the UI there is *no* (non-heroic) way to suppress "redundant" backups. Clearing the `c.changed` ivar would be dangerously wrong. It would prevent Leo from prompting for a save when the user closes the outline.

tbpassin:
> @tbpassin After the user saves the outline the plugin won't generate a message again until the user changes the outline. Imo, this message is important.
> 
> There is an unavoidable tradeoff here: If the plugin doesn't change the UI there is _no_ (non-heroic) way to suppress "redundant" backups. Clearing the `c.changed` ivar would be dangerously wrong. It would prevent Leo from prompting for a save when the user closes the outline.

Other programs I've worked with that do autosave do it silently.  One doesn't even know it's happening. Maybe it's different for Leo, because an outline can be very large and have many external files.

edreamleo:
@tbpassin Hehe. I was about to argue about preferences. I'll add `@bool mod-autosave-verbose` with default `False`.

edreamleo:
@tbpassin I've done everything I can think of. Please approve :-)

tbpassin (approved):
Looks good to me.

edreamleo:
@tbpassin Thanks for your approval. We can always file a new PR if we find other problems.

Issue #3121, opened 2023-02-01
Add 3 commands: insert and delete serial number


wangzhaohe:
There are 3 commands added when I write asciidoc article.

@edreamleo Please have a check wheather they are useful.

The test leo image:

![WX20230201-180117@2x](https://user-images.githubusercontent.com/2628013/216016372-a4cb9156-90d0-4522-91a5-350027670041.png)

The test leo file:

[insert_and_delete_sn.leo.zip](https://github.com/leo-editor/leo-editor/files/10555892/insert_and_delete_sn.leo.zip)

edreamleo:
Yes, these commands (`@button` nodes) are useful. They could remain code snippets or they could become commands.

What would you like me to do?

tbpassin:
I think these commands would be useful.  I see some places where the code could be improved, if you would not mind a small code review.  Among other things the command and function names could be made more evocative of what they do.

edreamleo:
@tbpassin Thanks for your help. The usual Leonine way to add commands is via a plugin. VR3 is a good example :-)

wangzhaohe:
@tbpassin would you mind helping to improve the code and give them proper names? Thanks a lot.

tbpassin:
Please understand that these code review suggestions are meant to share the accumulated experience of many people.  They are not meant as criticism.

1. Names of commands and functions should suggest their purpose more clearly.
    1a. `delete-all-subheads-sn` Perhaps `delete-subtree-numbering`.
    1b. `insert-all-subheads-sn` Perhaps `add-subtree-numbering`.
    1c. `insert-subheads-sn` Perhaps `add-child-numbering`.
    1d. `get_p_children()` - This function does not "get" any children; it changes their headlines.  Perhaps `renumber_child_headlines()`.

2. In `delete-all-subheads-sn` 
    2a. `for n,p in enumerate(p.self_and_subtree())` Use a different variable name; e.g., `for n, p1 in enumerate(p.self_and_subtree())`.  Otherwise the value of _p_ will be different at the end.  This may not be what you expect.
    2b. `if n == 0: continue` can be avoided by using `p.subtree()` instead of `p.self_and_subtree()`.  `n` would need to be changed to `n + 1` where it is used.

[...]

PR #3227, opened 2023-03-28
PR for #3121: headline numbers

edreamleo:
See #3121.

PR #3228 (merged into devel) adds related undo/redo helpers.

- [x] Add `hn-add-all`, `hn-add-subtree`, `hn-delete-all`, and `hn-delete-subtree` commands,
    including various aliases.
    - Redraw the resulting trees and mark nodes dirty.
    - Add docstrings.
- [x] Make all commands undoable.

### DIFF
--- a/leo/commands/editCommands.py
+++ b/leo/commands/editCommands.py
@@ -1103,6 +1103,7 @@ class EditCommandsClass(BaseEditCommandsClass):
         c = self.c
         for p in c.all_unique_positions():
             self.hn_add(p=p)
+        c.redraw()
     #@+node:ekr.20230328013557.1: *4* hn-add-tree
     @cmd('hn-add-tree')
     @cmd('headline-number-add-tree')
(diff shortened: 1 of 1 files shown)


tbpassin (changes requested):
1. Line 1089: def hn_add()
    Supply a docstring explaining what this method is intended to do.
    Given a good docstring here, the other related commands probably
    do not need docstrings.  Without a docstring, the purpose of this
    command is obscure.

2. Line 1119: hn_pattern = re.compile(r'^[0-9]+(\.[0-9]+)* ')
    Move this line to the top of the node.

edreamleo:
@tbpassin I'll add docstrings.

edreamleo:
@tbpassin I've made all your requested changes. Please review again :-)

edreamleo:
@tbpassin
- Named sections are always within an `@<file>` tree, so there is no longer any need to mention such nodes separately.
- The function appears inside the node because it is used nowhere else.

edreamleo:
@tbpassin `@<file>` means any kind of external file.

PR #3228, opened 2023-03-29
Boltex/issue3219

boltex:
Seems to fix headline undos along with providing methods for multiple-headline-changes operations.


edreamleo (approved):
Thanks for this work. I'll merge it into `devel` and  `ekr-3121-headline-numbers`. If any problems arise I'll fix them in the latter branch first.

PR #3229, opened 2023-03-30
Boltex/issue3219

boltex:
Switched to usage of c.setHeadString in commands that only change headlines and removed thus extraneous code in the undoer.


edreamleo (approved):
I'm glad the special-case undo code is gone.

Issue #3231, opened 2023-03-31
Leo inserts ~/.leo into sys.path but with wrong separators


tbpassin:
See PR #3233.

On Windows 10, my `sys.path` starts with an entry with the wrong path separators for Windows: `C:/Users/tom/.leo`.  

```
Leo 6.7.3-devel, devel branch, build fc31c98df4
2023-03-29 02:23:54 -0500
Python 3.10.9, PyQt version 6.4.1
Windows 10 AMD64 (build 10.0.19044) SP0
```
This could cause modules (like plugins) specified relative to `%USERPROFILE%/.leo` not to be found.

edreamleo:
@tbpassin Thanks for this report. I'll fix it next.

edreamleo:
Closed via PR #3233.

PR #3233, opened 2023-04-01
PR for #3231: append to sys.path

edreamleo:
See #3231.  Do *not* change `g.os_path_finalize`! Doing so will ruin all existing `.leo` files!

- [x] Use `os.path` functions (*not* `g.os_path_*`) functions whenever updating `sys.path`.
  The new pattern is:  `path = os.path.normpath.os.path.join(g.app.loadDir, ...))`
  This pattern is more explicit and avoids all complications.

The `g.os_path_*` wrappers functions are usually faux helpers. Alas, removing them is *not* generally possible.

To put the `leo-editor` directory on `sys.path`:

```python
path = os.path.normpath.os.path.join(g.app.loadDir, '..', '..'))
if path not in sys.path:
    sys.path.append(path)
```

To put the `leo-editor/leo` directory on `sys.path`:

```python
path = os.path.normpath.os.path.join(g.app.loadDir, '..'))
if path not in sys.path:
    sys.path.append(path)
```

### DIFF
--- a/leo/commands/checkerCommands.py
+++ b/leo/commands/checkerCommands.py
@@ -447,8 +447,8 @@ class MypyCommand:
             print('install mypy with `pip install mypy`')
             return
         root = p.copy()
-        # Make sure the leo directory is on sys.path.
-        path = os.path.normpath(os.path.join(g.app.loadDir, '..'))
+        # Make sure the leo-editor directory is on sys.path.
+        path = os.path.normpath(os.path.join(g.app.loadDir, '..', '..'))
         if path not in sys.path:
             sys.path.append(path)
         roots = g.findRootsWithPredicate(c, root, predicate=None)
@@ -484,8 +484,8 @@ class Flake8Command:
         c, root = self.c, p
         if not flake8:
             return
-        # Make sure the leo directory is on sys.path.
-        path = os.path.normpath(os.path.join(g.app.loadDir, '..'))
+        # Make sure the leo-editor directory is on sys.path.
+        path = os.path.normpath(os.path.join(g.app.loadDir, '..', '..'))
         if path not in sys.path:
             sys.path.append(path)
         roots = g.findRootsWithPredicate(c, root, predicate=None)
@@ -574,8 +574,8 @@ class PyflakesCommand:
         c, root = self.c, p
         if not pyflakes:
             return True
-        # Make sure the leo directory is on sys.path.
-        path = os.path.normpath(os.path.join(g.app.loadDir, '..'))
+        # Make sure the leo-editor directory is on sys.path.
+        path = os.path.normpath(os.path.join(g.app.loadDir, '..', '..'))
         if path not in sys.path:
             sys.path.append(path)
         roots = g.findRootsWithPredicate(c, root, predicate=None)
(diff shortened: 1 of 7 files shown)


tbpassin (changes requested):
leoflexx.py lines 46ff:
    
    # This is what Leo typically does.
    # pylint: disable=wrong-import-position
    path = os.getcwd()
    if path not in sys.path:
        sys.path.append(path)
    g.appendToSysPath(os.getcwd())

    # JS code can *not* use g.trace, g.callers or g.pdb.
    from leo.core import leoGlobals as g
    
Uses g before importing it.

tbpassin:
This why I didn't want to dive in and "fix" this issue myself.  I wouldn't have known where it should and shouldn't be used.

edreamleo:
@tbpassin Thanks for your review. Rev b7b0f14 fixes the problem you found.

Your caution concerning this issue was well founded. My first thought was to remove the weird special case for Windows in `g.os_path_finalize`. But after removing that code Leo couldn't find any of the `@file` nodes in my outline! I'm not sure why that happens but that doesn't matter&mdash;removing the code would be exceptionally dangerous.

So I had no choice but to find all the places where Leo updates `sys.path`.

edreamleo:
@tbpassin Please review the latest code.

The code no longer uses `g.appendToSysPath function`. It was a faux helper.

tbpassin (changes requested):
I missed seeing problems about this path because my launching script puts _git/leo-editor_ first on the path.

Don't we want to put the **parent** of the leo directory on the path rather than the leo directory itself?  My programs, and Leo's code base, are littered with imports like `from leo.core import leoNodes`.  That is:
```
import os.path
dir = os.path.normpath(os.path.join(g.app.loadDir, '..'))
g.es(dir)  # C:\Tom\git\leo-editor\leo
```
This isn't the parent of the leo directory.

What is really needed is for `g.app.leoDir\..` to be on the path, isn't it?

```
import os.path
dir = os.path.normpath(os.path.join(g.app.leoDir, '..'))
g.es(dir)  # C:\Tom\git\leo-editor

```

edreamleo:
@tbpassin Thanks for your comment.  I think you are correct, but I think you meant to say:

What is really needed is for g.app.leoDir\.. to be on the path, isn't it?

```python
dir = os.path.normpath(os.path.join(g.app.leoDir, '..', '..'))   # Go back an extra level.
g.es(dir)  # C:\Tom\git\leo-editor
```
Right?

tbpassin:
No, it's correct the way I wrote it.  You might be mixing up `g.app.leoDir` and `g.app.loadDir`:

```
g.es('g.app.leoDir:', g.app.leoDir)
g.es('g.app.loadDir:', g.app.loadDir)
# Displays:
g.app.leoDir: C:/Tom/git/leo-editor/leo
g.app.loadDir: C:/Tom/git/leo-editor/leo/core
```

edreamleo:
@tbpassin I see. Please review the latest code. I think it's equivalent to your suggestions.

edreamleo:
@tbpassin Also, please have a look at #3235. I'd appreciate your comments.

tbpassin (changes requested):
`sys.path.append(path)` - I still think this should be `sys.path.insert(0, path)`.  Otherwise some other Leo install might be found first.

edreamleo:
@tbpassin I agree. The pattern should use `sys.path.insert(0, path)`.

edreamleo:
@tbpassin I think I've complied with all your requests. Please approve this PR if so.

tbpassin (changes requested):
1. _runLeo.py_, line 37:

```
if path not in sys.path:
    # print('appending %s to sys.path' % path)
    sys.path.append(path)
```
Shouldn't this use `insert(0, path)` instead?

2. _checkerCommands_, line 608, comment:
`# Make sure the leo-editor directory is on sys.path.`
A nit, but a PyPi install won't have _leo-editor_ - how about "the parent directory of leo" instead?

Almost there!

Issue #3235, opened 2023-04-02
Simplify calls to g.os_path wrappers


edreamleo:
**Background**

The `g.os_path_*` wrappers are a legacy of python 2. However, they can *not* be removed from Leo's core because they work differently from the corresponding `os.path` functions:
-  `g.os_path_normslashes` converts backslashes to forward slashes on Windows.
- `g.os_path_join` supports `!!` and `.` args.

**Phase 1: PR #3241 (merged into devel)**

- [x] Use `c.fullPath(x)` instead of `g.fullPath(c, x)`.
- [x] Always call `g.os_path_normslashes` to do convert backslashes to forward slashes.
   This change simplifies future changes.

**Phase 2: (Won't do)**

Making `g.os_path_normslashes` a do-nothing would be an extremely bad idea:
- Making this change would have severe, unpredictable consequences.
- We can rationalize what `g.os_path_normslashes` does by saying that Leo uses POSIX slashes internally.

tbpassin:
I don't favor making serious changes like this for no immediate benefit - as you say, they will be risky.

Instead, I think that new (or code that is being revised anyway) Leo code should start using _pathlib_ methods.  They automatically handle the path separator issue, for one thing.  For another, in some cases their expressions can be more intuitive or natural - something like 

```
from pathlib import PurePath
dir = PurePath(g.app.leoDir) / '..'  # __div__ operator is overloaded - brilliant!
g.es(f'{dir}')  # C:\Tom\git\leo-editor\leo\..

dir_ = PurePath('A')/'B'/'C'
g.es(f'{dir_}')  # A\B\C
```

These paths can be used as is in `file.open()` and similar - they don't need to be converted to strings first.

edreamleo:
@tbpassin Thanks for your comments. I'll try a few experiments and report back.

edreamleo:
@tbpassin I wouldn't say that this issue is without benefit. The hack in `g.os_path_finalize` is a serious blemish. This hack is likely needed only in two or three places.

tbpassin:
`g.os_path_join` enhencements - who even knows about them or uses them?  Not me.

I think that `g.os_path_finalize()` should stay as is - it's working just about where ever it's used, and who knows how many users have code that uses it?

If it is not adequate for some task, then write a new method that is correct and use that.  Or use a few lines of standard methods (e.g., _PathLib_ methods) In the places that need them.

[...]

PR #3241, opened 2023-04-03
PR for #3235: path wrappers

edreamleo:
This PR is phase 1 of #3235. It makes *minimal* changes to Leo's core:

- [x] Create `c.fullPath`.  `g.fullPath` calls `c.fullPath`.
- [x] Change `g.fullPath(c, x)` to `c.fullPath(x)` everywhere.
- [x] `g.os_path_finalize` and `g.os_path_normpath` call `g.os_path_normslashes`.

### DIFF
--- a/leo/commands/checkerCommands.py
+++ b/leo/commands/checkerCommands.py
@@ -421,7 +421,7 @@ class MypyCommand:
             return
         self.unknown_path_names = []
         for root in roots:
-            fn = os.path.normpath(c.fullPath(root))  ###
+            fn = os.path.normpath(c.fullPath(root))
             self.check_file(fn, root)
     #@+node:ekr.20210727212625.1: *3* mypy.check_file
     def check_file(self, fn: str, root: Position) -> None:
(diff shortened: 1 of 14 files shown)


tbpassin (changes requested):
1. Line 2260 c.fullPath()

Path separators end up being forward slashes.  Is this what is intended?  If yes, please say so in the docstring.  If no, please correct the code.

2. os_path_finalize_join() - Say in docstring that the path separators will always be forward slashes.

NOTE - *All* such methods should say how they handle path separators.  It's too hard to remember which ones do what, and it can be too obscure to easilty figure it out.  I'm not going on a search for all methods in this review, but as they come up all the docstrings should be clarified this way.

3. g.fullPath = same comment.

I suggest that in the future, the names of any new path-handling methods should end with "_posix" if they output paths with forward slashes.

edreamleo:

The confusion about path separators is why I want to make `g.os_path_normslashes` a do-nothing function. This confusion pervades all the relevant `g.os_path` functions&mdash;documentation can't really help.

Converting (on Windows) backslashes to forward slashes must be wrong:

- It's a bizarre hack with no obvious purpose.
- Using forward slashes could easily interfere with calls to `os.path.exists`.
- Leo should (without hacks!) be able to handle `@file` nodes containing either forward or back slashes.

Nevertheless, making `g.os_path_normslashes` will take some work:

- In several places Leo calls `os.path.normpath` to undo the (indirect) effect of  `g.os_path_normslashes`.
- At present, making `g.os_path_normslashes` a do-nothing causes Leo not to find (or draw) external files.

My *guess*: there are just one or two places that need to change. Once I find and change those places everything will must work without mucking about with slashes.  We shall see.

edreamleo:
@tbpassin I am going to merge this PR. A second PR will contain all the dangerous stuff.

PR #3254, opened 2023-04-06
PR for #3262: Change @solarized-purple to @solarized-violet in all theme files

edreamleo:
See #3262.

Some theme files used `purple` or colors specified by value. Those files did not change.


tbpassin (commented):
1. `url_color` has been changed to "purple" in some themes (e.g., EKRDark) but to `@solarized-violet` for others (e.g.LeoBlackSolarized).  EKRDark does define other "solarized" colors so this seems inconsistent.  CSS "purple" is a dimmer color than `@solarized-violet`, and for my eyes (with less-than-stellar color vision) seems a little too pale.

2. The `solarized-violet` color used in these Leo themes had the css value '#6c71c4', which agrees with the recommendation on the Solarized web page.  So that seems right.

Issue #3238, opened 2023-04-03
Add command to make a .zip file from an outline and its external files


edreamleo:
See PR #3255.

Write to the `os.environ['LEO_ARCHIVE']` directory or the directory containing the `.leo` file.

Later or never: add additional resources to the `.zip` file.

tbpassin:
Excellent.  A user can always add more resources to the zip file by hand.

edreamleo:
Completed via PR #3255.

PR #3255, opened 2023-04-06
PR for #3238: Add write-zip-archive command

edreamleo:
See #3238.

### DIFF
--- a/leo/core/leoFileCommands.py
+++ b/leo/core/leoFileCommands.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import binascii
 from collections import defaultdict
 from contextlib import contextmanager
+from datetime import datetime
 import difflib
 import hashlib
 import io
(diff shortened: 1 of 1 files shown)


tbpassin (changes requested):
1. Lines 664ff:
```
       if not leo_file:
            print('Please name this .leo file.')
            return
```
Change to read 'Please save this .leo file first.'  I don't know of a way to "name" an outline except to save it, so the original wording is confusing.

2. Lines 687ff:
```
        with zipfile.ZipFile(archive_name, 'w') as f:
            f.write(leo_file)
                if p.isAnyAtFileNode():
                    fn = c.fullPath(p)
                    if os.path.exists(fn):
                        n += 1
                        f.write(fn)
        print(f"Wrote {archive_name} containing {n} file{g.plural(n)}")
```
I wonder if these should be put in a try/except block.

edreamleo:
@tbpassin Thanks for your review. I'll follow both your suggestions.

tbpassin (changes requested):
Lines 675ff:
```
        try:
            directory = os.environ['LEO_ARCHIVE']
            if os.path.exists(directory):
                g.es_print(f"Not found: {directory!r}")
                archive_name = rf"{directory}{os.sep}{g.shortFileName(leo_file)}-{time_s}.zip"
```

It looks like this should read:

`if not os.path.exists(directory):`

edreamleo:
@tbpassin Thanks for your suggestions. Kinda like slow-motion pair programming ;-)

tbpassin (approved):
Looks good from here.

tbpassin:
> @tbpassin Thanks for your suggestions. Kinda like slow-motion pair programming ;-)

It is, isn't is?  Slow or fast, code review is usually helpful.  It's surprising what a person can overlook on their own.  Just like one can't copy-edit his own writing well.

Issue #3232, opened 2023-03-31
runLeo Points To leoeditor.com


tbpassin:
runLeo still has the old address to Leo docs:

` msg += "\nhttps://leoeditor.com/installing.html"`

Yes, very minor, but we might as well take care of it.

PR #3257, opened 2023-04-07
PR for #3232: change Leo's URL in Leo's sources, including xml elements

edreamleo:
See #3232.

- [x] Change `http://leoeditor.com` to `https://leo-editor.github.io/leo-editor` everywhere,
    including in the message in `runLeo.py` written when Leo can't be started.
- [x] Change the `xmlns` element of `.leo` files
    from `http://edreamleo.org/namespaces/leo-python-editor/1.1`
    to `https://leo-editor.github.io/leo-editor/namespaces/leo-python-editor/1.1`
- [x] Change the `Created by Leo` comment in `.leo` files
        from `http://leoeditor.com/leo_toc.html`
        to `https://leo-editor.github.io/leo-editor/leo_toc.html`

### DIFF
--- a/leo/commands/commanderHelpCommands.py
+++ b/leo/commands/commanderHelpCommands.py
@@ -30,7 +30,7 @@ def about(self: Self, event: Event = None) -> None:
         "Copyright 1999-%s by Edward K. Ream\n" +
         "All Rights Reserved\n" +
         "Leo is distributed under the MIT License") % datetime.date.today().year
-    url = "http://leoeditor.com/"
+    url = "https://leo-editor.github.io/leo-editor/"
     email = "edreamleo@gmail.com"
     g.app.gui.runAboutLeoDialog(c, version, theCopyright, url, email)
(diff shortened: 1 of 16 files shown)


tbpassin (approved):
1. In _helpCommands.py_, `tutorial-programming.html` was changed to `tutorial-_scripting.html`.  The new URL points to the right page, but I wonder if there are links within the docs that have not been changed.

Good thing that the namespace is only being used as a formality!  If it were actually significant, we'd have a nightmare.

Actually, there is no requirement that an XML namespace point to anything real.  It only has to be a syntactically valid URI.  Many people argue that a namespace, when used as an part of an identifier, should in fact be possible to dereference so that one can retrieve a page that describes the thing being dereferenced.  This change of server for Leo's namespace highlights the opposing view.

edreamleo:

Keeping URLs up-to-date is always a chore. `@button find-bad-links` in `LeoDocs.leo` helps.

PR #3263, opened 2023-04-08
Added '*.leojs' as filetype extensions for particular dialogs.

### LINKED ISSUE #3261: Some dialogs are missing the .leojs file format choice.
Some save / load / choose file-name dialogs are missing the choice of the new .leojs (JSON) file format. (I've missed one or two when I added this file format to the obvious command's dialogs a few months ago) 

I'll add them (at the appropriate dialogs only). and submit a pull request shortly.

boltex:
Fixes #3261


edreamleo (changes requested):
Let's also support `.db` files. 

See the node `fc.getLeoFile (read switch)`

boltex:
@edward Indeed the .db format is a valid Leo file format. But the "*.db" extension string is already offered as a choice alongside *.leo and *.leojs in the relevant commands. ... Also, I looked at the node that you suggested, but I fail to see a missed place to offer .db files.

boltex:
@edreamleo Oh - I see 2 or 3 that could also have *.db added as a choice that I had overlooked, my bad, - i'll add that choice to those commands too ! :)

boltex:
@edreamleo Allright, you were right, there were many instances where the db extension should have been offered... :) 

I've added them and cleaned up a handful of docstring for commands that were just wrappers of 'saveTo'. (Just to underline that it makes a copy, without changing the actual title or save-state/untitled-state either.)

boltex:
@edreamleo I've made sure to run unit tests under windows too, just in case.

Note: I've left only 3 commands **without** the .db extension choice (readFileIntoNode, writeFileFromNode, and writeFileFromSubtree) because those write and read to/from body as non-binary string only. (therefore .leo's XML and .lsojs's JSON were acceptable in those few instances only.)

edreamleo (approved):
Many thanks for your careful, thorough, work.

Issue #3260, opened 2023-04-07
Allow only safe syntax in c.expand_path_expression


edreamleo:
See PR #3260.

Path expressions are desirable, but they must be made completely safe. Leo' present path expressions (`c.expand_path_expression`) are serious security violations that virus scanners will never find.

Aha: `c.expand_path_expression` need only use Python's existing `os.path` conventions.
Bye-bye `{{` and `}}`!

boltex:
it's already implemented in leojs. 

can you elaborate as to why it would be a security concern?

edreamleo:
Closed via PR #3264.

PR #3264, opened 2023-04-08
PR for #3260: safe path expressions

edreamleo:
See #3260.

- [x] Rewrite `c.expand_path_expression` so that it uses only standard `os.path` functions.
  However, it always returns forward slashes, regardless of platform.
- [x] Rewrite unit test.
- [x] Remove documentation of Leo's legacy path expressions.


edreamleo:
@boltex @tbpassin

At present the unit tests assumes that `os.environ['LEO_BASE_DIRECTORY']` throws `KeyError`.

Please review just `leoCommands.py` for now. I'll make the unit test more robust later.

edreamleo:
@boltex @tbpassin I've just improved the unit test. The PR is complete afaik.

tbpassin (changes requested):
I made some suggestions over on the Group discussion.  Let's see where those go.

edreamleo:
@tbpassin You've said the magic word: there is no need for `~` in path expressions.

Aha: there is no need for *any* path expressions! Bye-bye `{{` and `}}`!

`c.expand_path_expression` will collapse. At most it can look for a few special characters at the start of the path. But first, I'll experiment with what `os.path.normpath` *already* does.

edreamleo:
@boltex @tbpassin The code is ready for review.

It's much simpler than before, but there are subtleties that might not be correct.

Please don't ask for extra features. Imo it's urgent that we approve this PR asap so that it can be tested by all Leonistas.

tbpassin (changes requested):
I'm concerned that line 3027 won't give the intended result if the path to the file is an absolute path:

`path = os.path.normpath(os.path.join(base_dir, expansion))`

When the path is absolute, we don't want to prepend the base_dir.  So the expanded path needs to be checked to see if it is absolute path.  I believe that os.path has a function that does that.

This can happen any time a file is imported into Leo for the first time (and I have some number of external files that use an absolute path).

edreamleo:
@tbpassin Thanks for your careful attention to detail. I considered the question you raised, but not as thoroughly as you did. Here is a test that should put your concerns to rest:
```python
import os
isabs, join, norm = os.path.isabs, os.path.join, os.path.normpath
paths = ('/a', '/b', 'c', 'd')
for path1 in paths:
    for path2 in paths:
        s1 = f"isabs({path1}): {int(isabs(path1))}"
        s2 = f"isabs({path2}): {int(isabs(path2))}"
        s3 = f"norm(join({path1}, {path2})): "
        print(f"{s1:>14} {s2:>14} {s3:>24} {norm(join(path1, path2))}")
```
And here is the output:
```console
  isabs(/a): 1   isabs(/a): 1     norm(join(/a, /a)):  \a
  isabs(/a): 1   isabs(/b): 1     norm(join(/a, /b)):  \b
  isabs(/a): 1    isabs(c): 0      norm(join(/a, c)):  \a\c
  isabs(/a): 1    isabs(d): 0      norm(join(/a, d)):  \a\d
  isabs(/b): 1   isabs(/a): 1     norm(join(/b, /a)):  \a
  isabs(/b): 1   isabs(/b): 1     norm(join(/b, /b)):  \b
  isabs(/b): 1    isabs(c): 0      norm(join(/b, c)):  \b\c
  isabs(/b): 1    isabs(d): 0      norm(join(/b, d)):  \b\d
   isabs(c): 0   isabs(/a): 1      norm(join(c, /a)):  \a
   isabs(c): 0   isabs(/b): 1      norm(join(c, /b)):  \b
   isabs(c): 0    isabs(c): 0       norm(join(c, c)):  c\c
   isabs(c): 0    isabs(d): 0       norm(join(c, d)):  c\d
   isabs(d): 0   isabs(/a): 1      norm(join(d, /a)):  \a
   isabs(d): 0   isabs(/b): 1      norm(join(d, /b)):  \b
   isabs(d): 0    isabs(c): 0       norm(join(d, c)):  d\c
   isabs(d): 0    isabs(d): 0       norm(join(d, d)):  d\d
```

tbpassin:
Wow, os.path.normpath() really does a lot!

This test should go into the unit test if it's not there already.

tbpassin (commented):
Remember, we want Leo to export the LEODIR variable sometime before the first outline gets opened.  This behavior will also need to be mentioned in the docs.

edreamleo:
@tbpassin The unit tests cover all the important cases, I think.

I don't understand your comment about `LEODIR`. Is this a separate issue?

tbpassin:
It needs to be coordinated.  We ended up on the Google Groups thread saying that Leo should export a `LEODIR` variable so that it could be used in paths - e.g., as `$LEODIR` instead of the old syntax which might have been `{{g.app.leoDir}}`.

It needs to be set before any outlines get opened so that it's available for loading external files when the outline loads.

I'm not sure where that should happen but perhaps in `LM.computeStandardDirectories()`.

edreamleo:
@tbpassin Thanks for the clarification. I understand now what you mean.

However, imo Leo need not (should not?) set any environment vars. If users wants to specify a directory, they can assign whatever var they like and use it with the $ syntax.

PR #3273, opened 2023-04-09
PR: Cleanups of g.os_path wrappers and related methods.

edreamleo:
- [x] Remove non-standard path conventions from `c.scanAtPathDirectives`.
   It used a recent-removed setting.
- [x] Remove workarounds for an ancient Python bug.
- [x] Simplify wrappers in other ways.
- [x] Use `g.os_path_normslashes` in several places.

### DIFF
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -2527,18 +2527,7 @@ class Commands:
         """
         c = self
         c.scanAtPathDirectivesCount += 1  # An important statistic.
-        # Step 1: Compute the starting path.
-        # The correct fallback directory is the absolute path to the base.
-        if c.openDirectory:  # Bug fix: 2008/9/18
-            base = c.openDirectory
-        else:
-            base = c.config.getString('relative-path-base-directory')
-            if base and base == "!":
-                base = g.app.loadDir
-            elif base and base == ".":
-                base = c.openDirectory
-            else:
-                base = None  # Settings error.
+        base = c.openDirectory
         base = c.expand_path_expression(base)  # #1341.
         base = g.os_path_expanduser(base)  # #1889.
         absbase = g.os_path_finalize_join(g.app.loadDir, base)  # #1341.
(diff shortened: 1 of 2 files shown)


edreamleo:
@boltex @tbpassin Another round of cleanups. This round is much easier, but I won't merge without your comments.

tbpassin (commented):
This is going to take me some time.  But I don't see where the path in an `@path` node is included.  It's essential to include that path somewhere.

edreamleo:
@tbpassin Thanks for checking. I think the problem with `@path` happened in an earlier PR.

boltex (approved):
Those changes look good. And I'll make sure to mirror them in leojs as soon as they are officially merged, **But i dont know enough about os.path module in python to know if those are really appropriate changes.** I'll leave it to you and Thomas to give the final word on this.

tbpassin:
> Those changes look good. And I'll make sure to mirror them in leojs as soon as they are officially merged, **But i dont know enough about os.path module in python to know if those are really appropriate changes.**

The os.path functions can be complicated in what they do.  How will leojs be able to access the file system at all?  Normally one can't do that from a browser.

tbpassin:
Note that we have to make sure the new path code is going to work with `@rst `nodes.  I have several `@rst` trees that start with a `@path` node, like this:
```
- @path c:\Tom\devel\leo\docs
    - @rst project-X-docs
        - [more of the subtree]
```
It is essential that the paths work out right for these kinds of trees.

boltex:
@tbpassin Good observation, but it's not a problem at all really: The leojs extension can be installed as-is both in vscode on a computer to edit in local drives, or, in the browser t**o edit repositories directly** (on github, azure repos, etc..) **but not local files.** As I explain below:

When leojs runs in a computer's local installation of vscode, it has access to node's path module, which is somewhat equivalent to python's path module. (for working with local files on that computer as normal/typical usage) So no problem there for writing equivalent methods in typescript that match what the original Leo in python does! :)

When leojs runs in **vscode for the browser** it works **ONLY** on a virtual filesystems of github or azure (and others) online-repository directly. It provides a shiv (an equivalent API) for node's path module. So the same things can be done no problem. 

_I guess, if you (really) wanted to work with files from your computer, but with vscode in the browser, just push your working folder as a repo! :)_

Félix

edreamleo:
@boltex @tbpassin Thanks for your comments and approving review.

I agree with Thomas that more testing is needed, especially for the botched PR #3260 (it broke `@path`).

In fact, I think we should test `devel` for at least two weeks (after merging the last path-related PR) before releasing 6.7.3.

However, *this* PR is almost certainly safe. Besides removing a workaround for an ancient Python bug, this PR makes only obviously-correct substitutions.

I'm going to merge this PR now. It should form the base of the PR that fixes #3260.

Issue #3276, opened 2023-04-10
Simplify Leo's path code


edreamleo:
Recent work has shown that Leo's path wrappers can be greatly simplified. I do not wish leoJS to faithfully transliterate the old code.

The PR shows that the global changes can be done safely and easily!

The Aha: create *new* methods (`g.finalize` and `g.finalize_join`), leaving the corresponding legacy methods unchanged.

edreamleo:
Closed via PR #3277.

PR #3277, opened 2023-04-10
PR for #3276: simplify Leos path functions

edreamleo:
See #3276.  The new code is *much* better than the old. We aren't going back.

- [x] Leave all `g.os_path` functions unchanged for compatibility, except as discussed next.
- [x] Make `g.os_path_finalize_join` a deprecated alias for the (improved) `g.finalize_join`. The old code was dubious.
- [x] Make `g.os_path_finalize` a deprecated alias for `g.finalize`. The old code did not call `os.path.expanduser`.

**Renamed functions**

- [x] Replace `g.os_path_finalize` with `f.finalize` *everywhere*.
- [x] Replace `g.os_path_finalize_join` with `g.finalize_join` *everywhere*.

These two new names emphasize the differences with the corresponding `os.path` functions.

The following sections discuss all important changes to the code.

**g.finalize**

- [x] `g.finalize` calls both `os.path.expanduser` and (new) `os.path.expandvars`:

<details>

```python
def finalize(path: str) -> str:
    """
    Finalize the path. Do not call os.path.realpath.
    """
    if not path:
        return ''
    path = os.path.expanduser(path)
    path = os.path.expandvars(path)
    path = os.path.abspath(path)
    path = os.path.normpath(path)
    path = g.os_path_normslashes(path)
    return path

os_path_finalize = finalize  # compatibility.
```

</details>

**g.finalize_join**

- [x] `g.finalize_join` expands all parts *before* calling `os.path.join`:

<details>

```python
def finalize_join(*args: Any) -> str:
    """
    Join and finalize. Do not call os.path.realpath.
    """
    uargs = [z for z in args if z]
    if not uargs:
        return ''
    # Expand everything before joining.
    uargs2 = [os.path.expandvars(os.path.expanduser(z)) for z in uargs]
    path = os.path.join(*uargs2)
    path = os.path.abspath(path)
    path = os.path.normpath(path)
    path = g.os_path_normslashes(path)
    return path

os_path_finalize_join = finalize_join
```

</details>

**c.fullPath**

- [x] Simplify  `c.fullPath` using the improved `g.finalize_join`:

<details>

```python
def fullPath(self, p: Position, simulate: bool = False) -> str:
    """
    Return the full path (including fileName) in effect at p. Neither the
    path nor the fileName will be created if it does not exist.
    """
    c = self
    # Search p and p's parents.
    for p in p.self_and_parents(copy=False):
        aList = g.get_directives_dict_list(p)
        path = c.scanAtPathDirectives(aList)
        fn = p.h if simulate else p.anyAtFileNodeName()  # Use p.h for unit tests.
        if fn:
            return g.finalize_join(path, fn) if fn else ''
    return ''
```
</details>

**c.scanAtPathDirectives**

- [x] Simplify `c.scanAtPathDirectives`using the improved `g.finalize_join`:

<details>

```python
def scanAtPathDirectives(self, aList: List) -> str:
    """
    Scan aList for @path directives.
    Return a reasonable default if no @path directive is found.
    """
    c = self
    c.scanAtPathDirectivesCount += 1  # An important statistic.
    base = c.openDirectory
    absbase = g.finalize_join(g.app.loadDir, base)

    # Look for @path directives.
    paths = []
    for d in aList:
        # Look for @path directives.
        path = d.get('path')
        warning = d.get('@path_in_body')
        if path is not None:  # retain empty paths for warnings.
            # Convert "path" or <path> to path.
            path = g.stripPathCruft(path)
            if path and not warning:  # Silently ignore empty @path directives.
                paths.append(path)

    # Add absbase and reverse the list.
    paths.append(absbase)
    paths.reverse()

    # Compute the full, effective, absolute path.
    path = g.finalize_join(*paths)
    return path
```

</Details>

**VR plugin**

- [x] Simplify `VR.get_fn` using the improved `g.finalize_join`.

<details>

```python
def get_fn(self, s: str, tag: str) -> Tuple[bool, str]:
    pc = self
    c = pc.c
    fn = s or c.p.h[len(tag) :]
    fn = fn.strip()
    # Similar to code in g.computeFileUrl
    if fn.startswith('~'):
        fn = fn[1:]
        fn = g.os_path_finalize(fn)
    else:
        # Handle ancestor @path directives.
        if c and c.openDirectory:
            base = c.getNodePath(c.p)
            fn = g.finalize_join(c.openDirectory, base, fn)
        else:
            fn = g.finalize(fn)
    ok = g.os_path_exists(fn)
    return ok, fn
```

</details>

**VR3 plugin**

- [x] Simplify `VR3.get_fn` using the improved `g.finalize_join`.

<details>

```python
def get_fn(self, s, tag):
    pc = self
    c = pc.c
    fn = s or c.p.h[len(tag) :]
    fn = fn.strip()
    # Similar to code in g.computeFileUrl
    if fn.startswith('~'):
        fn = fn[1:]
        fn = g.finalize(fn)
    else:
        # Handle ancestor @path directives.
        if c and c.openDirectory:
            base = c.getNodePath(c.p)
            fn = g.finalize_join(c.openDirectory, base, fn)
        else:
            fn = g.finalize(fn)

    ok = g.os_path_exists(fn)
    return ok, fn
```

</details>

**Remove obsolete comments**

- [x] Remove comments referring to old issues. They are now neither useful nor accurate.

### DIFF
--- a/leo/commands/checkerCommands.py
+++ b/leo/commands/checkerCommands.py
@@ -656,14 +656,14 @@ class PylintCommand:
         local_dir = g.os_path_dirname(c.fileName())
         table = (
             # In the directory containing the outline.
-            g.os_path_finalize_join(local_dir, base1),
-            g.os_path_finalize_join(local_dir, base2),
+            g.finalize_join(local_dir, base1),
+            g.finalize_join(local_dir, base2),
             # In ~/.leo
-            g.os_path_finalize_join(g.app.homeDir, '.leo', base1),
-            g.os_path_finalize_join(g.app.homeDir, '.leo', base2),
+            g.finalize_join(g.app.homeDir, '.leo', base1),
+            g.finalize_join(g.app.homeDir, '.leo', base2),
             # In leo/test
-            g.os_path_finalize_join(g.app.loadDir, '..', '..', 'leo', 'test', base2),
-            g.os_path_finalize_join(g.app.loadDir, '..', '..', 'leo', 'test', base2),
+            g.finalize_join(g.app.loadDir, '..', '..', 'leo', 'test', base2),
+            g.finalize_join(g.app.loadDir, '..', '..', 'leo', 'test', base2),
         )
         for fn in table:
             fn = g.os_path_abspath(fn)
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -1073,7 +1073,7 @@ def open_theme_file(self: Self, event: Event) -> None:
     if not c:
         return
     # Get the file name.
-    themes_dir = g.os_path_finalize_join(g.app.loadDir, '..', 'themes')
+    themes_dir = g.finalize_join(g.app.loadDir, '..', 'themes')
     fn = g.app.gui.runOpenFileDialog(c,
         title="Open Theme File",
         filetypes=[
@@ -1085,7 +1085,7 @@ def open_theme_file(self: Self, event: Event) -> None:
     )
     if not fn:
         return
-    leo_dir = g.os_path_finalize_join(g.app.loadDir, '..', '..')
+    leo_dir = g.finalize_join(g.app.loadDir, '..', '..')
     os.chdir(leo_dir)
     #
     # #1425: Open the theme file in a separate process.
--- a/leo/commands/commanderHelpCommands.py
+++ b/leo/commands/commanderHelpCommands.py
@@ -71,8 +71,7 @@ def leoDocumentation(self: Self, event: Event = None) -> None:
     """Open LeoDocs.leo in a new Leo window."""
     c = self
     name = "LeoDocs.leo"
-    fileName = g.os_path_finalize_join(g.app.loadDir, "..", "doc", name)
-    # Bug fix: 2012/04/09: only call g.openWithFileName if the file exists.
+    fileName = g.finalize_join(g.app.loadDir, "..", "doc", name)
     if g.os_path_exists(fileName):
         c2 = g.openWithFileName(fileName, old_c=c)
         if c2:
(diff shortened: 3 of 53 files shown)


edreamleo:
@boltex @tbpassin I would like to merge this PR into devel later today.

I'll take full responsibility for any problems :-)

tbpassin (changes requested):
1) "`Make g.os_path_finalize` a deprecated alias for g.finalize. The old code did not call os.path.expanduser"

If this means that `g.os_path_finalize` will work differently from before, I think this change would be wrong.  It could break other people's code.

In this case, I think the old code should stay there for the old, now-deprecated g.os_path_finalize.

A similar remark applies to `os_path_finalize_join`.

2) In leoGlobals.py, `def finalize(path: str) -> str`:
Please explain in the docstring what "finalize" is intended to do.  Users whose initials are not "ekr" won't be able to know what to expect.

3) It seems to me that the newer pathlib methods ought to be able to do all these os.path-like things and probably better.  Please see if they would be better to use.

4) In _plugins/screenshots.py_, `def finalize(self, fn)`:
Recommend changing this method name so that it is not the same as `g.finalize`.  Since this method is only used in this plugin, such a change would not affect anyone else's code.

edreamleo:

At first I thought that a new unit test for `g.finalize_join` would put your fears to rest. Instead, I find that I don't understand `os.path.abspath` and `os.path.normpath` well enough to know why the new code works at all!

It's time for a break while I gather my wits...

tbpassin:
What you say is one reason to spell out in the docsrings exactly what each of those g.xx methods are supposed to do.

boltex:
Just found a small typo as commented above, all the rest makes a lot of sense and cleans up redundant/long naming. 

I'll be sure to copy all of those changes to leojs as soon as this makes its way into devel! :)

edreamleo:
@boltex Thanks for the review. I'll fix the typo and merge.

Issue #3281, opened 2023-04-17
Syntax Coloring Does Not Color Javascript Within HTML FIle Correctly


tbpassin:
See PR #3282.

Script elements within an HTML file don't get colorized as javascript.  Instead, either they don't get colorized  or the HTML colorizer gets very confused. 

**Background (EKR)**

The `jedit2py` script in `scripts.leo` transliterated all of jEdit's XML files to python. The intention of all the python files in `leo/modes` is to simulate jEdit's colorizing rules. The python files do this in close cooperation with Leo's jEdit colorizer in `leoColorizer.py`.

`leo/modes/html.py` delegates coloring for `<script>` elements to `leo/modes/javascript.py` with the following rule:
```python
def html_rule1(colorer, s, i):
    return colorer.match_span(s, i, kind="markup", begin="<SCRIPT", end="</SCRIPT>",
        at_line_start=False, at_whitespace_end=False, at_word_start=False,
        delegate="html::javascript", exclude_match=False,
        no_escape=False, no_line_break=False, no_word_break=False)
```
Most of this rule is boilerplate. All the kwargs look good. The only important line is:
```python
    return colorer.match_span(s, i, kind="markup", begin="<SCRIPT", end="</SCRIPT>",
```
The following bugs are apparent in the above line:
- `jedit.match_span` does an exact match, so it will match neither `<script>` nor `</script>`.
- Attributes of the `script` element will be colorized as JS, not html.
- The closing `>` of the '<script>` element will be colorized as JS, not html.

Bugs can arise in various ways:
1. In the rule itself. See above.
2. Leo might not handle delegation properly.
3. In the rules in the delegated colorer, `javascript.py` in this example.

**Debugging**: The `--trace=coloring` command-line argument provides a superb trace.

edreamleo:
@tbpassin Please send me an example that illustrates the problem. Thanks.

tbpassin:
Here is the example in which I first noticed the issue.  Scroll down to around line 70.  Before that, none of the js code has been colorized.  From that point on, the colorizer is very confused.

[js-in-html-colorizer-example.zip](https://github.com/leo-editor/leo-editor/files/11261164/js-in-html-colorizer-example.zip)

edreamleo:
@tbpassin Thanks. I have the file. This issue is much more complicated than I thought. I'll get to it when I can.

edreamleo:
Closed via PR #3282.

PR #3282, opened 2023-04-18
PR for #3281: html coloring problems

edreamleo:
See #3281.

### DIFF
--- a/leo/modes/html.py
+++ b/leo/modes/html.py
@@ -104,9 +104,21 @@ def html_rule1(colorer, s, i):
         at_line_start=False, at_whitespace_end=False, at_word_start=False,
         delegate="html::javascript", exclude_match=False,
         no_escape=False, no_line_break=False, no_word_break=False)
+        
+def html_rule1_lower(colorer, s, i):
+    return colorer.match_span(s, i, kind="markup", begin="<script", end="</script>",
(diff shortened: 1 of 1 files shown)


edreamleo:
@tbpassin The fixes seem to color `js-in-html-colorizer-example.leo` correctly. Please review this PR.

tbpassin (commented):
I didn't see anything to comment on.  What a lot of work!  With so many changes and so many new conditionals, I wouldn't be able to assess if it all works without actually testing it.

edreamleo:
@tbpassin I've merge the work so far. I'll create another PR if necessary.

tbpassin:
This seems to be working reasonably.  OTOH, now I'm not very happy with the actual colors, either for js or HTML.  One thing among many, in js:  the keywords _let_ and _var_ shoulf IMHO, be the same color by they are not.  Also, too many syntax features are in blue shades, which aren't very readable (perhaps this is only apparent with my dark solarized theme...).

Alas, picking colorizer colors is really hard, and I'm not sure that I'm up for it :(.  Is there a translation table between theme colors and colorizer colors somewhere?

edreamleo:
@tbpassin You can define your own colors in `myLeoSettings.leo`.

See the node `Language-specific fonts and colors` in `leoSettings.leo` for a guide/starting point.

Issue #3285, opened 2023-04-18
Fix problems with PR #3277


edreamleo:
There were several botches in PR #3277.

See PR #3286.

PR #3286, opened 2023-04-18
PR for #3285: path problems

edreamleo:
See #3285.

- Restore `g.os_path_join` and `g.os_path_expanduser`: they should *not* finalize their results.
- Add missing alias for `g.os_path_finalize`.


edreamleo:
See #3285.

Issue #3246, opened 2023-04-03
`save_as_zipped` Holds Lock On File After Write


tbpassin:
See PR #3294.

After writing an outline with the `save_as_zipped`, Leo retains an exclusive lock on the file, such that even a text editor cannot open it.

edreamleo:
@tbpassin Thanks for this report.

I'm not happy about all the caching code in Leo. I don't understand why it exists.

Does this work for you? in `fc.saveAs`, add the following statement in the `try:` block:

    g.app.commander_cacher.close()

edreamleo:
@tbpassin It looks like the same code should be added to `fc.saveTo`. What do you think?

tbpassin:
I tried it, but I put the line above the try block, just after the         `g.app.commander_cacher.commit()`  or `save()` depending on which command we're in.

After this, my text editor was able to open the file.

I don't know if it would be better to put that code into the try block or where I put it.  It seems to me that if the cacher has done its job regardless of whether the write succeeds or fails, then it should be closed before the write gets attempted.

Or did you mean to put in _after_ the write, like this ?
```
        try:
            self.write_Leo_file(fileName)
            g.app.commander_cacher.close()
        finally:
            c.ignoreChangedPaths = False
```

I don't have an informed opinion because I don't have the slightest idea of what the cacher is supposed to be doing or whether it would get reopened when needed after a close().

edreamleo:
@tbpassin Thanks for your testing. Please file a PR so I can see exactly what you have done. But your guess is as good as mine, and you are the one actually using the command :-)

tbpassin:
So now I know what closing the cacher does:
```
    g.app.commander_cacher.commit()  # store cache, but don't close it.
  File "c:\Tom\git\leo-editor\leo\core\leoCache.py", line 67, in commit
    self.db.conn.commit()
sqlite3.ProgrammingError: Cannot operate on a closed database.
```

This was when I tried to save a second outline after the first success. Now I'll try moving the close() command to after the write.

When I opened the saved outline, it contained the content of my external files but none of their directories had been created.  When I ran the `write-at-file-nodes` command, only the selected one got written, not all of them.  Ultimately, when we get the command writing to a zip archive, then the directories will be created as it is expanded an an archive manager.

We'll have to think about what should happen when an external file has an absolute path instead of a relative one.

tbpassin:
[...]

PR #3294, opened 2023-04-22
PR for #3246: file lock on .db files

edreamleo:
See #3246.


edreamleo:
@tbpassin I do not see any problem writing a `.db` file.

The way I read the code, the fix should be to close the connection in `fc.exportToSqlite`. 

Does the fix work for you?

Issue #3268, opened 2023-04-09
Leo Does Not Return Memory After A Tab Is CLosed


tbpassin:
I have noticed that Leo doesn't reduce its memory footprint after it closes a tab, at least in Windows.  To reproduce (on Windows):

1. Open a Leo instance from a console with only the workbook open;
2. Open the Windows Task Manager program. Identify the Python instance just opened and in the _Processes_ tab note its memory use.
3. Open LeoPyRef in the same Leo instance.  Note the increased  memory use.
4. Close the LeoPyRed outline, leaving Leo open.
5. Note that the memory use doesn't decline much if any.
6. Wait some period of minutes.  Note the memory use still has not decreased.

I don't know if the Qt call deleteLater() really does remove C++ Qt memory allocations, and if it doesn't there may not be much that we can do about this.

I recall reading a post several years ago about dealing with memory leaks in PyQt.  If I can find it again and learn something useful I'll post it here.

tbpassin:
I found the link (thanks, tabbed bookmarks manager!):
[PyQt. How to shoot yourself in the foot.](http://enki-editor.org/2014/08/23/Pyqt_mem_mgmt.html)

It's interesting, probably important, but I'm not so sure that it will help with this particular issue.

tbpassin:
I'll close this now, after discussion with ekr.  There is probably nothing to be done, it doesn't seem harmful, and if possible would take a major effort with no assurance of success.

edreamleo:
@tbpassin I agree with your conclusion.

PR #3295, opened 2023-04-22
PR for #3268: memory lead after deleting log

edreamleo:
See #3268.

### DIFF
--- a/leo/plugins/qt_frame.py
+++ b/leo/plugins/qt_frame.py
@@ -2895,7 +2895,8 @@ class LeoQtLog(leoFrame.LeoLog):
         if i is None:
             return
         w.removeTab(i)
-        del self.contentsDict[tabName]
+        if tabName in self.contentsDict:
+            del self.contentsDict[tabName]
         self.selectTab('Log')
         c.invalidateFocus()
         c.bodyWantsFocus()
--- a/leo/plugins/qt_frame.py
+++ b/leo/plugins/qt_frame.py
@@ -2895,6 +2895,7 @@ class LeoQtLog(leoFrame.LeoLog):
         if i is None:
             return
         w.removeTab(i)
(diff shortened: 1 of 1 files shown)


edreamleo:
@tbpassin This one-line fix is experimental. Please take a look.

tbpassin:
It's probably a good idea, though I would have thought that the garbage collector would take care of it.  When a tab in the log frame is closed, I wonder if `deleteLater()` should be called on each of its Qt objects.  The whole `deleteLater()` thing is still mysterious to me.

But I really meant something else.  I meant the tab for an outline, rather than a tab in the log panel.  That is, the outline's tab above the menu bar.  There's probably a different name for it that I don't remember.

edreamleo:
@tbpassin Thanks for the clarification.

The leak you speak of may occur in `qtFrame.destroySelf`. This method contains a comment that it can't call qtFrame.destroyAllObjects because event handlers can't be cancelled.

I would rather not mess with this problem. There may be a leak, but I don't think it's important. Remember that because of undo Leo can *never* delete VNodes for an outline. Leo must live with those "infinite" lifetimes.

tbpassin:
I thought that undo was on a per-outline basis.  Wouldn't that allow vnodes to be deleted when the outline were closed? I don't expect undo to survive an outline close/reopen event.

edreamleo:
@tbpassin What you say is true. I wasn't clear. I meant to say that Leo already "leaks" VNodes *until* the outline is closed.

Anyway, there is no way to fix this issue unless Leo can call `qtFrame.destroyAllObjects` when closing an outline. Doing so is harder than it looks because of queued event handlers.

tbpassin:
Let's just close this one...

edreamleo:
@tbpassin I agree with closing this issue. Thanks, though, for raising this issue.

Issue #3288, opened 2023-04-21
@string theme-name = None setting and --theme=None Do Not Do The Same Thing


tbpassin:
If we put this setting into myLeoSettings.leo, we get the original unthemed Leo colors and fonts:

`@string theme-name = None `

 Some people, I think, prefer them.

However, if we specify the _None_ theme on the command line, then Leo uses the theme specified in myLeoSettings:

`leo --theme=None`

This difference is unexpected and therefore confusing.  There are times I would like to get the unthemed appearance, and I should be able to get it with the command line option for _None_, since that is what works in the settings.

edreamleo:
@tbpassin Thanks for discovering yet another qwirp :-)

tbpassin:
I am not sure what has changed , but today, looking at the code for setting themes, and trying out `--theme=None`, the issue does not appear.  With the `--theme=None`, the original no-theme appearance is used.  The code clearly does what I expected it should.

(if nothing has been changed in the theme code, it must have been operator error on my part).

So I'm closing this issue.

tbpassin:
Whoops, spoke too soon.  The command-line setting doesn't over-ride the myLeoSetting value as I had expected

PR #3300, opened 2023-04-28
LM: Make theme-name = None setting and --theme=None command line arg …

tbpassin:
…do the same thing.

Fixes https://github.com/leo-editor/leo-editor/issues/3288


edreamleo (approved):
Thanks, Thomas.

Issue #3310, opened 2023-05-03
Problems with @language md


edreamleo:
See PR #3311.

Not only is `@language md` not always colored correctly, markdown is not always colored.

Sheesh: Markdown has not correctly colored constructs like `**Bold**` and `*Italics*` since rev 5167047, dated 2021/10/24! This commit was in response to a pylint complaint, but using raw strings changed the meaning of the affected regexs.

`git bisect` reports the first bad commit is 9e2bc3a. There are significant changes to `init_mode` and these changes are likely the culprit.

edreamleo:
Fixed via PR #3311.

PR #3311, opened 2023-05-03
PR for #3310: problems with @language md

edreamleo:
See #3310.

- [x] Fix two regexs in `md.py`. These errors predate Leo 6.7.2.
- [x] Require `kind` arg to be a kwarg.
- [x] Update `self.modes[rulesetName]` in `init_mode_from_module`. This fixes the reported problem.
- [x] Improve the traces produced by the `report` function in the `setTag` method:
  - Remove the `match_name` ivar. It was updated too late to appear in the traces.
  - The `match_name` var is the caller three-levels up.

### DIFF
--- a/leo/core/leoColorizer.py
+++ b/leo/core/leoColorizer.py
@@ -732,7 +732,6 @@ class BaseColorizer:
                 s2 = repr(s[i : i + 17 - 2] + '...')
             delegate_s = f"{self.delegate_name}:" if self.delegate_name else ''
             font_s = id(font) if font else 'None'
-            g.trace(g.callers())  ###
             print(
                 f"setTag: {full_tag:32} {i:3} {j:3} {colorName:7} font: {font_s:14} {s2:>22} "
                 f"{self.rulesetName}:{delegate_s}{self.matcher_name}"
--- a/leo/core/leoColorizer.py
+++ b/leo/core/leoColorizer.py
@@ -1020,8 +1020,9 @@ class JEditColorizer(BaseColorizer):
         language, rulesetName = self.nameToRulesetName(name)
         bunch = self.modes.get(rulesetName)
         if bunch:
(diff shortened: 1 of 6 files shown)


tbpassin (commented):
1. leoColorizer, line 2086.  I'm a little concerned that some existing code may break if the "kind" arg is made a KW arg.  I like the idea in principle.

2. Same comment for line 2214.

3. General style comment:  I thought we were moving to consistently putting spaces around "=" signs in function args.  E.g., `kind = "markup"` instead of `kind="markup"`.

4. This PR only patches a couple of mode files.  What about all the dozens of other ones?  Don't they need `kind = ...` too?

These are all pretty superficial comments, and of course the proof of the pudding will be in the eating.

edreamleo:

**Points 1, 2 & 4**: Requiring `kind` to be a kwarg reverts to the status quo ante. All existing mode files specify `kind` as a kwarg. Furthermore, there is a unit test that checks the syntax of many of Leo's files, including `leo/modes/*.py`. Those tests would fail if any mode file had a problem.

**Point 3**: Beautifying all files in `leo/modes` changed only one comment in `perl.py`. These mode files almost never change, so there is no need for `leo/scripts/beautify-leo.cmd` to contain:

    call py -m leo.core.leoAst --orange --recursive leo\modes

As you say, the proof will be in the eating. I'll merge this PR and we'll all continue testing.

tbpassin:
> Requiring `kind` to be a kwarg reverts to the status quo ante. All existing mode files specify `kind` as a kwarg

Sorry, I didn't make myself clear.  I agree with your proposed changes in the "kind" arg and was only concerned in case some other code that didn't get looked at uses it the old way.

edreamleo:
@tbpassin Afaik, only leoColorizer.py uses the mode files.

PR #3356, opened 2023-06-01
PR: improve archived position methods and related unit tests

edreamleo:


leo/unittests/core/test_leoNodes.py
```diff
@@ -29,13 +29,20 @@ def setUp(self):
     def test_archivedPosition(self):
         c, fc = self.c, self.c.fileCommands
         root_p, root_v = c.rootPosition(), c.rootPosition().v
+        g.trace(root_p._childIndex)
```
boltex, line 32: This outputs a string when testing

edreamleo: Fixed in devel.

PR #3372, opened 2023-06-11
PR: Fix two comments and simplify g.isValidLanguage

edreamleo:
`g.isValidLanguage` no longer tests whether a corresponding mode file exists.


boltex (approved):
Looks good to me!  :)
only typo i see is at line 137 of leoApp.py, in the comment: 'befor' instead of 'before',
and in the leo Structure of the outline of test_leoAtFile.py, the node called 'TsetAtFile.test_findSectionName' should be 'TestAtFile.test_findSectionName' instead

Issue #3381, opened 2023-06-15
Rewrite LM.scanOptions w/o argparse


edreamleo:
See PR #3382.  Issue #3351 is the last straw.  Python's [argparse](https://docs.python.org/3/library/argparse.html) module is ludicrously complex and rigid.  Parsing `sys.argv` should be straightforward.

edreamleo:
Closed via PR #3382.

PR #3382, opened 2023-06-15
PR for #3381: rewrite LM.scanOptions w/o argparse

edreamleo:
See #3381. This project is easier than expected! It is easy to compute:

- the list of valid options from the help message.
- the list of options that require arguments. This is not possible (easily) with argparse!

See LM.computeValidOptions and LM.checkOptions.


tbpassin (commented):
Line 2660, `'qui': self.doGuiOption(),`.  Was that supposed to be `'gui':` instead?

edreamleo:
@tbpassin Yes, the key should be `gui`  not `qui`.  I'll make the required changes in a new PR.

edreamleo:
@tbpassin See PR #3385.

PR #3389, opened 2023-06-16
PR: improve options handling

edreamleo:
- [x] `findComplexOption` calls `optionError` if the regex doesn't match.
    This change simplifies all the handlers for complex arguments.
- [x] `optionError` calls `sys.exit(1)`
- [x] Improve help message.
- [x] Remove cruft.
- [x] Convert four `LoadManager` methods to functions in `leoGlobals.py`.
    Surprisingly, doing so simplifies `LM.scanOptions` a bit.

### DIFF
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -2612,56 +2612,16 @@ class LoadManager:
         """Handle all options, remove them from sys.argv and set lm.options."""
         # Save a copy of argv for debugging.
         self.old_argv = sys.argv[:]
-        #@+<< define self.usage_message >>
-        #@+node:ekr.20230615062931.1: *6* << define self.usage_message >>
-        # g.computeValidOptions uses this message to compute the list of valid options!
-        # Abbreviations (-whatever) must appear before full options (--whatever).
-        self.usage_message = textwrap.dedent(
-            """
-            usage: launchLeo.py [options] file1, file2, ...
-
-            options:
-              -h, --help            show this help message and exit
-              -b, --black-sentinels write black-compatible sentinel comments
-              --diff                use Leo as an external git diff
-              --fail-fast           stop unit tests after the first failure
-              --fullscreen          start fullscreen
-              --ipython             enable ipython support
-              --gui=GUI             specify gui: browser,console,curses,qt,text,null
-              --listen-to-log       start log_listener.py on startup
-              --load-type=TYPE      @<file> type for non-outlines: @edit or @file
-              --maximized           start maximized
-              --minimized           start minimized
-              --no-plugins          disable all plugins
-              --no-splash           disable the splash screen
-              --quit                quit immediately after loading
-              --screen-shot=PATH    take a screen shot and then exit
-              --script=PATH         execute a script and then exit
-              --script-window       execute script using default gui
-              --select=ID           headline or gnx of node to select
-              --silent              disable all log messages
-              --theme=NAME          use the named theme file
-              --trace=LIST          add one or more strings to g.app.debug.
-                                    A comma-separated list of one or more of:
-                                    abbrev, beauty, cache, coloring, drawing, events, focus, git, gnx
-                                    importers, ipython, keys, layouts, plugins, save, select, sections,
-                                    shutdown, size, speed, startup, themes, undo, verbose, zoom
-              --trace-binding=KEY   trace commands bound to a key
-              --trace-setting=NAME  trace where named setting is set
-              --window-size=SIZE    initial window size (height x width)
-              --window-spot=SPOT    initial window position (top x left)
-              -v, --version         print version number and exit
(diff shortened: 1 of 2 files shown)


edreamleo:
@tbpassin Please take a look. The code is getting simpler.

tbpassin (commented):
1. Line 2708: ` gui = m.group(1).lower()`
Won't this raise an exception if there is no `m.group(1)`?

2. Line 2711 ff: I
 ```
        elif not gui.startswith('browser') and gui not in valid:
            arg = m.group(0)
            valid_s = ', '.join(valid)
            self.optionError(arg, f"Expected one of {valid_s}")
```
don't understand why an unknown gui should get added to the list of expected gui names.

3. Line 2798: 
```
try:
            h, w = int(m.group(1)), int(m.group(2))
        except Exception:
            arg = m.group(0)
            self.optionError(arg, 'Invalid value: expected int x int')
```
I think this exception will be a ValueError.  Better to use it instead of a general Exception.  Same remark applies at line 2810ff.

4. Lines 2927ff:
```
        assert '=' in regex, repr(regex)
        i = regex.find('=')
        prefix = regex[: i + 1]
        assert '=' in prefix, repr(prefix)
```
If the first assert is True, then I don't see how the second assert can fail.  Also, I'd just use `split()` here:
`prefix, _ = regex.split('=')[0]  # Will always exist`

Overall, a good simplification.

**Comment by edreamleo** [comment]
@tbpassin Thanks for the comments.

1. Won't `gui = m.group(1).lower()` raise an exception if there is no m.group(1)?
   Can't happen: if m matches there will be an m.group(1).
2. I think all the gui names are correct.  See `app.createDefaultGui`.
3. Yes, `ValueError` is better.
4. Yes, `prefix = regex.split('=')[0]` is better.

**Comment by edreamleo** [comment]
@tbpassin Please review the latest code. I plan no further changes.

As you can see, I have created four new functions in `leoGlobals.py`. Imo, these are useful utilities for any code that want's to parse options without argparse.

**Comment by edreamleo** [comment]
@tbpassin Oops: g.optionError doesn't print the usage message. I'll have to rethink.

**Comment by edreamleo** [comment]
@tbpassin The fix was to pass the usage message to two of the helper functions. A little clumsy, but I think everything is good now.

What do you think?

**Review by tbpassin — COMMENTED** [review]
[ Looks like our replies stepped on each other.  This one was written before your post about usage came through.]

1. Lines 6347ff: 
```
def checkOptions(obsolete_options: list[str], valid_options: list[str]) -> None:
    """Make sure all command-line options pass sanity checks."""
```
I don't think this works as advertised because it breaks out of the loop the first time it finds a valid option.  So not all options get checked.

2. Lines 6363ff:
 ```
            # Do a simple check for file arguments.
            if any(z in arg for z in ',='):
                g.optionError(arg, 'Invalid file arg')

```
 On the face of it, it checks whether a argument contains either the character "," or "=".  Is this the intent?  What makes this check specific to file names?  Also, Windows allows both characters in file names, and I'm pretty sure Linux does too.

3. Lines 6384ff:
```
def findComplexOption(regex: str) -> Optional[re.Match]:
    # """Return the complex argument starting with the given prefix."""
```
Either uncomment this line and move it into the docstring, or delete it.  Note that there is no "given prefix", so the wording needs clarification.  Also, the docstring should say what a "complex argument" is.  It should also make clear the the argument string MUST start with '--'.  IOW, we do not expect to see "theme" in the option string, but "--theme".

Also, I think there should be one error message for a missing value, and a different one for a value that is present but does not match the regex.

tbpassin:
> @tbpassin The fix was to pass the usage message to two of the helper functions. A little clumsy, but I think everything is good now.
> 
> What do you think?

I'm not sure, because I don't know how and when `self.defineUsage()` gets its information, and also I'm not clear about the new stand-alone convenience functions and how they would get the right usage message.  Perhaps the usage argument should default to None or an empty string.

edreamleo:
@tbpassin About your latest comments:

1. `checkOptions` breaks out of the loop the first time it finds an *invalid* option.
2. Yes, that's the intent. File options should contain neither commas nor equal signs.
   This check is intended to ensure that file names don't get confused with option names.
   I'm not sure how strong this check is, but I think it's ok.
3. The docstring and signature have changed, so I think this problem has been resolved.
4. `LM.defineUsage` *defines* the usage message, so it doesn't "get" any info from anywhere else.

edreamleo:
@tbpassin " I think there should be one error message for a missing value, and a different one for a value that is present but does not match the regex."

It could be done, but this is small beer. There error message should suffice.

edreamleo:
@tbpassin: I think the present code addresses all the big concerns. Thanks again for the careful review.

tbpassin:
Reviews are almost always good : )

PR #3390, opened 2023-06-17
PR: Repackage all the options code

edreamleo:
- [x] Create the stateless `g.OptionsUtils` class.
- [x] `LM.scanOptions` uses  only private helper functions.
- [x] Add unit tests for `g.OptionsUtils` and `LM.scanOptions`.
- [x] Fix several bugs revealed by unit tests.

The result: simpler namespaces and simpler code.

*Note*: use Leo's git-diff-pr to see the real diffs clearly.


edreamleo:
@tbpassin I've found an even simpler way. Please review.

tbpassin (commented):
1. Line 1507: `return list(sorted(list(set(valid))))`

`sorted(set(...))` is already a sorted list, so no need for all those casts.  Just do:

`return sorted(set(valid))`

2. Line 1509
```
    def find_complex_option(self, regex: str) -> Optional[re.Match]:
        # """Return the complex argument starting with the given prefix."""
        """
        Handle the common portion of complex arguments.
        Exit if the option exists but contains no value.
        """
```
My last review comments asked for this docstring to be improved:
- Move the commented line into the dostring (uncommented) or remove it;
- Explain what a "complex argument" is;
- Explain what an "option" is. E.g, is it the name of an arg with or without the '--'?; Is it the value of that arg? Or what?
- Explain what the "common portion" is.

Generally speaking, the code certainly looks simpler (good!).

edreamleo:
@tbpassin Fixed at rev 30fe1a in this branch.

Issue #3396, opened 2023-06-18
Retire --screen-shot. Improve screenshots plugin


edreamleo:
See PR #3397.

At present `--screen-shot` crashes. However, it would be better to have `screenshots.py` do all the work:

- [x] Remove all `--screen-shot` logic from Leo's core.
- [x] Define new commands in `screenshots.py`. These wait 5 seconds and then take the screenshot.

tbpassin:
BTW, Windows 10 comes with a "Snipping Tool" which works really well.  You can activate it with <WIN-SHIFT-S>, and it copies the image to the clipboard. Or, if That's not active yet, tap the <WINDOWS> key and type snip.

edreamleo:
Closed via PR #3397.

PR #3397, opened 2023-06-18
PR for #3396: Replace --screen-shot with new commands

edreamleo:
See #3396.


tbpassin (commented):
1. Line 356 `print('pip install PIL')` - this is not the only reason imports could fail.  BTW, it's called _Pillow_ now, not _PIL_ (though I think the PIL alias may still work for imports, though not for installing).

```
C:\Users\tom>py -m pip list |find /i "pil"
Pillow                        9.5.0
C:\Users\tom>py -m pip install --user --dry-run PIL
ERROR: Could not find a version that satisfies the requirement PIL (from versions: none)
ERROR: No matching distribution found for PIL
```
2. Line 361 `@g.command('help-for-slides')` - not a good name for the command any more.  Better to add a new command, say "help-for-screenshots".

3. Docstring needs to be updated.
4. Line 441 -`print(f"Screenshot saved in ~/{file_name}")`
I dislike writing too many things to the user's home directory.  These screenshots should preferably be written to `~/.leo/screenshots`.

tbpassin (commented):
1. Line 42: This URL doesn't exist any more.  Pillow's home page is at
https://python-pillow.org

2. Lines 335, 337: 
```
# Fail fast, right after all imports.
g.assertUi('qt')  # May raise g.UiTypeException, caught by the plugins manager.
```
This idiom is used in many plugins, but I think it is in the wrong place.  If we aren't using the qt gui, why bother trying to do any imports?  We will only fail the assertion later.  I suggest putting the assert right after importing g.  This same assertion needed to be moved the other day to handle the case of the curses ui causing VR3 to emit a message about failing qt imports.  So why not do the same thing here?

edreamleo:
@tbpassin

Point 1: Fixed. Good catch.
Point 2: pylint will complain if `g.assertUI` appears before the imports.

tbpassin (approved):
Line 328 in screenshot.py: `except Exception:` -

Replace with `except ImportError`

This of course is very minor and I'll go ahead and mark the review as "Approved".

edreamleo:
@tbpassin Thanks for the approval.

I used the more general exception on purpose. Sometimes third-party tools can throw exceptions other than `ImportError`.

Issue #3312, opened 2023-05-04
diff-leo-files command cannot pick two outlines if they are in different directories


tbpassin:
See PR #3399.

PR #3399, opened 2023-06-19
PR for #3312: diff-leo-files

edreamleo:
See #3312.

- [x] Ignore `mypy_cache` folder in addition to `.mypy_cache`
- [x] Prompt again if the user gives only one file.


tbpassin (commented):
Line 750: `assert len(paths) > 1`
Please provide a message explaining why this assert failed.  `len(paths)` is too abstract for a user.

edreamleo:
The assert should never fail. I'm fine with the code as it is.

PR #3406, opened 2023-06-29
PR: Change scripts to use ruff instead of flake8 and pylint

edreamleo:
- [x] Delete set-repo-dir.cmd.
- [x] Replace `call %~dp0\set-repo-dir` with `cd %~dp0..\..`
- [x] Use `ruff` instead of `flake8` and `pylint`.
- [x] Run `reindent.cmd` from `test-leo.cmd`

### DIFF
--- a/leo/scripts/beautify-leo.cmd
+++ b/leo/scripts/beautify-leo.cmd
@@ -1,5 +1,5 @@
 @echo off
-call %~dp0\set-repo-dir
+cd %~dp0..\..
 
 echo beautify-leo
 call py -m leo.core.leoAst --orange --recursive leo\core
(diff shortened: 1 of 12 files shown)


tbpassin (changes requested):
In ruff-leo.cmd, replace "python" with "py" (2 places).  On my system, "python" launches Python 3.9, but I'm really using 3.11.  Since this is Windows only, the use of the "py" launcher should be safe.

Will there be a Leo "ruff" command similar to "pylint"?  If so, remember to add ruff to the requirements.

edreamleo:
@tbpassin Thanks for your comments. I'm not sure about a `ruff` command. Anyway, that will be a separate issue.

Issue #3404, opened 2023-06-28
PR #3215 Destroys The Workbook


tbpassin:
PR #215 creates this bug and this PR will also fix this bug :-)

As discussed in [Google Groups](https://groups.google.com/g/leo-editor/c/N374jBVLSgw), PR #3215 reproducibly replaces my workbook outline with the default CheatSheet:

<details>

This happens after I check out the branch (ekr-3181-mypy-links branch, through several updates of this branch).  When this happens, other previously-opened outlines that should have gotten opened do not.

0. **Back Up Your workbook**.
1. **Check that the backup exists** and has the right file length.
2. Close Leo.
3. Checkout the ekr-3181-mypy-links branch.
4. Launch Leo using `py -m leo.core.runLeo` (I have not tried this using any other method of launching).
5. Observe **on this first launch** that the workbook survives.
6. Close and restart Leo.
7. Observe that the workbook has changed to the default.
8. Restore the workbook from backup.  Repeat and observe replacement of the workbook.
9. Restore the workbook, check out the devel branch, and re-launch Leo.
10. Observe, **on this first launch**, that the workbook still gets replaced.
11. Restore the workbook and repeat.  Observe that this time the workbook survives untouched.
12. Check out the suspect branch again. Launch Leo. Close Leo and restore the workbook if necessary (probably not if this is the first launch).
13. Launch Leo with the path to the workbook specified on the command line. Observe that the workbook survives unchanged.

This behavior, along with the fact that other outlines should get opened but don't, make me think that Leo writes some resource that it later uses to locate the outlines to open.  In the suspect branch, this resource is wrong or corrupted, so it can't find any of the outlines.  When Leo can't find the workbook it creates a new default workbook. When an outline is specified on the command line, this resource does not get consulted, the workbook gets found, and so the workbook does not get replaced.

After switching away from the suspect branch, the resource is still corrupted so that on the first launch the workbook gets replaced.  After that, the resource is written correctly and normal behavior resumes.

I don't know what this hypothetical resource is.  I know it's not the db or the recent files list because I deleted or cleared them and the behavior continued.

```console
Leo 6.7.4-devel, ekr-3181-mypy-links branch, build c47ab83f17
2023-06-28 09:19:09 -0500
Python 3.11.4, PyQt version 6.4.3
Windows 10 AMD64 (build 10.0.19045) SP0
```

</details>

**Analysis**

[...]

PR #3407, opened 2023-06-29
PR for #3404: Remove unls from sessions logic. Make sure workbook survives

edreamleo:
**Won't Do**  I have deleted all of `leoSessions.py` instead.

- [ ] Ensure that Leo never overrides an existing notebook.
- [ ] Remove all unl-related code from the `SessionManager`.
- [ ] Use Leo's cache instead of `~/.leo/leo.session`.
- [ ] Maybe: Store both archived positions and gnxs in the cache.


edreamleo:
I am closing this PR. See the first comment.

Issue #3417, opened 2023-07-02
Deprecate all session commands


edreamleo:
See PR #3419.

As of PR #3215 Leo handles sessions as in most other editors. There is no reason for any of Leo's `session-*` commands. However, deleting these commands might cause problems for Leo's users and devs later on.

My compromise:

- Add deprecation warnings to docstrings to most docstrings in `leoSessions.py`.
  All sessions commands will remain in Leo, but I hope not in leoJS :-)
- Remove the documentation for all `session-*` commands.

edreamleo:
I am going to close this issue. Leo's session-related code will remain most as it is.

PR #3419, opened 2023-07-03
PR for #3417: deprecate all session commands

edreamleo:
See #3417.

- [x] Add deprecation comments to all appropriate docstrings.
- [x] Remove unnecessary error class.
- [x] Move the documentation for session commands from `LeoDocs.leo` to the attic.

### DIFF
--- a/leo/core/leoSessions.py
+++ b/leo/core/leoSessions.py
@@ -1,8 +1,6 @@
 #@+leo-ver=5-thin
 #@+node:ekr.20120420054855.14241: * @file leoSessions.py
 """Support for sessions in Leo."""
-#@+<< leoSessions imports  & annotations >>
-#@+node:ekr.20120420054855.14344: ** << leoSessions imports & annotations >>
 from __future__ import annotations
 import json
 from typing import Optional, TYPE_CHECKING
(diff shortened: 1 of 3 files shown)


tbpassin (commented):
1. LeoSesions.py, Line 110:
    `Leo always calls this method during shutdown`

I would prefer it the older way, only saving the session file if there were no outlines on the command line.

Also, it is possible to have a file on the command line that is not a Leo outline.  Leo will try to import such a file.  Is a session restore supposed to reload such a file, too?

Otherwise, looks OK to me.

edreamleo:

I don't think it's possible to use a Leo setting to choose whether Leo always writes session data or not.

Iirc, the OP thought that Leo was broken because Leo didn't always store session data. I don't want to back to the old way for that reason.

Always storing session data seems simplest. If you don't specify a file you always open what you were last working on.

I don't know about the edge case re importing a file. I'd like to wait and see if anyone complains.

edreamleo:
@tbpassin You have convinced me that I don't know enough to say or do anything definitive about sessions :-)

I am going to close this PR and #3417. I'll discuss a planned Easter Egg today in an Engineering Notebook post.

PR #3215, opened 2023-03-25
PR for #3181: new unls

edreamleo:
This PR fixes two bugs: #3181 (a small bug) and #3404 (a serious bug).
This PR also completes #1868 (gnx-based UNLs).

**Summary of changes**

This is the biggest little bug fix in Leo's history.

- Fix the original bug [#3181](https://github.com/leo-editor/leo-editor/issues/3181) by adding a special case in `goto.find_file_line`.
- Support *optional* gnx-based unls. See the new settings below.
  `g.findGnx`, `g.parsePathData`, and `g.openUNLFile` are powerful new resources for scripts and plugins.
- Add the `--save-session` command-line argument.
- The `convert-unls` command converts legacy (path-based) unls to gnx-based unls.
- Document clickable links in Leo's Users Guide.
- Retire the  dangerous `settings_finder` plugin.
- Retire the  `backlink` and `quickMove` plugins.
  Cross-file gnxs render these plugins obsolete.

**New Settings**

leoSettings.leo contains the following new settings, with defaults as shown.

- `@string unl-status-kind = legacy`.
- `@bool full-unl-paths = True`.
- `@data unl-path-prefixes` (with only comments in the body).

With these settings, Leo should treat unls *exactly* as before.

<details><summary>Code-level details</summary>
<br>

**Most important methods**

- `g.openUNLFile` opens the commander for a unl using the data in `@data unl-path-prefixes`.
- `g.parsePathData` creates a dictionary from the `@data @data unl-path-prefixes` setting.
- `g.findUnl` (aka `g.findUNL`) resolves legacy (path-based) unls (to a position).
- `g.findGnx` resolves new (gnx-based) unls.
- `g.handleUNL` opens other outlines using the file part of a legacy unl.
- `p.get_UNL` returns the (clickable) unl for p. This is not the same as p.gnx!

*Note*: Leo's `@auto` code creates and uses links similar to Leo's legacy unls.
However, the `PersistenceDataController` class uses neither `p.get_UNL` nor `g.findUnl`.

**Code-level changes and cleanups**

- [x] Update `SM.load_session` so it handles new unls correctly.
- [x] Call `urllib.parse.quote` in `LeoQtLog.put`. Call `urllib.parse.unquote` in `LeoQtLog.linkClicked`.
    Remove url/unl hacks in `LeoQtLog.put`.
- [x] Do *no* gui-related operations in `g.findUNL`.
- [x] Remove `link_root` kwargs and related ivars.
[...]

PR #3424, opened 2023-07-04
PR: tweaks to Leo's new unls code

edreamleo:
Small changes to PR #3215.

**Compatibility Guide**

`p.get_UNL` previously returned path-based unls. It now returns gnx-based unls:
**Pros**: existing code will start using gnx-based unls automatically.
**Cons**: Plugins that must use path-based unls will have to change. See below.

Plugins writers should do the following to adapt to the changes made in PR #3215 (and this PR). 

**Case 1**: The plugin will create and read (support) only legacy unls:

Instead of calling `p.get_UNL`, call one of `p.get_legacy_UNL`, `p.get_full_legacy_UNL`, or `p.get_short_legacy_UNL`.
That's all. There is no need to call `g.findAnyUnl` instead of `g.findUnl`.

**Case 2**: The plugin will *create* gnx-based unls, but will *read* all kinds of unls:

Call `g.findAnyUnl` instead of `g.findUnl`.
That's all. There is no need to change calls to `p.get_UNL`.

<details><summary>Summary of <em>this</em> PR</summary>
<br>
This PR makes *no* substantive changes to `leoNodes.py`.

- [x] `g.openUNLFile` does a quick test to match within c.
- [x] `g.openUNLFile` returns None if no match was found.
- [x] Change default: `@string unl-status-kind = gnx`.
- [x] Change default: `@bool full-unl-paths = False`.
- [x] Apply tbrown gnx `#@+node:tbrown.20111010104549.26758: *4* p.get_UNL` to `p.get_full_legacy_UNL`. See below.

Terry's Original Code

Here is the node from the `master` branch. The gnx originally applied to `p.get_UNL`, but `p.get_full_legacy_UNL` now corresponds most closely to Terry's original code. I moved the code because `p.get_UNL` now returns gnx-based unls. 

```python
#@+node:tbrown.20111010104549.26758: *4* p.get_UNL
    def get_UNL(self) -> str:
        """
        Return a UNL representing a clickable link.
        See the section < define global error regexs > for the regexes.

        New in Leo 6.6: Use a single, simplified format for UNL's:

        - unl: //
        - self.v.context.fileName() #
        - a list of headlines separated by '-->'

        New in Leo 6.6:
        - Always add unl: // and file name.
        - Never translate '-->' to '--%3E'.
        - Never generate child indices.
        """
        base_unl = (self.v.context.fileName() + '#'
            + '-->'.join(list(reversed([z.h for z in self.self_and_parents(copy=False)])))
                    )
        encoded = base_unl.replace("'", "%27")
        return 'unl://' + encoded
```

</details>

### DIFF
--- a/leo/config/leoSettings.leo
+++ b/leo/config/leoSettings.leo
@@ -11667,11 +11667,7 @@ False: do nothing when the mouse hovers over nodes.</t>
 <t tx="ekr.20230627130611.1">gnx:    Show gnx-based unls in the status area.
 legacy: Show legacy (headline-based) uns in the status area.</t>
 <t tx="ekr.20230627131213.1"></t>
-<t tx="ekr.20230627131227.1">This setting has effect only when `@string unl-status-kind = legacy`.
-
-Regardless of this setting, Leo will resolve unls using `@data unl-path-prefixes`.
-
-True: (legacy): Display full path names in the status area.
+<t tx="ekr.20230627131227.1">True: (legacy): Display full path names in the status area.
 False:          Display only the .leo file's name in the status area.
 </t>
 <t tx="ekr.20230627202839.1"># lines have the form:
--- a/leo/config/leoSettings.leo
+++ b/leo/config/leoSettings.leo
@@ -11667,9 +11667,12 @@ False: do nothing when the mouse hovers over nodes.</t>
 <t tx="ekr.20230627130611.1">gnx:    Show gnx-based unls in the status area.
 legacy: Show legacy (headline-based) uns in the status area.</t>
 <t tx="ekr.20230627131213.1"></t>
-<t tx="ekr.20230627131227.1">True: (legacy): Display full path names in the status area.
+<t tx="ekr.20230627131227.1">This setting has effect only when `@string unl-status-kind = legacy`.
+
+Regardless of this setting, Leo will resolve unls using `@data unl-path-prefixes`.
+
+True: (legacy): Display full path names in the status area.
 False:          Display only the .leo file's name in the status area.
-                Use @data unl-path-prefixes to resolve names.
 </t>
 <t tx="ekr.20230627202839.1"># lines have the form:
 # x.leo: &lt;absolute path to x.leo&gt;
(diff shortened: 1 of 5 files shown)


edreamleo:
@tbpassin Please test the ekr-tweak-unls branch. I won't merge this PR until I get your approval.

tbpassin:
I'm looking at the code and will check it out.  But before that, I have a question.  Since only the basename is checked for a match with the current outline, there would seem to be no way to distinguish two outlines of the same name but in different locations. 

I don't think that's acceptable.  I can easily picture cases where one wants to work on two versions of an outline at the same time, if only to compare them.  Leo can of course open them, but not by UNLs with this version of the branch, it seems.

Why not compare the full paths, instead?

edreamleo:
@tbpassin You're right about being unable to distinguish files with the same name on different paths. However, this is not sufficient reason to put full paths in all unls, of any type.

I've never encountered this problem and don't think it's worth solving. The obvious workaround is to rename one of the files.

edreamleo:
@tbpassin I've just added a compatibility guide to the first comment of this PR.

Please let me know whether you find this guide helpful and correct. Thanks.

tbpassin:
I'm confused because the changes I see in the GitHub diffs display are not the same as the ones I see in the actual files, now that I've just fetched the most recent changes.  I wonder if there is a time delay.

I will have to wait to finish my review until GitHub catches up or whatever is going on stops going on.

edreamleo:
@tbpassin The diffs apply only to the `ekr-tweak-unls` branch.

To see the actual file, select 'View file' from the menu at the top left (on the diffs page). You can double-check the actual definition of, say, `g.openUNLFile` by searching for `def openUNLFile`.

You should see exactly the same thing in Leo when you are on the `ekr-tweak-unls' branch.

edreamleo:
@tbpassin I've merge this branch into devel so that more people can use and test the code.

I delayed 6.7.4 so that we all will have more time to test the code thoroughly.

Please report any problems, new or old.

Issue #3429, opened 2023-07-06
paste-node should retain gnxs if possible


edreamleo:
See PR #3430.

@tbpassin proposes that copy-node or cut-node followed by paste-node shall work as follows.

**Case 1**: the gnx of the pasted node does **not** exist in the target outline:

The pasted node should retain its gnx (as in paste-node-as-clone).

**Case 2**: The gnx of the pasted node *does* exist in the target outline:

The pasted node must get a new gnx.

**Notes**

- paste-as-clone will work as before&ndash;it retains the gnx, creating a clone if necessary.
- EKR: Leo must now have a `paste-reallocating-gnxs` command to force reallocation!

tbpassin:
As I thought more about it, I realized that what applies to a node should apply to its subtree, since the entire subtree gets copied/pasted.

edreamleo:
I am going to close this item, for reasons given [here](https://groups.google.com/g/leo-editor/c/_wAzGsUJcxU/m/WL36jyKKAwAJ).

PR #3430, opened 2023-07-06
PR for #3429: paste retains gnxs if possible

edreamleo:
See #3429.

- [x] Add `fc.preReadLeoOutlineFromClipboard`.
  This **preread method** contains the opening line from `fc.getLeoOutlineFromClipboard`.
  Duplicating code in this method minimizes changes everywhere else.
- [x] `c_oc.pasteOutline` looks ahead using the preread method, choosing its course of action accordingly.
- [x] Create a `paste-node-reallocating-gnxs` command to force reallocation.

### DIFF
--- a/leo/commands/commanderOutlineCommands.py
+++ b/leo/commands/commanderOutlineCommands.py
@@ -57,7 +57,7 @@ def cutOutline(self: Self, event: Event = None) -> None:
         c.copyOutline()
         c.deleteOutline(op_name="Cut Node")
         c.recolor()
-#@+node:ekr.20031218072017.1551: *3* c_oc.pasteOutline
+#@+node:ekr.20031218072017.1551: *3* c_oc.pasteOutline & helper (changed)
 @g.commander_command('paste-node')
 def pasteOutline(
     self: Self,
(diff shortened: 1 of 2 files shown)


edreamleo:
@tbpassin Ready for review! Make backups first :-)

edreamleo:
@tbpassin You mentioned somewhere just now about checking the entire subtree for gnx clashes.

The private helper function `anyGnxClashes` does check the entire tree. But I'm not sure that's the desired behavior!

tbpassin:
> You mentioned somewhere just now about checking the entire subtree for gnx clashes.

Maybe the check would not be needed for all the subtree nodes after all.  I'm mostly thinking about an organizer node that includes cloned and non-cloned descendants in its subtree. 

I can now see that the concept of operation is simple. Within an outline, for  a parent-and-subtree: 
- Ask for a move, get a move;  gnxs do not change;
- Cut-paste is the same as a move;
- Ask for a copy, get a copy, meaning all nodes get new gnxs.
- To paste a copy when the top of the copied tree already exists means *by definition* to ask for a copy. 

I think this is fairly intuitive.  What other behavior would an ordinary user expect?

Note that copying a node and then pasting it into a different outline would appear the same as a move, so gnxs would not change.  Is this also intuitive?  I'm not so sure about this case.  I think that a cross-outline paste would be expected to be a copy.  But I don't see how the paste code can know what outline the clipboard contents came from.

Need to think about this one a bit more!

boltex:
I like this new feature, i've looked at the code Edward wrote. I also wondered if he checks only the top node against all gnx of the receving outline, or if he checks all nodes in outline being pasted  against all gnx of the receiving outline...

...I've thought about it for a while: I think it's ok to check only the top node of the outline being pasted against all gnx, no need to check all inner nodes also. 

It gets the job done: cut + paste will be equivalent to a move. which i like. ...also, copy+paste in another outline will preserve gnx, which is also great! 

Thanks to you both for this cool behavior tweak!! cant wait to alter leojs to match this! 😄

edreamleo:
@boltex Thanks for your comments. I am going to experiment with various checks. Also, undo needs work.

tbpassin:
> But I don't see how the paste code can know what outline the clipboard contents came from.

It would be easy if the XML format in the clipboard could be changed to include the source file.  I'm not sure how the xml is processed, but the obvious possibilities would be a new attribute, perhaps in the `<leo-header>` element:

<leo_header file_format="2" source-file="c:\test\leo\cutpaste.leo"/>

Or a new element after the leo_header element:
`<source file="c:\test\leo\cutpaste.leo"/>`

Or perhaps a processing instruction:
`<?source file="c:\test\leo\cutpaste.leo"?>`

Depending on how the xml in the clipboard is processed, one or another of these forms should be backwards compatible with existing code.  The paste code would check to see if the source outline equaled the receiving outline.  if it does, the paste code would check the parent gnx as proposed in my last comment.  If not, the paste operation would be automatically be a copy, i.e., create new gnxs.

If the operation were "paste-as-clone" then gnxs would ge retained as is.  This would let me copy parts of a zettel-kasten without destroying their cross-links, just as the current code does.

edreamleo:
@tbpassin @boltex I'm unlikely to accept this PR. Not now. Not ever.

The distinction between copy/paste and cut/paste is way too cute. It depend on history-dependent state.

Issue #3416, opened 2023-07-02
​refreshFromDisk may work for the wrong reason


edreamleo:
See PR #3433.

From Félix:

​In ​​refreshFromDisk, near the very beginning of that method, there is a check to see if there's a filename, and that the filename is not a 'folder' instead of an actual file. 
```python
    if not fn:
        g.warning(f"not an @<file> node: {p.h!r}")
        return
    # #1603.
    if os.path.isdir(fn):
        g.warning(f"not a file: {fn!r}")
        return
```
When it's checking for the fn string to be a possible 'directory', it's checking with the filename only. eg. myFile.txt, which is a non-absolute path/string...  instead of what a call to c.fullPath would provide. e.g. C:\myFolder\myFile.txt. 

(which is what is used in leoAtFile.py's methods read, readOneAtAutoNode, readOneAtCleanNode, readOneAtEditNode when they want to have a full path string of the intended file to work with.)

So my concern is that the check in refreshFromDisk done with os.path.isdir(fn) always returns false, which still makes most every case unintentionally work fine, work for the wrong reason.

tbpassin:
I think the check should be
```
if not os.path.isfile(full_file_path):
        g.warning(f"not a file: {full_file_path!r}")
        return
```

instead - using the full path, of course.

edreamleo:
@boltex @tbpassin PR #3433 fixes two other quirps :-)

edreamleo:
Closed via PR #3433.

PR #3433, opened 2023-07-09
PR for #3416: quirps in refresh-from-disk and two other methods

edreamleo:
See #3416.

- [x] Check `c.fullPath(p)` when checking to see if `p` (an `@<file>` node) refers to a file.
- [x] Remove the `fn` var completely. Check `p.isAnyAtFileNode()` instead.
- [x] Remove the unused `fn` *positional* arguments in `at.readOneAtEditNode` and `at.readOneAtAsisNode`.
   In both methods, `fn = c.fullPath(p)` immediately replaces the `fn` arg!

### DIFF
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -310,8 +310,8 @@ def refreshFromDisk(self: Self, event: Event = None) -> None:
     if not fn:
         g.warning(f"not an @<file> node: {p.h!r}")
         return
-    # #1603.
-    if os.path.isdir(fn):
+    full_path = c.fullPath(p)
+    if os.path.isdir(full_path):
         g.warning(f"not a file: {fn!r}")
         return
     b = u.beforeChangeTree(p)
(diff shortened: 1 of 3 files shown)


tbpassin (changes requested):
I would like there to be no more changes to signatures of long-established methods and functions.  You have no idea who else is using these calls.  One of Leo's principles has been that any part of the leo code can be called by any plugin or script. This means that it's all part of the API.  If there were a published API that included only a subset of Leo's calls, then it would be fine to change ones not in that subset.  But that's not the case.

In this PR it would have been simple to create a new method with a new name and call signature.  It's not too late to do this.

Summary: I won't approve this PR as is because it changes the API.

edreamleo:
@tbpassin I basically only care whether *you* use `at.readOneAtEditNode` and `at.readOneAtAsisNode` :-)

I understand your complaint, but I don't think it's worth delaying the change.  Rev 103f5c5 of the `6.7.4` documents the change.

Also, Félix should remove the `fn` positional arg in leoJS.

boltex (approved):
This is a small break in the API because of the signature change, and should be documented as such in the **'breaking changes'** section of the **release notes/readme** (as is the case when such things happen. )

edreamleo:
@tbpassin I'll be sure to document the change in the 6.7.4 branch.

tbpassin (commented):
I didn't really find anything to remark on, except for the signature change.

edreamleo:
@tbpassin The release notes now prominently mention the changes to Leo's API.

Issue #3435, opened 2023-07-09
undo/redo 'refresh-from-disk' doesn't work


boltex:
Closed via PR #3438.

<details><summary>Original Report (edited)</summary>
<br>

To test this behavior, I tried making small text changes *without* changing the structure of the outline in external files. 

**Changes to c.p.b**

For both `@clean` and `@file`:

- `refresh-from-disk` does not update `c.p.b`.
- Undo/redo of `refresh-from-disk` do not undo/redo `c.p.b`.

**Structural changes**

Everything works as expected:

- `refresh-from-disk` for `@file` updates structural changes.
- Undo/redo `refresh-from-disk` work for `@file`.
- `@clean` never updates structure, so there is nothing to test.

</details>

**Summary of changes**

- `refresh-from-disk` and several `read-at-x-nodes` commands are no longer undoable.
  As a result, *the original report is moot*.
- All non-undoable commands and scripts should clear Leo's undo state by calling `u.clearAndWarn`.
- All commands that previously used `u.before/afterChangeTree` now use `u.before/afterInsertNode`.
  This works because the commands only insert one top-level node!

edreamleo:
Closed via PR #3438.

PR #3438, opened 2023-07-10
PR for #3435: undo for refresh-from-disk

edreamleo:
See #3435. See [this thread](https://groups.google.com/g/leo-editor/c/6Cs7JDZ-Rfk).

- [x] Remove `u.saveTree`, `u.before/afterChangeTree`, and related methods.
  I do not understand how those methods could handle clones properly.

**Summary of changes**

- `refresh-from-disk` and several `read-at-x-nodes` commands are no longer undoable.
- All non-undoable commands and scripts should clear Leo's undo state by calling `u.clearAndWarn`.
- All commands that previously used `u.before/afterChangeTree` now use `u.before/afterInsertNode`.
  New unit tests cover these commands and their undo/redo logic.

<details><summary><b>Remove cruft from refresh-from-disk</b></summary>
<br>

- [x] Remove unused `refresh_pattern` regex.
- [x] Remove weird dependence on `c.sqlite_connection`.
- [x] Use `p.isX` methods to test headlines.
- [x] Remove the `redraw` flag.

</details>

<details><summary><b>Add new utils</b></summary>
<br>

- [x] Add `u.clearAndWarn`, to be called by all undoable commands.
- [x]  `gdc.create_root` and `loc.create_root` now call `u.beforeInsertNode`.
   These two methods return the undoData for later use.
- [x] Add bespoke helpers for `parse-body`:
    `u.beforeParseBody`, `u.afterParseBody` and related methods.

</details>

<details><summary><b>Make 4 commands undoable</b></summary>
<br>

The `refresh-from-disk`, `read-at-auto-nodes`, `read-at-file-nodes`, `read-at-shadow-nodes` commands are one-way updates. Undoing such commands would be dangerous and error prone.

- [x] Replace the undo logic with a call to `u.clearAndWarn` in the above commands.

</details>

<details><summary><b>Revise undo logic for 7 commands</b></summary>
<br>

- [x] Make `parse-body` undoable using  `u.beforeParseBody` and `u.afterParseBody`.
- [x] Change the following 6 commands so they call `u.before/afterInsertNode`:
    `diff-marked-nodes`, `gdc.diff_revs`, `gdc.diff_two_branches`, `gdc.diff_two_revs`,
    `loc.diff_list_of_files`, `ric.run`.

</details>

<details><summary><b>New unit tests</b></summary>
<br>

- [x] Add unit tests for all (still undoable) commands whose undo/redo logic has changed.
  - [x] Add 2 tests to the `TestCompare` class: `test_diff_list_of_files` and `test_diff_marked_nodes`.
  - [x] Add 3 tests to the `TestEditFileCommands` class: `test_diff_two_branches`, `test_diff_two_revs`, `test_git_diff`.
  - [x] Add 2 tests to the `TestLeoImport` class: `test_parse_body` and `slow_test_ric_run`.

</details>


edreamleo:
@boltex @tbpassin I'll be interested in your comments :-)

I wonder how much undoable code remains that should call `u.clearAndWarn` but doesn't. Ah well. this PR is a start.

tbpassin:
> * Add `u.clearAndWarn`, to be called by all undoable commands.

I think this was intended to read "all **non-undoable** commands".  This apparent editorial error appears in a number of places.

tbpassin (commented):
1. abbrevCommands.py, Line 542 `def replace_selection` -
 
I think this same command exists elsewhere in the core (possibly with another name).  Wouldn't it be better to reuse it?

2. commanderFileCommands.py, line 342 `c.undoer.clearAndWarn('refresh-from-disk')` -

It occurs to me that if one of the several lines ahead of this call were to fail, then the `clearAndWarn() call would never get issued.  Since it would seem to be important for it to be called, I wonder if it should be moved earlier.  A similar configuration occurs at some other places, too.

3. editFIleCommands.py, lines779:
```
        u, undoType = c.undoer, 'diff-two-branches'
        if not self.get_directory():
            return
```
Move first line to after the `return` statement (it won't be needed it we return here).

4. leoImport.py, line 1092 - 1094 -
```
        c = self.c
        d = g.app.language_extension_dict
        u, undoType = c.undoer, 'parse-body'
        if not p:
            return
```
Move the first 3 lines after the return.

This is the most review I can do given that the real changes to undo code and import code are too complex to digest without a major time commitment.

edreamleo:
@tbpassin 

Point 1: The abbreviation code is too complex to change, so I won't.

Points 2, 3, 4: My common practice is to create a "bindings" section at the start of each method, even if some of them don't get used before the method returns.

boltex:
Thanks for all of this! I'll now make sure leojs has those changes too! :)

edreamleo:
@boltex You're welcome. It's good to be appreciated :-)

Issue #3005, opened 2022-12-13
Revise or delete load-leo.html


tbpassin:
See PR #3441.

**Temporary solution**

The PR rewrites `load-leo.html` to say that vscode.dev will (eventually!?!) support leoInteg.

**Original report**

In Leo's Readme om Github, there is a link near the bottom : [A web page that displays .leo files](http://leoeditor.com/load-leo.html). Trying to load the default outline offered, my browser gave an error "File is too large":

```
Upload from the Internet.
Enter the URL below.
https://github.com/leo-editor/leo-editor/raw/master/leo/doc/LeoDocs.leo

File is too large

Error uploading file
```

A smaller outline should be found for this demonstration.

edreamleo:
@tbpassin I see the behavior you describe.  Notes:

- Loading `LeoDocs.leo` *from disk* does work, so the problem is something other than the file's size.

- I also get the "too large" message when trying to upload a smaller file, say `leo/test/test.leo`.

**Conclusion**: `load-leo.html` is buggy in a way I do not understand. I'll investigate, but no promises.

edreamleo:
Fixing the JS is beyond me at present.

Someone else will have to fix this issue.

tbpassin:
I think the error is coming from the php file _do-leo.php_.  The error message wording appears in lines 635 and 683.

tbpassin:
Currently, the php file mungs the LeoDocs.leo outline to insert a stylehseet PI (Process Instruction).  If the stylesheet PI (_<?xml-stylesheet ..._) could be permanently put into LeoDocs.leo, there would be no need for the php file, jquery, or any of that.  When a browser loads the .leo file, it would automatically be displayed using the xslt stylesheet.

If the current code that opens a Leo outline in Leo cannot accept (and ignore) a stylesheet PI, then I would regard that as a bug that should be fixed.  (I have not yet tested for this).

tbpassin:
I added a stylesheet PI to my workbook outline and Leo opened it without a problem.

However, when I opened the same outline in the browser from the filesystem, the xslt stylesheet didn't get applied.  That surprises me because it does get applied when the LeoDoecs.leo outline is downloaded from the internet (i.e., the Leo site).

edreamleo:
@tbpassin Thanks, Thomas, for these comments. With your permission, I'll delegate this issue to you.

tbpassin:
[...]

PR #3441, opened 2023-07-16
PR for #3005: load-leo.html

edreamleo:
See #3005.

- [x] Rewrite  `load-leo.html`, referencing `vscode.dev`.


edreamleo:
@tbpassin @LewisNeal This PR is ready for review. I would appreciate your comments and testing.

tbpassin (changes requested):
1. Loading the link in README.md for load-leo.html into the browser fails.  This is because load-leo.html is looking for the two newly copied scripts in a _scripts_ directory:
 ```
<script type="text/javascript" src="scripts/upclick-min.js"></script>
<script type="text/javascript" src="scripts/jquery.min.js"></script>
```
but they are actually in a _docs_ directory.

2. Why are these scripts in two places, _leo-editor/docs_ and _leo-editor/tools/webroot_?  They should only be in one location.

edreamleo:
@tbpassin Thanks for your testing, Thomas.

I don't know why the scripts are in two places. Perhaps they get copied as part of the distribution process.

Good catch about the scripts directory. I'll create the directory and test for myself.

edreamleo:
Still no joy after moving the .js file and deploying to Leo's website.

I get `405 Not Allowed` when trying to upload a local file.

Nothing happens when I try to open https://github.com/leo-editor/leo-editor/raw/master/leo/doc/LeoDocs.leo

tbpassin:
I don't know how this can be tested on the github.io site without being in the branch that gets deployed by the io machinery.  I assumed that the directory structure on the .io site is the same as the file tree.  If that's not right the relative path to the scripts would have to be adjusted.

I also don't know how to examine the directory structure on the .io site since it's only addressable by a browser AFAIK.

OTOH, not being able to upload a local outline is not the same problem as the original issue.

Personally I have never saw this particular capability as making much sense.  Someone who doesn't know Leo would look at the generated page and just see an expandable tree like a million others.  What makes this Leo version different?  One can't do anything with it.  Plus it looks crude, which is not a strong selling point.

A Leo user wouldn't have much use for it either.  So I would just retire the link altogether.

edreamleo:
@tbpassin Great minds thinking alike! The way-way better alternative is vs-code in the browser! https://vscode.dev/

Works only with Chrome and Edge. Also, leoInteg doesn't yet work with vscode.dev, but when it does work it will blow the old page away.

PR #3444, opened 2023-07-16
PR: Quirp in editFileCommands not properly fixed

edreamleo:
- [x] Create m.group(1) the `section_pat` regex.

### DIFF
--- a/leo/commands/editFileCommands.py
+++ b/leo/commands/editFileCommands.py
@@ -38,7 +38,7 @@ class ConvertAtRoot:
     errors = 0
     root = None  # Root of @root tree.
     root_pat = re.compile(r'^@root\s+(.+)$', re.MULTILINE)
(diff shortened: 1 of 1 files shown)


edreamleo:
@boltex Please take a look :-)

boltex (commented):
The one just before the make_clones  method was this:
section_pat = re.compile(r'\s*<\<(.*)>\>')

The original had (.*) instead of (.+) which is **match zero or more of the preceding character** instead of **one or more of the preceding character** so perhaps that was also an important thing in the proper regex which was used.

edreamleo:
@boltex Yes, there is a change.  Imo `.+' is better than `.*`.

boltex:
Oh ok great then, i'll approve it

boltex (approved):
As per your last email,  (.+) is even better :)

PR #3462, opened 2023-07-25
add private repo support for func get_all_issues

HaveF:
The interface of the function is not changed here, but the environment variable is simply used to support private repo


edreamleo (approved):
Thanks for this PR. My apologies for delay in approving it.

Issue #3476, opened 2023-08-01
diff-marked-nodes creates two undo beads (instead of one.)


boltex:
Fixed via PR #3478. The culprit was a duplicate call to `u.afterInsertNode`.

Trying to do 'undo' after issuing this command works, (it removes the main diff node it created) but, obviously, undoing past that point creates an erroneous state eventually (as it tries to remove an _already removed_ node )

edreamleo:
Fixed via PR #3478.

PR #3478, opened 2023-08-02
PR for #3476: diff-marked-nodes

edreamleo:
See #3476.

- [x] Modify an existing unit test for `diff-marked-nodes` so it fails with legacy code and passes with the new code.
- [x] Remove redundant call to `u.afterInsertNode`.


boltex (commented):
Looks good ! 😄

edreamleo:

PR #3484, opened 2023-08-08
PR: Remove updateBodies

edreamleo:
- [x] Remove `updateBodies`.
- [x] Reorganize test classes.
  - All test classes should be subclasses of either `unittest.TestCase` or `LeoUnitTest`.
  - `LeoUnitTest` should contain all setup utilities.
- [x] Fix failing unit test by disabling dubious test code.


edreamleo:
@boltex Please review this PR.

I got the failing unit test to pass by commenting out the check on `p.b`. I think that's all we can do for now.

PR #3487, opened 2023-08-12
Fixed 'revert' command's condition reaction and message.

### LINKED ISSUE #3486: 'Revert to saved' command shows inconsistent messages when untitled/unsaved.
It shows both lines:
can not revert unnamed file.
Can not revert unsaved file: None

It should have returned after the first message. 
Capitalization should be more consistent between those two messages.

NOTE: not a bug per say as it doesn't break anything.

boltex:
Fixes #3486


edreamleo (approved):
Thanks for this.

PR #3489, opened 2023-08-13
Fixed docstring of tag-node and tag-children.

### LINKED ISSUE #3488: tag-node and tag-children docstring contain the command itself as prefix.
They should start directly by describing the command, like all other command docstrings.  (instead of starting by re-stating their name) 

This will make command listing in the minibuffer (and menus, etc.)  cleaner.

boltex:
Fixes #3488


edreamleo (approved):
I like this docstring pattern for commands.

Issue #3492, opened 2023-08-14
Retire the read-outline-only command


boltex:
See PR #3499.

Original title: `read-outline-only` opens with incomplete interface.

Leo's devs agree that this command may be dangerous.  It's of little use.

<details>

The command opens the file without reading external files, as its docstring states, but, **the interface is lacking the top menu**, and right-click **context menu does not open over outline nodes**, and, some tabs are not present below alongside the find panel. 

(On the other hand, the Keyboard commands work. So its still possible to edit the outline and body pane)

See screenshot below. _I tried both ubuntu 20 and windows 11, and got this same result.

![image](https://github.com/leo-editor/leo-editor/assets/847716/1c8bda2d-238a-49f9-a210-c3d0da7dea69)

</details>

boltex:
@edreamleo I found this while trying to add it to both leointeg & leojs. Not sure how to go about this, so I'll wait until you take a quick look at it, and I'll focus on some other commands/features in the meantime.

boltex:
@edreamleo Maybe this was supposed to create a new, untitled commander, and just fill up the outline that readOutlineOnly creates from the Leo file. (And stay new&untitled after being opened, as to not conflict with a potential already opened document with the same name, which Leo usually prevents when trying to open an already opened file.)

edreamleo:
@boltex Not sure what `read-outline-only` should do. Or why the command should exist.

edreamleo:
@boltex I have reassigned this issue to 6.7.5. The code is too complex to change for 6.7.4.

boltex:
@edreamleo Very good. no problem at all.
In the meantime I'll comment it off in leoInteg and LeoJS until this is resolved (or not if  the method is considered deprecated.) 
It can also be somewhat dangerous to try, because if the user uses it and then presses ctrl+s (save), this will overwrite his external files with empty files.

edreamleo:
@boltex Thanks for your comments. I am leaning towards removing this command entirely.

edreamleo:
Closed via PR #3499.

PR #3499, opened 2023-08-18
PR for #3492: retire read-outline-only

edreamleo:
See #3492.


boltex (changes requested):
The 'read-outline-only' entry listed in leoserver's "good commands" should be removed also :) 
Otherwise this looks good to me 👍

edreamleo:
@boltex Good catch.

Issue #3503, opened 2023-08-19
(consecutive) Backslashes in Find box are escaped passed the first one.


boltex:
Fixed via PR #3507.

The find result is unexpected because they are escaped: 
To reproduce: try finding the following text `a \ b` then try this one: `a \\ b`  in a leo file containing those three text entries.

```
a \ b
a \\ b
a \\\ b
```

_Found when testing find input box with consecutive backslashes in LeoInteg!_ 😄

edreamleo:
Closed via PR #3507.

PR #3507, opened 2023-08-22
PR for #3503: back slashes in search patterns

edreamleo:
See #3503.


edreamleo:
@boltex The fix appears straightforward. I'll wait for your review before merging.

PR #3527, opened 2023-09-02
Fixed all possibly unpaired afterChangeGroup.

### LINKED ISSUE #3526: beforeChangeGroup and afterChangeGroup need to be paired. Conditional 'afterChangeGroup' produce bugs.
_(Discovered when testing mark-node-and-parents with ALREADY marked nodes, thus creating no marks, and not going in the condition 'changed' explained below)_

The documentation in cheatsheet.html is wrong, It shows that the afterChangeGroup is optional: this will produce an error when trying to undo a group command that was not finished with afterChangeGroup it the 'changed' variable is false.:

```
     if changed:
        u.afterChangeGroup(c.p, undoType,
```

I can see in the undo code that afterChangeGroup used to actually add the bead. (this is now in an 'if 0' branch which was added 16 years ago. ) so perhaps afterChangeGroup could be called conditionally, or, even with no commands to be undone between the 'before' and 'after'. But it's not the case anymore.

**What I propose:**  see note below this comment (_making the before on the same condition as the after would fix everything)_

 ~~1- We go through all the codebase and pull out the afterChangeGroup calls out of conditions, making those calls unconditional, thus always **finishing methods with a afterChangeGroup, if a beforeChangeGroup was called**, as to make them always 100% paired throughout Leo's codebase. (_an exception: if the 'beforeChangeGroup' was conditional, then the 'afterChangeGroup' can also be conditional if it's applied on the same condition_)~~

~~2- We modify the afterChangeGroup method to check if the group of command to be undone is empty. If so, we remove the bead that beforeChangeGroup already prepared in waiting for the afterChangeGroup to happen. (**perhaps just decremeting the bead pointer will suffice**) As to not create an undo-group of zero actions.~~

boltex:
Fixes #3526

### DIFF
--- a/leo/commands/commanderOutlineCommands.py
+++ b/leo/commands/commanderOutlineCommands.py
@@ -1442,6 +1442,7 @@ def markSubheads(self: Cmdr, event: Event = None) -> None:
         if not p.isMarked():
             if not changed:
                 u.beforeChangeGroup(current, undoType)
+            changed = True
             bunch = u.beforeMark(p, undoType)
             c.setMarked(p)  # Calls a hook.
             p.setDirty()
@@ -1458,8 +1459,6 @@ def unmarkAll(self: Cmdr, event: Event = None) -> None:
     if not current:
         return
     c.endEditing()
-    changed = False
-
     changed = False
     p = None  # To keep pylint happy.
     for p in c.all_unique_positions():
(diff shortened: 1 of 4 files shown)


boltex:
I also made sure to run unit tests.

edreamleo:
@boltex I agree that changes are necessary. Unlike other "before" commands, `u.beforeChangeGroup` pushes a `g.Bunch` on `u.beads`.

edreamleo (approved):
Good work.

Issue #3531, opened 2023-09-02
Add move-outline-to-first/last-child commands


edreamleo:
`move-outline-to-first-child` is convenient when a creating a section.

With default settings Leo creates the section definition node as the *last* child of its parent. If a node has many children it is surprisingly inconvenient to move (manually!) the newly-created node to be the first child of its parent.

`move-outline-to-last-child` is less useful (with default settings), but it should be added as a companion command.

These commands do nothing if a hoist is in effect.

**Aha!** The `extract` commands should create section definition nodes as the *first* child of their parent node.

edreamleo:
Closed via PR #3532.

PR #3532, opened 2023-09-02
PR for #3531: create move-outline-to-first/last-child

edreamleo:
See #3531.

- [x] Create the commands.
- [x] Create new unit tests.
- [x] Extract commands create section definitions as the *first* child.

### DIFF
--- a/leo/unittests/commands/test_outlineCommands.py
+++ b/leo/unittests/commands/test_outlineCommands.py
@@ -605,6 +605,65 @@ class TestOutlineCommands(LeoUnitTest):
         u.undo()
         result_children = [z.h for z in self.root_p.v.children]
         self.assertEqual(result_children, original_children)
+    #@+node:ekr.20230902053728.1: *3* TestOutlineCommands.test_move_outline_to_first_child
+    def test_move_outline_to_first_child(self):
+        # Setup.
+        c, u = self.c, self.c.undoer
+        root = self.root_p
+        assert root.h == 'root'
+        self.create_test_outline()
+        original_children = root.v.children[:]
+        last_child = root.lastChild()
+        assert last_child
+        last_gnx = last_child.gnx
+        # Move.
+        c.selectPosition(last_child)
+        c.moveOutlineToFirstChild()
+        # Tests.
+        assert c.checkOutline() == 0
+        first_child = root.firstChild()
+        assert first_child
+        first_gnx = first_child.gnx
+        assert first_gnx == last_gnx, (first_gnx, last_gnx)
+        u.undo()
+        assert c.checkOutline() == 0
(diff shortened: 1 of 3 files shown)


edreamleo:
@boltex I hope you don't mind this PR. After all these years I realized how useful `move-outline-to-first-child` can be.

` test_move_outline_to_last_child` revealed a bug, so I think the unit tests are strong enough.

boltex:
@edreamleo No problem at all, I'm adding it to leojs easily :)

edreamleo:
@boltex Thanks. I really *do* think I'm done tweaking Leo, for now :-)

boltex:
@edreamleo Do 'assert' and 'self.assert' actually perform the same thing? ( in unit tests, comparing those new tests to the other tests which use the unittest system's "self.assert" instead of normal "assert")

boltex:
Note: I meant also to ask about using "self.assertEqual" which is used in all other tests instead of a normal 'assert'.

leo/unittests/commands/test_outlineCommands.py
```diff
@@ -605,6 +605,65 @@ def test_sort_siblings(self):
...
+        assert first_child
+        first_gnx = first_child.gnx
+        # Move.
+        # c.selectPosition(first_child)
+        c.selectPosition(first_child)
+        c.moveOutlineToLastChild()
+        # Tests.
+        assert c.checkOutline() == 0
+        last_child = root.lastChild()
+        assert first_child
```
boltex, line 655: Perhaps this was supposed to be an assertion of last_child?

leo/unittests/commands/test_outlineCommands.py
```diff
@@ -605,6 +605,65 @@ def test_sort_siblings(self):
         u.undo()
         result_children = [z.h for z in self.root_p.v.children]
         self.assertEqual(result_children, original_children)
+    #@+node:ekr.20230902053728.1: *3* TestOutlineCommands.test_move_outline_to_first_child
+    def test_move_outline_to_first_child(self):
+        # Setup.
+        c, u = self.c, self.c.undoer
+        root = self.root_p
+        assert root.h == 'root'
```
boltex, line 613: Should self.assertEqual be used ?

edreamleo:
@boltex Good question about `assert` vs. `self.assertEqual`, etc.

The two are not equivalent, but the choice between the two typically matters only when either assert fails :-)

At first I thought perhaps unit tests should always use one or the other. Now I just use whatever is most convenient.

`self.assertEqual` can provide more helpful messages when it fails. Also, iirc, `self.assertX` do *not* report their failures when run with pytest: stdout gets sent to dev/null.

Issue #3551, opened 2023-09-10
Saving .db files once already created on disk and modified asks for reload, as if dirtied externally..


boltex:
Hitting ctrl+s on a .db file which was already saved, and modified, makes Leo detect that own file change, erroneously asking if the user wants to reload the .db file - not recognizing that it was itself that just changed it by saving on disk.

The logic for .db files should be the same as .leo files regarding the self-change detection behavior.

edreamleo:
@boltex This issue remains even in the branch that supposedly fixes #3550. I'll track it down.

edreamleo:
Closed via PR #3554.

PR #3553, opened 2023-09-10
PR for #3551 db checks

edreamleo:
See #3551.

**Closed in favor of PR #3554.**

- [x] `efc.has_changed` always returns `False` for `.db` files. This *is* dubious: it disables all checks for `.db` files.
    I'll create a new PR that attempts a proper fix.
- [x] Remove several special cases: `efc.has_changed` handles the special case for `.db` files.
- [x] Remove tracing code in `at.shouldPromptForDangerousWrite`.

### DIFF
--- a/leo/core/leoAtFile.py
+++ b/leo/core/leoAtFile.py
@@ -2915,50 +2915,33 @@ class AtFile:
 
         See #50: https://github.com/leo-editor/leo-editor/issues/50
         """
-        trace = 'save' in g.app.debug
-        sfn = g.shortFileName(fn)
-        c = self.c
         efc = g.app.externalFilesController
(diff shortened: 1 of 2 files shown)


edreamleo:
@boltex I am unwilling to make complex changes to the code. I think this PR is the best that can be done for now.

Please let me know if you disagree.

edreamleo:
@boltex On second thought, I'll create a new PR that attempts a better fix.

Imo, the new PR must also remove `c.sqlite_connection`. This status var is evil. Instead, Leo should keep the connection open only while reading and writing the db. But this cleanup may be too much for Leo 6.7.5. If it is we can live with *this* PR.

edreamleo:
Closed in favor of PR #3554.

PR #3554, opened 2023-09-10
PR 2 for #3551

edreamleo:
**This PR will remain a draft until Leo 6.7.6**.

See #3551. This PR replaces PR #3553.

<details><summary><b>Detailed summary of changes</b></summary>
<br>

The following list is based on Leo's `git-diff-pr` command:

**Delete ivars**

- [x] Delete `g.app.openingSettingsFile` and `c.sqlite_connection` ivars.

**Delete methods**

- [x] Delete `lm.openAnyLeoFile`, `lm.openLeoFile`, and `lm.readOpenedLeoFile`.
   None of these methods belong in the `LoadManager` class.
- [x] Delete `fc.getLeoFile` and `fc.openLeoFile`. 
    These methods contained a `theFile` arg that was an open file: it was a union of a file-like object and an sqlite connection!
    In effect the `theFile` arg was a hidden kwarg.
    `fc.getLeoFile` *closed* its `theFile` arg *unless* the arg was a sqlite connection!!

**Add methods**

- [x] `fc.getAnyLeoFileByName` replaces both `fc.getLeoFile` and `fc.openLeoFile`.
    `fc.getAnyLeoFileByName` uses two new private helpers: `fc._getLeoDBFileByName` and `fc._getLeoFileByName`.

*Note*: New methods take `path` args instead of `fn` or `fileName` args; signatures of all legacy methods remain unchanged.

**Change methods**

- [x] Numerous methods now call the added methods instead of the deleted methods.
- [x] Refactor `lm.openFileByName` using the `complete_inits` function.
- [x] Change several `finally` clauses. Add comments that `finally` clauses should not contain `return` statements.
- [x] Remove all logic involving  `c.sqlite_connection`.
   `fc.exportToSqlite` closes sqlite connections immediately after using them.
- [x] Remove all tests on the file extensions. Only  `fc.getAnyLeoFileByName` does this test.

</details>


boltex:
_note: I'm really a newbie when it comes to sql._ 

I wish i knew why the .db was treated so differently from .leo files when it comes to opening/reading, and saving/writing.

Vitalije had honorable intentions, thanks to him for this feature, and indeed .db file support "works"... **but it could be streamlined and made to work a lot more like the other .leo/.leojs formats**. 

e.g.: no need to keep the connection to the db opened once read to get the contents when a .db file is opened. (just read the tables to recreate the vnodes and close it) Because, when saving with exportToSqlite, it just open the file/connection from scratch and commit/write to the file and close the connection+file. Bam, as simple as that.

 Also, this write operation should be viewed by Leo just like the writting of .leo or .leojs files and obviously not trigger the 'Document has changed externally!' event, which is the original reason this issue was filed.

Could the .db support be totally simplified like i just described? or am I missing something because of my lack of sql experience?

edreamleo:
@boltex I agree completely with your comments.

I plan to eliminate the following anti-patterns:

- There is no need to hold the the sqlite connection open, and there is no need for the `c.sqlite_connection` ivar.
- The `theFile` arg is annotated as `Any`, but in fact is a union of a normal file-like object and a sqlite connection.
- Passing open files as arguments is poor, especially when the file is a union.

*The union args are a form of hidden (evil) kwargs*.

My strategy will be create new, more specialized methods that take *known* (non-union) arguments. Best would be to only pass filenames. This strategy may create code duplication, but that's often true when eliminating kwargs. We'll see how this works :-)

boltex:
Good! :)

In leojs, i already have made a slight bifurcation from how it was done in Leo, in that I use the filename only, because the concept of an 'opened file' does not exist in a vscode extensions. So I cant 'pass that around' as a parameter... I have to read/write 'on the spot' and be done with it right then and there. **But it's simpler that way.**

And obviously, as you have also realized, it could be the same with .db files. simplifying and unifying everything concerning all the "file read/write" related stuff in Leo that happens in the "new, open, save, save-as, etc..." commands in general.

edreamleo:
@boltex I agree that the code will *eventually* be simpler, but the changes required are frightening.

I now have code that passes all unit tests, but I'm nervous. Time for a break ;-)

edreamleo:
@boltex Ready for review:
- The new code is much better than the old.
- The original issue seems to have been fixed.
- All tests pass.

Having said that, this PR must be tested thoroughly tested. I look forward to your review. There's no rush.

boltex:
@edreamleo I'm almost done with the importers/writers, so I'll review and test this on the weekend. I'll also translate it into leojs as  soon as i do so :)

edreamleo:
@boltex Thanks for your comment :-)

edreamleo:
@boltex I always welcome your comments, but I think we should definitely leave this PR for later.

boltex:
@edreamleo Yeah I also agree. let's leave this fix for after Leo 6.7.5 and leojs first beta release.

PR #3597, opened 2023-10-08
PR for filename-related quirps in various save commands

edreamleo:
The existing logic is bewildering, but probably not buggy.

Attempting to improve the logic is fraught with danger.

### DIFF
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -357,7 +357,8 @@ def save(self: Self, event: Event = None, fileName: str = None) -> None:
     """
     c = self
     p = c.p
-    # Do this now: w may go away.
+
+    # Save w now: w may go away.
     w = g.app.gui.get_focus(c)
     inBody = g.app.gui.widget_name(w).startswith('body')
     if inBody:
@@ -365,16 +366,20 @@ def save(self: Self, event: Event = None, fileName: str = None) -> None:
     if g.app.disableSave:
         g.es("save commands disabled", color="purple")
         return
+
     c.init_error_dialogs()
+
     # 2013/09/28: use the fileName keyword argument if given.
     # This supports the leoBridge.
     # Make sure we never pass None to the ctor.
+
     if fileName:
         c.frame.title = g.computeWindowTitle(fileName)
         c.mFileName = fileName
     if not c.mFileName:
         c.frame.title = ""
         c.mFileName = ""
+
     if c.mFileName:
         # Calls c.clearChanged() if no error.
         g.app.syntax_error_files = []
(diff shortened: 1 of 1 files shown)


tbpassin (commented):
1. Tiny typo at Line 29 of commanderFileCommands.py:
`Raise error dialogs and restore the focut.`  Should read "focus".

2. In same file, can _save_focus_data()_ really work?  Because to invoke the command, the user will be in a menu or invoking a command from the Minibuffer, so one of those would have the focus.

3. In same file at line 62, the line in the docstring `Compute c.mFileName and related ivars.` should be the first line of the docstring, if we are following PEP8.

4. In same file, line 427, do we really want to update RecentFiles if the save command failed?
```
        c.fileCommands.save(fileName)
        g.app.recentFilesManager.updateRecentFiles(fileName)
```
5. In same file, line 435ff:
```
    # Don't prompt if the file name is known.
    given_file_name = fileName or c.mFileName
```
_fileName_ would have to be _None_ or an empty string for _c.mFileName_ to win.  Should the _fileName_ argument have an empty string default?

6. Same file, line 509: ` new_file_name = set_name_and_title(c, fileName)`
In comment 5, _fileName_ might be empty.  Don't we need to handle that possibility here?

7

edreamleo:
@tbpassin Thanks for your comments!

1. I'll fix the typo.
2. Good question.  `save_focus_data` will work sometimes, and maybe more often than we think :-)
  Anyway, the new code is equivalent to the old in this respect, so the PR hasn't *added* a bug.
3. Almost none of Leo's docstrings follow PEP8 in this respect because headlines show context better than the docstring.
4. Another good question! The code calls `do_save` only when *trying* save the file.
  But perhaps `fc.save` will raise an error dialog instead.
  As in 2. above, the new code is equivalent to the old in this respect, so the PR hasn't *added* a bug.
5. Initing the `fileName` is proper Python style. The code is as intended.
6. `set_name_and_title` always returns the "finalized" `c.mFileName`, regardless of the `fileName` kwarg.
  This is a subtle point, but I believe the code is correct in this respect.
  Note: the legacy code did not always finalize `c.mFileName`, so the new code is clearly better, imo.

Thanks for your careful reading of the changes.

edreamleo:
@tbpassin I misunderstood your comment about the docstring. I'll do as you suggest.

edreamleo:
@tbpassin @boltex I have discovered a better way to handle focus issues, namely in the Qt dialog code. This changes further simplifies the three save commands.

Please review the code again.  Thanks.

edreamleo:
@boltex @tbpassin I think the code is complete. The last Aha: use try/except to indicate functions that must be paired.  Doh!

Unless some new ideas crop up I plan to merge this PR tomorrow, Wednesday.

I've put a lot of effort into the first comment. I think everything has settled down.

tbpassin (commented):
In file _commanderFileCommands.py_, around line 585, `finish_save_command()` will be called even if the file dialog does not return a file string.  Pleases check to make sure that _finish_save_command()_  can handle this case.

I'm not happy with the organization of this part anyway.   I'd rather see something like this:
```
save_filename = fileName or g.app.gui.runSaveFileDialog(...)
if save_filename:
    do_save_to(c, save_filename)
    finish_save_command(c, p, inBody)
else:
    # handle case of missing file name
c.bringToFront()
```

Otherwise, it was hard to see additional changes in the code from my original review, so the above are my only comments.

Issue #3604, opened 2023-10-12
Improve warning when Qt is not installed


edreamleo:
The present message is:

```console
'NoneType' object has no attribute 'gui'

** Leo could not be started **
```

This message should be improved, if possible.

edreamleo:
Closed via PR #3605.

PR #3605, opened 2023-10-12
PR for #3604: startup message

edreamleo:
None of these changes are likely to affect leoJS.

- [x] Improve messages issued when the requested gui (usually Qt) was not imported.
- [x] Fix bug that prevented the emergency (Tk) dialog from being shown.

**Bonus**

Remove dependencies on Qt in various places:

- [x] The `NullGui` class no longer depends on Qt!
- [x] The `nodetags` plugins no longer depends on Qt.  Previously, it had said it didn't, but it did :-)
- [x] Improve unit testing framework. All tests pass (skipping Qt-related tests) when Qt is not available.

### DIFF
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -993,15 +993,10 @@ class LeoApp:
         app = self
         try:
             from leo.core.leoQt import Qt
+            from leo.plugins import qt_gui
             assert Qt
         except Exception:
-            return None
-        try:
-            from leo.plugins import qt_gui
-        except Exception:
-            g.es_exception()
-            print('can not import leo.plugins.qt_gui')
-            sys.exit(1)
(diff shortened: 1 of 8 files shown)


tbpassin (commented):
leoApp.py, Line 973: `https://leo-editor.github.io/leo-editor/installing.html`

I'm concerned because it will be very easy for this reference to the needed dependencies to get out of date.  It may be out of date even now.  However, I don't have a solution.  It's even harder because there isn't a single, easy to find and read list of dependencies anymore.

I wonder if during startup we could have Leo collect each dependency that failed to load.  Then that list could be emitted during the error message.

Otherwise the new message is certainly an improvement.

edreamleo:
@tbpassin Thanks for your review. The link is valid now, and I doubt that it will ever change.

Issue #3606, opened 2023-10-12
Add file extensions for C++ and Scheme


edreamleo:
See PR #3607.

LilyPond uses `.hh` and `.cc` for C++. Leo already associates `.scm` with scheme.

Adapt the `elisp` importer for scheme.

edreamleo:
Closed via PR #3607.

PR #3607, opened 2023-10-12
PR for #3606: scheme importer

edreamleo:
See #3606.

- [x] Associate `.hh` and `.cc` file extensions with C++.
- [x] Add Scheme importer and associated unit test.
- [x] Add Scheme to a weird list of exceptions in `jedit.match_span` so that Scheme strings are colorized properly.
- [x] Fix bug in `Elisp_Importer`: Only double quotes start strings!


tbpassin (approved):
This was not a deep review but I don't see anything to comment on.

edreamleo:

Issue #3546, opened 2023-09-07
Allow different kinds of imported files on the command line


tbpassin:
See PR #3645:

- Allow both  `@edit` and `@file` nodes to be imported from the same command line.
- Retire the `--load-type` command-line option.

<details><summary>Original Report</summary>
<br>

Leo used to import a file on the command line into a new outline (if the file were not a .leo file).  Currently nothing gets imported, although a new node is created with the headline `none <filename>` where `<filename>` is the name typed on the command line.

The help message says 

`--load-type=TYPE      @<file> type for non-outlines: @edit or @file`

but that doesn't work either.  It just replaces "none" by "file" or "edit" (without the "@" character) and does not actually import anything.  Using `@@edit` on the command line just produces an error.

</details>

tbpassin:
The reason for this is that the value of `--load-type` does not store the leading `@` but the using code in the LoadManager expects it to be there.  The LM code in question is

`load_type = self.options.get('load_type', '@edit')  # #2489.`

This could be fixed when the command line option is stored or in the LM code.  For example:
```
load_type = self.options.get('load_type', '@edit')  # #2489.
if not load_type.startswith('@'):
    load_type = '@' + load_type
```
This would solve the problem that `--load-type` doesn't work.  But it would not provide for importing a non-Leo file on the command line.  Perhaps that capability should be considered to be an enhancement and not part of this bug.

tbpassin:
In testing the fix, it works **but** `@file` loadtype fails with the message 
```
read_into_root Invalid external file: <file>
```
where `<file>` is the actual file path.  

Here is an example of a failing command line (_py-leo-git_ is my batch file that launches Leo using my git clone) (paths_ are specific to my computer, of course):

`C:\Tom\git\gf4-project>py-leo-git --load-type=@file C:\temp\python\pyqt-spreadsheet.py`

This happened for both of my test files, a .js file and a .py file.  The same command line but with `@edit` successfully imported those files into @edit nodes.

Should I submit this PR as is and then we can work on the bug in a different PR, or how should we handle this?

edreamleo:
@tbpassin Please submit a PR. Thanks.

edreamleo:
Fixed via PR #3645.

PR #3614, opened 2023-10-15
Replace lost "@" character when command line option "--load-type" is processed

tbpassin:
This PR fixes issue #3546. But the option `@file` fails with an error message:

`read_into_root Invalid external file:`

This is an unrelated bug.  The `@edit`  option acts as expected.


edreamleo (approved):
@tbpassin Thanks for this PR.

PR #3618, opened 2023-10-23
PR: Fix quirp in check_outline

edreamleo:
@boltex I'm happy to merge this PR into devel now, or after leoJS debuts. It's your call.

mypy was always good with the "overloaded" name `actual`. Surely there are other overloaded variable names scattered throughout Leo. I assume Typescript is more strict than mypy in this regard.

- [x] Split the `actual` var in `check_outline` into `actual_list` and `actual_tuple` vars.
- [x] Fix several valid pylint complaints by using the `coffeescript` name throughout `test_importers.py`.

### DIFF
--- a/leo/unittests/test_importers.py
+++ b/leo/unittests/test_importers.py
@@ -41,10 +41,10 @@ class BaseTestImporter(LeoUnitTest):
         """
         try:
             p0_level = p.level()
-            actual = [(z.level(), z.h, z.b) for z in p.self_and_subtree()]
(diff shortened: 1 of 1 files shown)


edreamleo:
@boltex Do you want to merge this PR? It should not affect the unit tests for the importers!

edreamleo:
@boltex Last call for comments on this PR. If I don't hear from you I'll merge this into devel tomorrow (Monday).

PR #3623, opened 2023-10-29
Fix quirps re @asis and @nosent

edreamleo:
- [x] Fix a typo in the documentation for `@asis`.
- [x] Simplify the dialog raised when Leo detects changes to `@asis` and `@nosent` nodes.

**Félix**: Original report from private email (slightly edited):

QQQ
When external file changes are detected for `@nosent` and `@asis`, a dialog message appears. The message says those nodes can't be read, and that *using the refresh from disk will clear the node*.

When I manually tried `refresh-from-disk` on those nodes, Leo just prints a message in the log warning the user, *without* changing the node.

The dialog message should be amended to say `@nosent` / `@asis` do not read from file.
QQQ

The original report mentioned `@edit` rather than `@asis`.

**First, do no harm**

In fact, the user *can* recover the external changes as follows:

- Change `@asis` or `@nosent` to `@edit` *without* saving the file.
- Execute refresh-from-disk or restart Leo *without* saving the outline.

However, this procedure will overwrite changes made externally if the user *does* save the outline.

Imo, it's best not to mention this recovery method in the dialog! Long error dialogs can cause panic, and panic might lead to data loss.

### DIFF
--- a/leo/core/leoExternalFiles.py
+++ b/leo/core/leoExternalFiles.py
@@ -641,7 +641,7 @@ class ExternalFilesController:
             c=c,
             message=(
                 f"{path_name} has changed outside Leo.\n\n"
-                f"An {kind} file created this file.\n\n"
+                f"An {kind} node created this file.\n\n"
                 'Warning: Leo can not update this file!\n'
                 'Neither refresh-from-disk nor restarting Leo will work.\n'
             ),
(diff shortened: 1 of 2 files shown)


edreamleo:
@boltex Many thanks for your close reading of Leo's code and documentation.

PR #3644, opened 2023-11-09
PR fixes #3596: Fix two quirps in revert command

### LINKED ISSUE #3596: Revert to saved does not refresh the buttons in the top button bar
To reproduce: Making a node containing a script,  `g.es('hello')` and while having that node as the current selection, Adding a new button with the **script-button** button, which adds a button with the currently selected node.

 While _not saving the file_ and using the **'revert' command** -> the button bar does not refresh the buttons.

(the button is still there, but it points to inexistent code and gives an error when pressed.)

edreamleo:
See #3596.

- `LM.revertCommander` safely calls `c.theScriptingController.createAllButtons`.
- `sc.createAllButtons` calls `sc.deleteAllButtons` when `g.app.reverting` is True.

These changes fix the UI problem.

**Another quirp**

During a `revert`, `sc.createAllButtons` re-registers the *commands* created by `@button` nodes.

There is no reasonable way to *unregister* the commands that the *old* (potentially zombie) buttons created!

However, executing the zombie *commands* is safe because because the `AtButtonCallback` (ABC) class is safe:

- `<Alt-X>zombie<Return>` eventually(!) calls `ABC.execute_script`.
- `ABC.execute_script` calls `ABC.find_script`, which returns None because the node with the zombie gnx no longer exists!
- `ABC.execute_script` returns *without calling* `ScriptingController.executeScriptFromButton`.

In short, the *zombie script does nothing*, safely.

**Extras**

- [x] Improve the warning about non-existent file.
- [x] Suppress the warning about an already-open file.
- [x] Report successful `revert` operations to the console and the log.
- [x] Improve traces for `--trace=keys`.

### DIFF
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -553,7 +553,7 @@ def revert(self: Self, event: Event = None) -> None:
         g.es('Can not revert unnamed file.')
         return
     if not g.os_path_exists(fn):
-        g.es(f"Can not revert unsaved file: {fn}")
+        g.es(f"Can not revert non-existent file: {fn}")
         return
     reply = g.app.gui.runAskYesNoDialog(
         c, 'Revert', f"Revert to previous version of {fn}?")
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -3255,7 +3255,13 @@ class LoadManager:
 
         # Re-read the file.
         c.fileCommands.initIvars()
-        v = fc.getAnyLeoFileByName(fn, readAtFileNodesFlag=True)
+        try:
+            g.app.reverting = True
+            v = fc.getAnyLeoFileByName(fn, readAtFileNodesFlag=True)
+        finally:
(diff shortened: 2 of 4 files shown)


boltex:
I'll try again and report about it soon

edreamleo:
@boltex Thanks. Am I correct in assuming that this issue affects LeoJS 0.1?

edreamleo:
@boltex I've fixed the original issue.

Leo doesn't have the machinery to unregister commands.

edreamleo:
@boltex Please review and approve this issue.
 
The first comment here discusses why the remaining quirp is harmless.

boltex (approved):
I've also tried it, and although the new buttons end up after the prev, next and chapters buttons, this is still much better than the previous state of things where the buttons remained after a 'revert' so I approve of this pull request.

edreamleo:
@boltex Thanks for your review. Any remaining anomalies are too small to care about.

PR #3645, opened 2023-11-09
PR fixes #3545: Allow Leo to import any kind of file on the command line

### LINKED ISSUE #3545: Make set-ua, clear-ua, clear-all-ua commands undoable.
See PR #3732.

edreamleo:
See #3546. Allow Leo to import any number of **external** (neither `.leo`, `.db,` nor `.leojs`) files from the command line.

This PR refactors and simplifies the code the reads files. This PR should help simplify leoJS.

All plugins and scripts (including Leo's core) should call `g.openWithFileName` instead of calling methods of the `LoadManager` class.

This refactoring may have *slight* impacts on existing scripts and plugins:

<details><summary><b>Changes to Leo's API</b></summary>
<br>

This PR deletes several *publicly visible* helpers of the `LoadManager` class.
There is a slight possibility that existing scripts or plugins may be using those deleted helpers.

In all such cases, the fix is to use `g.openWithFileName` instead.
Leo's core now uses `g.openWithFileName` to open files of *any* type.

**Changes to the `LoadManager` class**

- Delete `lm.openFileByName`, an internal helper for `lm.loadLocalFile`:
- Replace `lm.initWrapperLeoFile` with `lm.openExternalFile`:
- Add the `fn` arg to the the signature of `lm.openEmptyLeoFile`:

**Other changes**

- Delete `c.createNodeFromExternalFile`, another helper for `lm.openFileByName`.
- Delete a strange Easter Egg involving the `k.givenArgs` ivar.
  `leoVim.py` referenced `k.givenArgs`, but only in *disabled* code.
- Delete the unused `c.wrappedFileName` ivar.
- Delete the unused `efc.openOutlineByName` method.

</details>

<details><summary><b>Tasks</b></summary>
<br>

- [x] Retire the `--load-type` command-line option, adding it to the list of obsolete options.
- [x] Improve error message in `load_sessions`.
- [x] Fix a pre-existing bug in `LM.initWrapperLeoFile`: set `c.p` after creating an `@file node`.
- [x] Correct the annotation of arguments to `LM.loadLocalFile`.
- [x] Rename `open-outline` to `open-file` (in the menu), retraining the old name for compatibility.
- [x] Remove Easter Egg involving `c.k.givenArgs` in the `open-file` logic.
- [x] Make `g.openWithFileName` the common helper for all open-related logic.
- [x] Set the name of wrapper `.leo` files to `untitled` so the wrapper files are not written unless the user agrees.
- [x] Refactor `lm.openWithFileName` and its helpers.
- [x] Use `lm.finishOpen` in all helpers of  `lm.openWithFileName`.

</details>


edreamleo:
@tbpassin Imo `--load-type` is an evil command-line option.

`git blame` says I am responsible, though I have no memory of it.

Imo, loading a non-outline from the command line should always create an `@edit` node. Nothing else seems like would make sense.

What do you think?

tbpassin:
I think putting a file on the command line with a load option should act  like the "Import Any File" menu option (except for not needing a file dialog).  I was surprised when I learned that it didn't.  Of course, it should open a Leo outline as an outline instead of importing it, but otherwise it should act just like the menu item.

I think this behavior follows the Principle of Least Surprise, and I think it's what anyone would want.  For example, if I want to work on a javascript file, or a rust file, I would want it to be imported, not sucked into an @edit node when there is a perfectly good importer available.

So let's change the option name to "import-file", which will make the intent more clear.  For backward compatibility keep the current name too, and have it behave the same as import-file. Some requirements:

1. If the file argument is a Leo outline, open that outline; otherwise
2. If the file is already an external file in any open Leo file, navigate to that outline/file once Leo finishes opening; otherwise
3. If there is an importer for that kind of file, open a new outline and import the file into it; otherwise
4. Open a new outline and suck the file into an `@edit` node.
5. Only allow one import-file argument on the command line;
6. Process the import-file arg after all the other command line outlines have been opened (because of 2. above).

edreamleo:
@tbpassin "[importing from the command line] should act like [import-file]"

Super idea! I also had the idea of using the logic of `c.looksLikeDerivedFile`. Using the logic of `c.importAnyFile` (w/o needing a dialog) is much better: it's a more general solution.

*Note*: Now we can see another problem with the `--load-type` command-line option. The option declares that *all* files imported from the command line shall be imported the same way (all `@edit` or `@file`). Instead, using `c.looksLikeDerivedFile` allows some files to generate `@file` nodes while others generate `@edit` nodes.

edreamleo:
@tbpassin Imo, Leo should be able to import multiple non-outline files from the command line.

I'm dubious about steps 2,  5 and 6 above. Putting, say, `x.py`, and `y.csv` should create the appropriate wrapper outlines for both files, regardless of whatever outlines appear on the command line. It's up to the user to avoid data race conditions. In practice, this is unlikely ever to be a problem.

In short, there is no need to complicate `LM.doPostPluginsInit` by deferring imports.

edreamleo:
@tbpassin This PR passes various hand tests. From the `leo-editor` directory the following works:

`leo CONTRIBUTING.md leo/core/leoAst.py leo/core/LeoPyRef.py`

As mentioned in the first comment of this issue, there may be a quirp or two with restoring a session involving non-outline files, but I haven't been able to reproduce the quirps. In any case, the obvious cases appear to work.

tbpassin:
> Imo, Leo should be able to import multiple non-outline files from the command line.

OK by me if the syntax is simple.  Would that be a comma-separated list? E.g. `--import-file=file1.js,file2.py,file3.cs`?

tbpassin:
> _Note_: Now we can see another problem with the `--load-type` command-line option. The option declares that _all_ files imported from the command line shall be imported the same way (all `@edit` or `@file`).

Yes, the old way always bothered me, especially since the _Import Any File_ command already existed.

tbpassin:
> > Imo, Leo should be able to import multiple non-outline files from the command line.
> 
> OK by me if the syntax is simple. Would that be a comma-separated list? E.g. `--import-file=file1.js,file2.py,file3.cs`?

Oh, I see.  No need for the _import-files_ param at all.  Even better!

tbpassin (approved):
I didn't see anything to remark on.  Of course it depends on what `loadLocalFile()` does, which is apparently hardly changed for the PR, and I'm going to assume it will import non-Leo files according to their file type.

edreamleo:
@tbpassin Thanks for your comments. This PR is still a work in progress. I pretty much know what I want, but it's not clear how much new/revised code there will be.

edreamleo:
@tbpassin Thomas, please review the summary of API changes in the first comment of this PR. Does it look good to you?

tbpassin:
1. `A generalization: g.openWithFileName now uses g.app.gui if the gui kwarg is None`
Will the method still work with a null gui (i.e., the LeoBridge)?

I didn't notice anything else.  I can't speak to the completeness of the listed changes, of course.  Thanks for addressing backwards compatibility!

edreamleo:
@tbpassin Good catch. Leo's bridge would have to instantiate a NullGui. I'll remove the generalization.

edreamleo:
@tbpassin I'm going to back out of all non-essential changes.

I can defend the refactoring of `lm.openWithFileName` on the grounds that the previous code was virtually impossible to understand or to fix. But there is no reason for anything "extra".

boltex (commented):
Wow ! what a cleanup! :) I'l make sure to bring all those changes to leojs after the alpha launch in one go with all the changes since 6.7.5.

edreamleo:
@boltex Thanks for your review. The cleanup was long overdue. Glad you approve.

PR #3650, opened 2023-11-14
PR: Change http to https in the header of .leo files

edreamleo:


tbpassin:
Do not make this change.  It has no point because the actual value of the namespace is a formality.  It will never be dereferenced in Leo's case.  But if anyone has been processing leo files with proper XML processors, this change will likely break their processing.

In sum: this change is pointless and has the potential to break processing by others.

boltex:
This was changed a few weeks ago when we changed the main site from leoeditor.com to leo-editor.github.io. It was not changed properly and this small fix is just the proper correction to be made to the XML header. this is relatively recent, It's not like it was in place for decades and that many big projects depend on it being immutable.
Respectfully, I think people who process XML automatically who have big problems if this entry of the xml header change, for which has no impact because its a formality in the xml file, should take this opportunity to adapt their parsers to ignore that part of the XML file header.

tbpassin:
The point is that namespace URLs (which don't actually have to be URLs) should never change.  It's true that Leo doesn't use the value, but since the file is an XML file it should be used as XML intends, which is that namespaces are persistent.

edreamleo:
@tbpassin I am going to merge this PR for the following reasons:

- I see no reason not to revert Leo's header to the status quo ante.
- Neither Leo nor any other processor is likely to object to the change.

PR #3663, opened 2023-11-23
PR fixes #3660: fix window title computations

### LINKED ISSUE #3660: 'New' document tabs & window-title are sometimes created with the 'branch name' prefix before the title.
See PR #3663.

As a context to explain this bug, shown in the this first screenshot, 'small2.leo' does not show the branch name in the tab, only in the 'windows' title bar, as is expected...

But, strangely, sometimes creating a new document with ctrl+N adds a branch info prefix for the untitled new file, (same screenshot below, see the rightmost tab) which should not be assigned to any folder/branch or anything really, as it is a new, unsaved, document. (Note: when this happens, it also has it in the  'windows' title bar instead of just 'untitled'.)

[small2.leo](https://github.com/leo-editor/leo-editor/assets/847716/4b55e888-c95a-43d8-a4de-7fb7572e394e)

After doing a few experiments, after opening leo who also opens the last session files normally, creating new files is ok, but after I manually open another leo file, then after that point the new files are created with the branch-name prefix in both tab and windows title bar.

 (see below where they are arranged from left to right to show the behavior more clearly: From left to right: I start with  'myLeoFile.leo' already opened, and I create 2 new files, so far so good, then I open 'small2.leo', oh-oh, subsequent new files are created with the branch prefix anomaly in the tab and windows title bar. )

[myLeoFile.leo](https://github.com/leo-editor/leo-editor/assets/847716/298e2a9c-e36e-4dc4-a162-69f48c99a62f)

edreamleo:
See #3660.  This PR may affect LeoJS.

- [x] Retire `g.computeWindowTitle`. Use `c.computeWindowTitle` everywhere.
- [x] Simplify `c.computeWindowTitle`.
- [x] Update `g.app.numberOfUntitledWindows` *only* in `c_file.new`.
- [x] Add `c.computeTabTitle`. It never adds a branch name.
- [x] Only `c.init_objects` calls  `c.computeTabTitle`.
- [x] Only `slotCurrentChanged` calls `c.computeWindowTitle`.
- [x] Revise `c.computeWindowTitle` as Félix suggests.  See rev e52296d.

### DIFF
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -256,6 +256,7 @@ def new(self: Self, event: Event = None, gui: LeoGui = None) -> Cmdr:
             shortcutsDict=lm.globalBindingsDict,
         ))
     t3 = time.process_time()
+    g.app.numberOfUntitledWindows += 1
     frame = c.frame
     g.app.unlockLog()
     if not old_c:
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -248,6 +248,7 @@ def new(self: Self, event: Event = None, gui: LeoGui = None) -> Cmdr:
     g.app.lockLog()
     # Retain all previous settings. Very important for theme code.
     t2 = time.process_time()
+    g.app.numberOfUntitledWindows += 1
     c = g.app.newCommander(
         fileName=None,
         gui=gui,
@@ -256,7 +257,6 @@ def new(self: Self, event: Event = None, gui: LeoGui = None) -> Cmdr:
             shortcutsDict=lm.globalBindingsDict,
         ))
     t3 = time.process_time()
-    g.app.numberOfUntitledWindows += 1
     frame = c.frame
     g.app.unlockLog()
     if not old_c:
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -46,7 +46,7 @@ def set_name_and_title(c: Cmdr, fileName: str) -> str:
         c.mFileName = g.ensure_extension(fileName, g.defaultLeoFileExtension(c))
 
     # Set various ivars.
-    title = c.computeWindowTitle(c.mFileName)
+    title = c.computeWindowTitle()
     c.frame.title = title
     c.frame.setTitle(title)
     try:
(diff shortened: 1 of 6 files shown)


tbpassin (commented):
1. in computeTabTitle(), Lines 169 -
```        if c.fileName():
            return c.fileName()
```
Suggested -
```
        fn = c.fileName()
        if fn:
            return fn

Otherwise looks fine.

edreamleo:
@tbpassin Hi Thomas. I considered the code you suggest. Not sure why I didn't do it :-)

edreamleo:
@tbpassin Rev d6d0cc3 does as you suggest.

boltex (changes requested):
Oops, this PR does not work for me... there's a bug with the 'branch' name and the file title is not presented before the path, helping seeing clearly the filename itself (could be after a long path, which is hard to distinguish) here's what I found while testing it: 

(I had a strange feeling when I saw that leoGlobal's computeWindowTitle was completely gone, along with it's logic to  'pretty-print' the title nicely with the filename presented first, followed by **'in folder/path/x/y/z**')

Check this screenshot: (before it showed the branch **of the file**, not **of Leo itself**, and it had the title of the file **first**, followed by '**in folder/path/x/y/z**'
example_title

boltex:
Adding **fileName or c.fileName()** as parameters to the call to gitBranchName solves the first bug (wrong branch name) I mentionned above.
image

edreamleo:
@boltex Thanks for your review! I'll change `c.computeWindowTitle` as follows:

- Add the pretty-print logic from `g.computeWindowTitle`.
- Change the call to `g.gitBranchName` as you suggest.

edreamleo:
@boltex I've made the changes you requested. See rev e52296d.

edreamleo:
@boltex Rev f1ea634 contains all the changes that I think are needed:

- The change to `c.computeTabTitle` comes from another branch. It works for n in (0, 1).
- The change to `c.computeWindowTitle` returns `untitled` as you suggest.

If (as you show) something is calling `c.computeWindowTitle` too often, the solution is to find the culprit rather than hacking `c.computeWindowTitle`!

edreamleo:
@boltex Rev 1e7a5e8 contains the regex you suggest. It's a good change. Thanks.

I'm off to bed. If problems remain I would be willing to zoom about them tomorrow.

boltex (approved):
I've tested it thoroughly, The last commit changes do fix the issue. 👍

edreamleo:
@boltex Thanks again for all your careful work on this PR.

PR #3669, opened 2023-11-26
PR fixes #3668: don't call c.computeWindowTitle in createRecentFilesMenuItems

### LINKED ISSUE #3668: computeWindowTitle seems to be called 50+ times repeatedly when creating/opening documents.
Here's a quick test i did:
I just added this to to computeWindowTitle. 
```
print('hi')
g.trace(g.callers())
```
image

I'm getting the same output repeated 50+ times below when opening / creating new documents. etc.

image

I'll do some more tests and investigation on this, and I'll report by adding comments to this issue. ... or maybe a PR if I find a good solution! :)

boltex:
Fixes #3668.

`createRecentFilesMenuItems` should not call `c.computeWindowTitle` repeatedly. That would be a performance bug.

### DIFF
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -3416,7 +3416,10 @@ class RecentFilesManager:
                     c.add_command(recentFilesMenu, label=baseName,
                         command=recentFilesCallback, underline=0)
             else:  # original behavior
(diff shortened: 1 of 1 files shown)


edreamleo (approved):
Looks good to me.  I have simplified the title of this PR and fixed a merge conflict.

PR #3672, opened 2023-11-27
PR fixes #3661: Update docstrings for c.fullPath and g.fullPath

### LINKED ISSUE #3661: c.fullPath() does not return what the docstring says
The c.fullPath() docstring says in part

`Return the full path (including fileName) in effect at p`

In fact, it returns the path **not** including the fileName when there is no \@file directive (or, I assume, other similar directives) in effect.  When there is such a directive, the full path including file name is returned.

g.fullPath()  acts the same.  

This method should consistently return the full path including the file name.

edreamleo:
See #3661.

This PR does *not* changes the actual code. Such a changes would break existing code.

### DIFF
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -2367,8 +2367,12 @@ class Commands:
     #@+node:ekr.20230402232100.1: *4* c.fullPath
     def fullPath(self, p: Position) -> str:
         """
-        Return the full path (including fileName) in effect at p. Neither the
-        path nor the fileName will be created if it does not exist.
+        Return the full path in effect at p.
+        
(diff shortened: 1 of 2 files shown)


edreamleo:
@tbpassin Please approve this PR. I am not going to change the code itself.

tbpassin (commented):
leoCommands.py, line 2372:
`If p is an @<file> node, return the path will including the filename.`
Edit to read:
`If p is an @<file> node, return the path will include the filename.`

tbpassin (commented):
I hit "Submit" too soon by mistake. Make the same edit in

edreamleo:
@tbpassin I chose a different way to fix the typos :-)

Issue #3671, opened 2023-11-27
unsaved untitled files should not make an entry in sessions. 


boltex:
See PR #3675.

When saving session, to be reopened at the next Leo startup, an UNL is saved for each opened window frame that was opened upon closing it... 

But,  unsaved untitled files cannot be reopened and should not be used to create an invalid UNL in the sessions data, which gives those lines in the terminal upon reopening Leo after closing with with any number of unsaved untitled file:

[image](https://github.com/leo-editor/leo-editor/assets/847716/1832740b-7e6b-4199-9f78-31f5961b8a2f)

`SessionManager.get_session` should skip those files.

edreamleo:
Closed via PR #3675.

PR #3675, opened 2023-11-27
PR for #3671: don't save unnamed files to session data

edreamleo:
See #3671.


boltex (approved):
Looks good to me! I also tested the code in Leo. Works fine with that change!

Issue #3656, opened 2023-11-18
Bugs re zipped files


edreamleo:
Fixed via PR #3645 and PR #3677.

From Félix:

`LM.isZippedFile` and `LM.isLeoFile` do not cover all the possible cases.

**Original report**

I renamed a `.leo` file to `.zip` or `.xyz` for the following tests:

- Case 1: Opening a *valid* `.leo` file renamed as a `.zip` or `.xyz` puts the content of the file in a `newHeadline` node.
- Case 2: Opening a non-Leo file like `abc.xyz` also puts the contents of the file in a `newHeadline` node.
- Case 3: Opening an *invalid* file ending with `.leo` prints `bad .leo file: x.leo` but still opens a new document with failing panels.

**Update (EKR)**

PR #3645 (merged into devel) changes the reported behavior as follows:

- Case 1 works correctly:
  Leo creates a new outline in a tab called `whatever.xyz` (or `whatever.zip`).
  The outline consists of a single `@edit` or `@file` node containing the expected (xml) contents.
  This is an edge case. Imo, Leo should *not* examine the contents of the file.
- Case 2 works correctly:
  Leo creates a new outline in a tab called `abc.xyz`.
- Case 3 works *incorrectly* as before:
  Leo `.leo` prints `bad .leo file: x.leo` but opens a outline without any nodes!

PR #3677 fixes Case 3: Leo opens the bad `.leo` file in an `@edit` node *without* any error message.

edreamleo:
Fixed via PR #3645 and PR #3677.

PR #3677, opened 2023-11-28
PR for #3656: Open bad .leo files as external files

edreamleo:
See #3656.

- [x] Improve error reporting in `FastRead.readWithElementTree`.
- [x] Open bad .leo files gracefully, treating them as external files.
- [x] Add `lm.openBadLeoFile`.
  Tested by calling this one-line script: `g.openWithFileName('C:\Test\invalid.leo')`.

*Note*: Leo has never tested for binary files before. There is no foolproof way of doing so. ChatGPT4 gives a method that is simply wrong! Imo, we can ignore this edge case.


edreamleo:

boltex (commented):
Looks good to me.

PR #3680, opened 2023-11-30
PR: Fix db quirp by adding a new arg

edreamleo:
- [x] Add an optional `path` kwarg to `fc.initNewDb`.
   This change clarifies the logic while retaining strict compatibility with existing scripts and plugins.
   *Note*: it would be wrong for the `path` arg to default to `c.fileName()` instead of `c.mFileName`.

**Analysis**

The calling sequence for all calls to `fc.initNewDb` is as follows:

- Leo's code contains four calls to `fc.getAnyLeoFileByName(path)`. See below.
- `fc.getAnyLeoFileByName` calls `fc._getLeoDBFileByName(path)`.
- `fc._getLeoDBFileByName(path)` now contains:
    `v = fc.retrieveVnodesFromDb(conn) or fc.initNewDb(conn, path)`

This last call makes explicit that `fc.initNewDb` should use the given path rather than `c.mFileName`.

As mentioned above, four methods call ` `fc.getAnyLeoFileByName`:

- `gdc.make_leo_outline`
- `lm.openExistingLeoFile`
- `lm.openSettingsFile`
- `lm.revertCommander`

For each of these calls, `path` *might* always be equivalent to `c.mFileName`, but it's tricky to say for sure.

**Summary**: This PR most likely clarifies the code's intention.

### DIFF
--- a/leo/core/leoFileCommands.py
+++ b/leo/core/leoFileCommands.py
@@ -1136,7 +1136,7 @@ class FileCommands:
         c.setCurrentPosition(p)
         return rootChildren[0]
     #@+node:vitalije.20170815162307.1: *6* fc.initNewDb
-    def initNewDb(self, conn: Any, path: str) -> VNode:
+    def initNewDb(self, conn: Any, path: str = None) -> VNode:
         """ Initializes tables and returns None"""
         c, fc = self.c, self
         v = leoNodes.VNode(context=c)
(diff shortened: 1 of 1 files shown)


edreamleo:
@boltex Please review and approve this PR.

boltex (approved):
This looks good to me.

edreamleo:
@boltex Thanks!

PR #3688, opened 2023-12-03
PR Fix valid mypy complaints re leoServer.py

edreamleo:
Tested with leoInteg with either Python 3.11 and 3.12.

All tests pass with Python 3.11. Unrelated tests fail on Python 3.12.

- [x] Use the `e` value per other logging code.
- [x] Raise `ConnectionClosed(None, None)` to supply two required args.


edreamleo:
@boltex I'm going to merge this PR immediately.

Please let me know if you see further problems.

Issue #3670, opened 2023-11-26
Ctrl-Click on new-style UNL may not navigate to target in other outline


tbpassin:
It looks like the construction of new-style UNLs makes assumptions about the location of a node's outline.  These assumptions can prevent Ctrl-Click navigation.

Example:
Here is the UNL of a node in LeoPyRef:

`unl:gnx://LeoPyRef.leo#ekr.20140901062324.18565`

If I paste this UNL into a node in my workbook, an attempt to Ctrl-Click on it fails with this error message:

`Not found: 'unl:gnx://LeoPyRef.leo#ekr.20140901062324.18565'`

The same failure occurs with the new version of old-style UNLs.

I have been under the impression that the new UNLs would support navigation to nodes in other outlines. Perhaps this will only work if the two outlines are in the same file tree.

edreamleo:
@tbpassin Control-clicking unls searches all *open* outlines.

This may have implications for your zettlekasten work.

tbpassin:
> @tbpassin Control-clicking unls searches all open outlines.

That is incorrect.  I just tried it again. I pasted a node's UNL into a node in the workbook.  Ctrl-clicking on the UNL string in the workbook failed to navigate to the original node even though its outline was open:

`Not found: 'unl:gnx://os-notes.leo#tom.20231124223010.1'`

tbpassin:
> This may have implications for your zettlekasten work.

It uses gnx's and only within an outline.  I suppose it would have implications if I wanted to reference a node in a different outline, but that would be an ordinary reference - not part of the zettelkasten system.

I liked the old way where Leo would open the right outline if it weren't already open.

edreamleo:
@tbpassin I am also going to delegate this issue to you. Please feel free to ask for help.

tbpassin:
> @tbpassin Control-clicking unls searches all _open_ outlines.

That was what was supposed to happen but there were a few errors that prevented it.  I sorted that out, added a search of the files in the Recent Files list,  and added a few other files one might like to link to, like LeoPyRef.leo and LeoDocs.leo, so that they can be opened and navigated to even if they aren't in the recent files list.

boltex:
@tbpassin @edreamleo I've now made UNL work in LeoJS (and soon in LeoInteg after i add a small modification to the server) :)

[...]

PR #3691, opened 2023-12-03
openUNLFile(): Fix logic and add more places to look for target files…

tbpassin:
….  Enhance docstr.

UNLs do not carry full path information so we have to try to get it from other places.  This version now looks in the following places (added 4) and 5):

        1. The currently selected outline
        2. All open outlines
        3. Files specified in an @data setting named 'unl-path-prefixes'
        4. Certain well-known files like PyLeoRef.leo
        5. The outlines listed in the Recent Files menu.
       
pytest found an assertion failure:

```
FAILED leo/unittests/core/test_leoGlobals.py::TestGlobals::test_g_openUNLFile - AssertionError: None != Commander 1693112220992: 'C:\\Tom\\git\\l[35 c...
```
but I think this must be caused by looking for some file not on my machine or not in an `@data` node.

Fixes https://github.com/leo-editor/leo-editor/issues/3670.


tbpassin:
This PR has been superseded by events and I will close it now.

Issue #3718, opened 2023-12-19
Remove an unnecessary class in leoCache.py


boltex:
See PR #3721.

<details><summary>hypothesis & explanation</summary>
<br>

**Hypothesis**

One of the two databases in .leo/db/ is unused: The one in .leo/db/global_data. (which is used in Leo as **g.app.commander_cacher**)

_(Confusing name btw because the one with the actual 'global' data is in .leo/db/g_app_db, which is used in Leo as **g.app.db**)_

that **g.app.commander_cacher** seems to be used to keep a link to a db that is common to all opened commanders in leo.

But, It's used only directly to commit+save when closing a document, saving a document or quitting leo. Not to get/set data per say. So do those commit+save serve a purpose if nothing is ever set into it? 

I dont think so because that db is not the one that contains any relevant data. Only the global db known as  **g.app.db** contains data.

**Explanation**

The commanders each keep a link to a db to store their specific (non global) data. This is set by the line :

`c.db = g.app.commander_cacher.get_wrapper(c)`

which happens in the initObjects method of the Commands constructor.

Look at the get_wrapper method. It sets its own db to the main g.app.db, which is the global db. therefore the g.app.commander_cacher serves no purpose.

Félix

</details>

boltex:
To add to the previous post: 

Once that .leo/db/global_data _(g.app.commander_cacher)_ db is removed, the commit+save that was done on it when closing a document, saving a document or quitting leo should be done on g.app.db instead I guess ! :)

Félix

edreamleo:
Closed via PR #3721.

PR #3721, opened 2023-12-19
PR for #3718: Remove CommanderCacher class and related code

edreamleo:
See #3718.

**Overview**

The following relationships *remain unchanged*:

- `g.app.db` is a singleton instance of the `GlobalCacher` class.
-  For each commander, `c.db` is an instance of the `CommanderWrapper` class.
   However, Leo creates `c.db` without using the (deleted) `CommanderCacher` class.

The `.leo/db` folder now contains only a `g_app_db` folder.
Leo no longer writes per-commander sub-folders in the  `.leo/db` folder!

**Details**

- [x] Delete the `CommanderCacher` class.
- [x] Instantiate per-commander `CommanderWrapper` instances in `c.initObjects`.
- [x] Remove all references to the `g.app.commander_cacher` and `g.app.commander_db` ivars.
- [x] Fix error recover code in `GlobalCacher.clear`.
- [x] Remove mypy suppression in `GlobalCacher.clear`.

**Extras**

- [x] Remove the never-used `fn` kwarg to `CommanderWrapper.__init__`.
- [x] `app.finishQuit` no longer calls `pyzo_close_handler`.

Later (in a separate PR) or never: `clear-all-caches` deletes entire contents of `./leo/db'.

### DIFF
--- a/leo/core/leoCache.py
+++ b/leo/core/leoCache.py
@@ -100,13 +100,13 @@ class GlobalCacher:
         if 'cache' in g.app.debug:
             g.trace('clear g.app.db')
         try:
-            self.db.clear()
+            self.db.clear()  # SqlitePickleShare.clear.
         except TypeError:
-            self.db = {}
+            self.db = {}  # self.db was a dict.
         except Exception:
             g.trace('unexpected exception')
             g.es_exception()
-            self.db = {}  # type:ignore
+            self.db = {}
     #@+node:ekr.20180627042948.1: *3* g_cacher.commit_and_close()
     def commit_and_close(self) -> None:
         # Careful: self.db may be a dict.
@@ -273,9 +273,12 @@ class SqlitePickleShare:
         return fnmatch.fnmatch(basename(s), pattern)
     #@+node:vitalije.20170716201700.15: *3* clear (SqlitePickleShare)
     def clear(self) -> None:
-        # Deletes all files in the fcache subdirectory.
-        # It would be more thorough to delete everything
-        # below the root directory, but it's not necessary.
+        """
+        Deletes all files in the fcache subdirectory.
+        
+        It would be more thorough to delete everything
+        below the root directory, but it's not necessary.
+        """
         self.conn.execute('delete from cachevalues;')
     #@+node:vitalije.20170716201700.16: *3* get  (SqlitePickleShare)
     def get(self, key: str, default: Any = None) -> Any:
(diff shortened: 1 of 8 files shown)


edreamleo:
@boltex I'm fuzzy about why the calls to ` g.app.commander_cacher.save` aren't needed in `leoFileCommands.py`.

Everything seems to work, but I don't understand why? Could you explain?

edreamleo:
@boltex Oh. I think I understand. There is only one db. The wrapper simply creates commander-specific keys, as you mentioned in #3718. So the save commands need not save anything. Correct?

boltex:
@edreamleo  In Leo, the db is opened in auto-commit mode. this may explain why saves and commits are unnecessary.

`
        self.conn = sqlite3.connect(dbfile, isolation_level=None)
`

The default autocommit is LEGACY_TRANSACTION_CONTROL, which makes the isolation_level=None setting enforce the 'autocommit' behavior. **This is as per: https://docs.python.org/3/library/sqlite3.html#sqlite3.Connection.autocommit**   

notice the _recommended_ VS _default_ value of 'autocommit' which I had to re-read 3 times to make sure this is all true regarding isolation_level having the final word !

(I had found that out while studying the implementation of the sqlite used by Leo for this. and the options it uses when using the python sqlite library.)

Also, to confirm everything you can peek very easily into the db at crucial points in time while using Leo with https://sqlitebrowser.org/ to make sure stuff is properly set in the db when you think it should.

boltex:
I'll test and review this major fix soon.

edreamleo:
@boltex Thanks for your comments. I had no idea the db settings were important.

leo/core/leoCache.py
```diff
@@ -143,9 +65,16 @@ def __getitem__(self, key: str) -> Any:
     def __setitem__(self, key: str, value: Any) -> None:
         self.user_keys.add(key)
         self.db[f"{self.key}:::{key}"] = value
-#@+node:ekr.20180627041556.1: ** class GlobalCacher
+#@+node:ekr.20180627041556.1: ** class GlobalCacher (g.app.db)
 class GlobalCacher:
-    """A singleton global cacher, g.app.db"""
+    """
+    A calss creating a singleton global database, g.app.db.
```
boltex, line 71: class instead of calss

boltex (approved):
Looks Good! Can't wait to also clean that up completely in leojs!!
I also tried it with sessions and expanded/marked node changes, etc.. Everything seems to work just fine 👍

leo/core/leoCache.py
edreamleo: Typo fixed at 5199444.

edreamleo:
@boltex Thanks for your comments and testing. This PR is a big step forward.

Issue #3719, opened 2023-12-19
Markdown (.md) files contain an extra newline when writing when a title node body is 'empty'


boltex:
**EKR**: As noted in PR #3722, this is a round-tripping problem.

**Original Report**

The problem happens when writing @auto nodes of markdown content. (reading seems fine)

The problem arises only if a title node **without body text** has a child of a lower title. 

For example, this valid markdown will be read properly but an extra newline will be inserted between **# 1st level title A** and **## 2nd level title B** when re-written to disk with **ctrl+shift+w** _(it will have 2 empty lines between instead of one)_

Original markdown: (note the absence of text in node "# 1st level title A, which will provoke the double newline insertion when written to file)
```
# 1st level title X

some text in body X

## 2nd level title Z

some text in body Z

# 1st level title A

## 2nd level title B 

some body content of the 2nd node 
```
I understand that when writing a newline is placed below the headline which generates the title (with the proper number of '#', but perhaps a check could be made to see if the body pane contained **specifically nothing, or a single newline**, if so, it should skip writing that extra newline.

Here's a visual example of the problem:

Leo reads [the original markdown](https://github.com/leo-editor/leo-editor/assets/847716/b52cb193-2fd1-4a94-998c-e4435425941b) correctly as shown  [here](https://github.com/leo-editor/leo-editor/assets/847716/cb58e294-2a05-418f-bdc9-60eeef42e242).

But after a write [the 'empty' section now has an extra newline](https://github.com/leo-editor/leo-editor/assets/847716/3b325ed2-0a72-4627-bd4f-8e54ed8081d0). _Note where the cursor row is highlighted_.

The whole read - write cycle would be perfect without this single small annoyance. 

Markdown has become the #1 standard for documentation, so I think this is important.

boltex:
I've spotted why the test didn't work: I've annotated the PR.

While annotating the PR I thought that perhaps the problem could be solved upon reading/refreshing the @auto node instead of when writing it. As I've observed that an extra newline is present in the '1st level title A' node's body upon reading from file. 

So this issue could perhaps be 'fixed' at the reading (importer) of the @auto file instead of the writing. (writer) But I'm not sure about that, just a thought I had. :)

edreamleo:
@boltex Thanks for your comments. I'll investigate possible solutions after getting the unit test to fail :-)

edreamleo:
Fixed via PR #3722.

PR #3722, opened 2023-12-19
PR for #3719: markdown writer

edreamleo:
See #3719.

- [x] Add a failing unit test: `test_markdown_sections`.
- [x] Fix `MarkdownWriter.write`.
- [x] Add a unit test: `test_markdown_image`.
- [x] Fix `MarkdownWriter.write_headline`: write all headlines except `!Declarations`.
- [x] Change `MarkdownWriter.write` so it doesn't write headlines of placeholder nodes.
- [x] Add a unit test: `test_markdown_placeholders`.


leo/unittests/test_writers.py
```diff
@@ -50,6 +52,48 @@ def test_dart_writer(self):
...
+
+            ## 2nd level title B 
+
+            some body content of the 2nd node 
+        """).strip() + '\n'
+
+        c, root = self.c, self.c.p
+        child = root.insertAsLastChild()
+        child.h = 'h'
+        child.b = contents
```
boltex, line 85: The contents is not to be put into a body as a whole directly, it's loaded from an @auto node, not an @edit node. So it builds a tree when 'refreshed from file'. (Could it be that perhaps the extra newline can be omitted when reading into the tree instead of when writing as I first proposed?)

edreamleo: I agree with your analysis.  I thought I had put a comment to exactly that effect here, but apparently I forgot. I'll get the test to fail soon :-)

edreamleo:
@felix Ready for review. The fix was a special case in `MarkdownWriter.write`.

I'm not convinced that there are enough unit tests for markdown.

boltex:
I've reviewed the code, and the bug about inserting extra lines is fixed! 😄 

But trying it out with LeoJS readme markdown file, I've discovered one other bug with markdown writting: (_probably easy to fix, but let me know if I should open a new issue instead_) I've found a particular headline that is not outputed. One that starts with an image The markdown image format is **!\[label]\(imageURL)**

So simply stated, a headline starting with `# label` instead of `# someAlphanumericCharacters`  will strangely not be written... You can see the diff in the image captured below as the very first line of the markdown is missing: 

(in the foreground below, Leo shows the outline with the problematic headline label selected, and above, in the dark editor windows, on the right is the original being read, and on the left is the written output, missing the first line.)
image

edreamleo:
@boltex Thanks for your report of an additional problem. I'll fix it in this PR.

edreamleo:
@boltex Rev 2f33d69 fixes mdw.write_headline. Previously, it didn't write *any* headline starting with '!'. Now it writes all headlines except the special headline `!Declarations`.

boltex (changes requested):
When testing this branch for reliability, I've found that this fixes the title headlines that begin with an image, eg `# name` Yay! Thanks for that!

But, there is one more bug that remains when testing the @auto with markdown files, but I think it would be super easy and simple to fix!

EXPLANATION: It is legal to have level 3 titles under a level 1 title without intermediate level 2 titles. But the importer does not make a level 3 title directly, it makes a 'placeholder level X' node, where X is the missing level in between. 

Of course, this is acceptable because the importer has no choice in order to build a tree that will also generate appropriately the 'level' of the title when writing. 

So, since it was imported successfully and flawlessly without those intermediate levels, (Leo is wise enough to spot that and insert them as I already explained above hehe) The only thing that would make the roundtrip 100% exact and perfect would be that the 'placeholder level X' headlines were **ignored by the markdown writer** !  **By ignored I mean just ignored when actually outputting to file**, so that they're just absent from the output, but their effect on the title level of the logic is computed as normal like before. So no major code change is required here I suppose.

Hope I'm right and that you find that extra small change request acceptable in your view.  Please feel free to ask for more details if this small change does not seem trivial and simple to you. I'll make sure to expose it in a more comprehensive manner to you if its the case :)

boltex:
I thought it might be simpler to see with images. I use the LeoJS readme file to test, here the 'placeholder level X' are both as child of 'ISSUES' and 'ACKNOWLEDGMENTS':

image

And here is the diff that this produces:
image

 _(Note that the extra newline along with the title 'placeholder level X' to be 'skipped' could be either the newline above or below the title, it makes no difference)_

So to reiterate, if the markdown writer could just 'filter out' as not to write headline lines starting with 'placeholder level X' and its following single `'\n'` newline... this would make the roundtrip flawless! 😄

edreamleo:
@boltex Good idea. I shall do as you suggest re not writing placeholders.

edreamleo:
@boltex Rev bd8edc3 does as you requested.

boltex:
I'm out if town I'll review all pending pull requests tonight or tomorrow 🤓

boltex:
@edreamleo I've tried out your modifications and reviewed the code. I should have been more precise and explicit in my description of the problem: The 'placeholder level X' nodes should be 'skipped' as if they did not exist, as they indeed, did not exist in the original file. :)

But, fear not! hehe! **I've found the perfect solution that makes exact round-trips** of large, and complicated (but legal, valid ans linted) markdown files! Even with all sorts of stuff (like in the LeoJS README.md file) now work perfectly with the **write** method provided below. 😄 

This in fact simplifies the modification required. It can all be done simply with the regex you had already in place to check for 'placeholder level X' headlines. (_You then should revert **write_headline** to what it was._)

Instead of modifying your branch directly, I'll provide this improved **write** method right here (or, if you'd like me to push to your branch directly instead, just say so) :

```python
def write(self, root: Position) -> None:
    """Write all the *descendants* of an @auto-markdown node."""
    self.root = root
    self.write_root(root)
    total = sum(1 for _ in  root.subtree()) 
    count = 0
    for p in root.subtree():
        count += 1
        lastFlag = count == total  # Last node should not output extra newline
        if g.app.force_at_auto_sentinels:  # pragma: no cover
            self.put_node_sentinel(p, '<!--', delim2='-->')
        if self.placeholder_regex.match(p.h):
            # skip this 'placeholder level X' node
            pass
        else:
            self.write_headline(p)
            # Ensure that every section ends with exactly two newlines.
            if p.b.rstrip():
                s = p.b.rstrip() + ('\n' if lastFlag else '\n\n' )
                lines = s.splitlines(False)
                for s in lines:
                    if not g.isDirective(s):
                        self.put(s)
            elif not lastFlag:  # #3719.
                self.put('\n')
    root.setVisited()
```

In the tests, the class for markdown in test_writers.py should now look like this:
```python
#@+node:ekr.20231219151314.1: ** class TestMDWriter(BaseTestWriter)
class TestMDWriter(BaseTestWriter):
    """Test Cases for the markdown writer plugin."""
    #@+others
    #@+node:ekr.20231219151402.1: *3* TestMDWriter.test_markdown_sections
    def test_markdown_sections(self):
        
        c, root = self.c, self.c.p
        #@+<< define contents: test_markdown_sections >>
        #@+node:ekr.20231221072635.1: *4* << define contents: test_markdown_sections >>
        contents = textwrap.dedent("""
            # 1st level title X

            some text in body X

            ## 2nd level title Z

            some text in body Z

            # 1st level title A

            ## 2nd level title B

            some body content of the 2nd node 
        """).strip() + '\n'  # End the last node with '\n'.
        #@-<< define contents: test_markdown_sections >>

        # Import contents into root's tree.
        importer = Markdown_Importer(c)
        importer.import_from_string(parent=root, s=contents)

        if 0:
            for z in root.self_and_subtree():
                g.printObj(g.splitLines(z.b), tag=z.h)
            print('\n=== End dump ===\n')

        # Write the tree.
        writer = MarkdownWriter(c)
        writer.write(root)
        results_list = c.atFileCommands.outputList
        results_s = ''.join(results_list)
        if contents != results_s:
            g.printObj(contents, tag='contents')
            g.printObj(results_s, tag='results_s')
        self.assertEqual(results_s, contents)
    #@+node:ekr.20231225025012.1: *3* TestMDWriter.test_markdown_image
    def test_markdown_image(self):
        
        c, root = self.c, self.c.p
        #@+<< define contents: test_markdown_image >>
        #@+node:ekr.20231225025012.2: *4* << define contents: test_markdown_image >>
        contents = textwrap.dedent("""
            declaration text

            # ![label](https://raw.githubusercontent.com/boltext/leojs/master/resources/leoapp.png)

            Body text
        """).strip() + '\n'  # End the last node with '\n'.
        #@-<< define contents: test_markdown_image >>

        # Import contents into root's tree.
        importer = Markdown_Importer(c)
        importer.import_from_string(parent=root, s=contents)

        if 0:
            for z in root.self_and_subtree():
                g.printObj(g.splitLines(z.b), tag=z.h)
            print('\n=== End dump ===\n')

        # Write the tree.
        writer = MarkdownWriter(c)
        writer.write(root)
        results_list = c.atFileCommands.outputList
        results_s = ''.join(results_list)
        if contents != results_s:
            g.printObj(contents, tag='contents')
            g.printObj(results_s, tag='results_s')
        self.assertEqual(results_s, contents)
    #@+node:ekr.20231227225308.1: *3* TestMDWriter.test_placeholders
    def test_markdown_placeholders(self):
        
        c, root = self.c, self.c.p
        #@+<< define contents: test_markdown_placeholders >>
        #@+node:ekr.20231227225358.1: *4* << define contents: test_markdown_placeholders >>
        # There must be two newlines after each node.
        contents = textwrap.dedent("""
            # Level 1

            Level 1 text.

            ### Level 3

            Level 3 text.
        """).strip() + '\n'  # End the last node with '\n'.
        #@-<< define contents: test_markdown_placeholders >>

        # Import contents into root's tree.
        importer = Markdown_Importer(c)
        importer.import_from_string(parent=root, s=contents)

        if 0:
            for z in root.self_and_subtree():
                g.printObj(g.splitLines(z.b), tag=z.h)
            print('\n=== End dump ===\n')

        # Write the tree.
        writer = MarkdownWriter(c)
        writer.write(root)
        results_list = c.atFileCommands.outputList
        results_s = ''.join(results_list)
        if contents != results_s:
            g.printObj(contents, tag='contents')
            g.printObj(results_s, tag='results_s')
        self.assertEqual(results_s, contents)
```

I've tried it out personally with md files, and ran the unit tests also.

edreamleo:
@boltex Thanks for your fixes.

Yes, please push your modifications to the ekr-3719-markdown branch.

boltex (approved):
As per Edward's request, I've added the fix to 'sip very last newline' to ensure perfect round-trips of import/write of markdown files. :)

edreamleo:
@boltex Thanks for making round-trip work!

Issue #3724, opened 2023-12-22
Another 3.12 syntax coloring bug


edreamleo:
Leo does not color this following properly with Python 3.12:
```python
# Leo 6.6.4: preserve spacing.
self.put(f"{'#' * level} {p.h.lstrip()}")
# oops!
```
The f-string does not terminate at the second double quote.

edreamleo:
Closed via PR #3725.

PR #3725, opened 2023-12-22
PR for #3724: another colorizer bug

edreamleo:
See #3724.

### DIFF
--- a/leo/core/leoColorizer.py
+++ b/leo/core/leoColorizer.py
@@ -1909,14 +1909,12 @@ class JEditColorizer(BaseColorizer):
         """
         escape, escapes = '\\', 0
         level = self.f_string_nesting_level
-        alt_delim = '"' if delim == "'" else "'"
+        alt_delim = '"' if delim == "'" else "'"  # Works for triple delims.
         in_alt_delim = False
 
         # Scan, incrementing escape count and f-string level.
-        g.trace(i, 'delim:', delim, s.rstrip())
         while i < len(s):
             progress = i
-            ### g.trace(i, 'escapes', escapes, 'level', level, repr(s[i]))
             if g.match(s, i, delim):
                 if (escapes % 2) == 0 and level == 0 and not in_alt_delim:
                     return i + len(delim)
--- a/leo/core/leoColorizer.py
+++ b/leo/core/leoColorizer.py
@@ -1909,18 +1909,26 @@ class JEditColorizer(BaseColorizer):
         """
         escape, escapes = '\\', 0
         level = self.f_string_nesting_level
+        alt_delim = '"' if delim == "'" else "'"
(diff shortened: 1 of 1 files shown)


tbpassin (commented):
I see what you are doing with the `alt_delimiter`s and it looks all right as far as I can tell.  But I am not familiar with all the ins and outs of f-string quoting so I cannot say if all cases have been covered.  If you want me to I can study up on f-string quoting.

Shouldn't the AST module be able to work this out more  definitively?

edreamleo:
On Thu, Dec 21, 2023 at 10:09 PM Thomas Passin ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
>
> But I am not familiar with all the ins and outs of f-string quoting
>
See PEP 701 <https://peps.python.org/pep-0701/>, especially considerations
regarding quote reuse
<https://peps.python.org/pep-0701/#considerations-regarding-quote-reuse>.

Imo, this PEP was a mistake, for reasons discussed in the PEP itself.
However, there is nothing Leo can do *now* about this mistake.

I considered not changing the colorizer at all, but that point of view
seems a bit petulant :-)

Anyway, you don't have to comment further, Thomas, if you don't want to!

Edward

tbpassin:
I did read that PEP when this issue came up, and I think it's completely wrong-headed.  String interpolation doesn't require you to insert a specific quoted string inside curly braces, it only requires Python to evaluate the contents of the curly brace before completing the construction of the f-string. And why worry that much about "compatibility" with those other languages when Python's use of whitespace is wildly incompatible?

But as you said we seem to be stuck with it for now.

edreamleo:
@tbpassin "Reposting" as a separate comment to allow markup.

I could not have said it better myself.

The PR provides *no use case whatsoever* for the additional complications.

Earlier versions of Python provide straightforward ways of building complex f-strings from previous strings and f-strings.

Here is an example from leoAst.py:
```python
def compute_node_string(self, node: Node, level: int) -> str:
    """Return a string summarizing the node."""
    indent = ' ' * 2 * level
    parent = getattr(node, 'parent', None)
    node_id = getattr(node, 'node_index', '??')
    parent_id = getattr(parent, 'node_index', '??')
    parent_s = f"{parent_id:>3}.{parent.__class__.__name__} " if parent else ''
    class_name = node.__class__.__name__
    descriptor_s = f"{node_id}.{class_name}: {self.show_fields(class_name, node, 20)}"
    tokens_s = self.show_tokens(node, 70, 100)
    lines = self.show_line_range(node)
    full_s1 = f"{parent_s:<16} {lines:<10} {indent}{descriptor_s} "
    node_s = f"{full_s1:<62} {tokens_s}\n"
    return node_s
```

boltex (approved):

Issue #3734, opened 2023-12-27
Crash in Leo's git-diff command if a file has been deleted


edreamleo:
After deleting `lib.rs`, Leo's `git-diff` command crashes as follows:
```console
Traceback (most recent call last):
  File "C:\Repos\leo-editor\leo\commands\editFileCommands.py", line 1311, in get_file_from_rev
    with open(path, 'rb') as f:
         ^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'c:/Repos/leo-editor/leo/rust/orange/src/lib.rs'
```

edreamleo:
Closed via PR #3735.

PR #3735, opened 2023-12-27
PR for #3734: Suppress git-diff errors

edreamleo:
See #3734.

### DIFF
--- a/leo/commands/editFileCommands.py
+++ b/leo/commands/editFileCommands.py
@@ -1312,8 +1312,7 @@ class GitDiffController:
                 b = f.read()
             return g.toUnicode(b).replace('\r', '')
         except Exception:
-            g.es_print('Can not read', path)
-            g.es_exception()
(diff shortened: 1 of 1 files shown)


tbpassin (changes requested):
This needs a bit more thought.
```
        try:
            with open(path, 'rb') as f:
                b = f.read()
            return g.toUnicode(b).replace('\r', '')
        except Exception:
            g.es_print('Can not read', path)
            g.es_exception()
            # This is not an error. The user has inserted or deleted the file.
```
1. Check the file existence first before trying to read it.  There is no reason to try to save a line or two of code by asking the except clause to do it, because:
2. There might be other exception reasons and this code cannot tell them apart. 
3. Better not to use the ancestral _Exception_ class.

edreamleo:
@tbpassin Good idea. I'll do as you suggest.

edreamleo:
@tbpassin I've made the change you requested.

tbpassin (approved):

Issue #3736, opened 2023-12-28
old-style unls do not work


edreamleo:
See PR #3737.

First reported as a quirp by @boltex. This is a serious bug.

The following unl does not work: `unl://#Code-->About this file` because of a bug in `g.findAnyUnl`.

edreamleo:
Closed via PR #3737.

PR #3737, opened 2023-12-28
PR for #3736: old-style unls do not work

edreamleo:
See #3736.

- [x] Fix `g.findAnyUnl`: see the diffs.
- [x] Fix a quirp in `<< About clickable links >>`: Remove redundant `//`.
- [x] Remove the unnecessary `found` var in `<< look for gnx >>` in `g.openUrlHelper`.
- [x] `test.leo`: move the node that tests unls to the top level and update some comments.
- [x] Make the unit test functional: `TestGlobals.test_g_findAnyUnl`.

<details><summary>Won't do</summary>
<br>

**Note**: The present code breaks on the *first* match of `gnx_regex.finditer`, regardless of `target`.

"Simplifying" the code as follows would be a real (and subtle) change. I see no need to do it.

```python
for match in gnx_regex.finditer(line):
    # Don't open if we click after the gnx.
    if match.start() <= col < match.end():
        target = match.group(0)[4:]  # Strip the leading 'gnx:'
        if target:
            if c.p.gnx == target:
                return target
                if p.v.gnx == target:
                    c.selectPosition(p)
                    c.redraw()
                    break
            return target
```

</details>

### DIFF
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -6794,8 +6794,8 @@ def findAnyUnl(unl_s: str, c: Cmdr) -> Optional[Position]:
             unl = unl[len(prefix) :]
             break
     else:
-        if not g.unitTesting:
-            print(f"Bad unl: {unl_s}")
+        # Unit tests suppress this output.
+        print(f"Bad unl: {unl_s}")
         return None
     file_part = g.getUNLFilePart(unl)
     c2 = g.openUNLFile(c, file_part)
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -6779,15 +6779,31 @@ def findAnyUnl(unl_s: str, c: Cmdr) -> Optional[Position]:
     unl = unl_s
 
     if unl.startswith('unl:gnx:'):
-        # Resolve a gnx-based unl.
+        # Init the gnx-based search.
         unl = unl[8:]
         file_part = g.getUNLFilePart(unl)
-        c2 = g.openUNLFile(c, file_part)
-        if file_part and not c2:
-            return None
-        c3 = c2 or c
         tail = unl[3 + len(file_part) :]  # 3: Skip the '//' and '#'
-        return g.findGnx(tail, c3)
+
+        # If there is a file part, search *only* the given commander!
+        if file_part:
+            c2 = g.openUNLFile(c, file_part)
+            if not c2:
+                return None
+            p = g.findGnx(tail, c2)
+            return p  # May be None.
+
+        # New in Leo 6.7.7:
+        # There is no file part, so search all open commanders, starting with c.
+        p = g.findGnx(tail, c)
+        if p:
+            return p
+        for c2 in g.app.commanders():
+            if c2 != c:
+                p = g.findGnx(tail, c2)
+                if p:
+                    return p
+        return None
+
     # Resolve a file-based unl.
     for prefix in ('unl:', 'file:'):
         if unl.startswith(prefix):
@@ -6797,14 +6813,31 @@ def findAnyUnl(unl_s: str, c: Cmdr) -> Optional[Position]:
         # Unit tests suppress this output.
         print(f"Bad unl: {unl_s}")
         return None
+
+    # Init the headline-based search.
     file_part = g.getUNLFilePart(unl)
-    c2 = g.openUNLFile(c, file_part)
-    if file_part and not c2:
-        return None
-    c3 = c2 or c
     tail = unl[3 + len(file_part) :]  # 3: Skip the '//' and '#'
     unlList = tail.split('-->')
-    return g.findUnl(unlList, c3)
+
+    # If there is a file part, search *only* the given commander!
+    if file_part:
+        c2 = g.openUNLFile(c, file_part)
+        if not c2:
+            return None
+        p = g.findUnl(unlList, c2)
+        return p  # May be None
+
+    # New in Leo 6.7.7:
+    # There is no file part, so search all open commanders, starting with c.
+    p = g.findUnl(unlList, c)
+    if p:
+        return p
+    for c2 in g.app.commanders():
+        if c2 != c:
+            p = g.findUnl(unlList, c2)
+            if p:
+                return p
+    return None
 #@+node:ekr.20230624015529.1: *3* g.findGnx (new unls)
 find_gnx_pat = re.compile(r'^(.*)::([-\d]+)?$')
 
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -6778,17 +6778,12 @@ def findAnyUnl(unl_s: str, c: Cmdr) -> Optional[Position]:
     """
     unl = unl_s
 
-    def oops() -> None:
-        if not g.unitTesting:
-            print(f"Bad unl: {unl_s}")
-
     if unl.startswith('unl:gnx:'):
         # Resolve a gnx-based unl.
         unl = unl[8:]
         file_part = g.getUNLFilePart(unl)
         c2 = g.openUNLFile(c, file_part)
         if file_part and not c2:
-            oops()
             return None
         c3 = c2 or c
         tail = unl[3 + len(file_part) :]  # 3: Skip the '//' and '#'
@@ -6799,12 +6794,12 @@ def findAnyUnl(unl_s: str, c: Cmdr) -> Optional[Position]:
             unl = unl[len(prefix) :]
             break
     else:
-        oops()
+        if not g.unitTesting:
+            print(f"Bad unl: {unl_s}")
         return None
     file_part = g.getUNLFilePart(unl)
     c2 = g.openUNLFile(c, file_part)
     if file_part and not c2:
-        oops()
         return None
     c3 = c2 or c
     tail = unl[3 + len(file_part) :]  # 3: Skip the '//' and '#'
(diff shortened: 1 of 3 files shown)


tbpassin (dismissed):
LeoGlobals.py:
```
                    if c.p.gnx == target:
                        return target
                        if p.v.gnx == target:
                            c.selectPosition(p)
                            c.redraw()
                            break
                    return target
```
This loop will return a spurious position if no node's gnx matches the target. There needs to be an `else:` branch to handle this (presumably it would return _None_).

edreamleo:
@tbpassin Good catch. Rev e7b6d2d should suffice, given the context.

edreamleo:
@boltex @tbpassin I am going to mark this PR as a draft until I can get all the tests in tests.leo to work.

edreamleo:
@boltex @tbpassin For the first time there is a functional unit test that passes.

This PR is ready for review.

boltex:
@edreamleo  I've tried the new behavior, **the 'gnx' type UNLs now work** without the optional 'outline' part (containing only the gnx) like such.

`unl:gnx://#felix.20231228140244.1`

So thanks for that! 😄 

But, using the 'short unls' _which do not provide the **full path** of the outline part_ fail to open, nor switch to another document. (even if it's already opened)

 In the _**screenshot below,**_ I illustrate the situation, I've got the testmd3.leo file opened and ctrl+click on a short link pointing to the testmd1.leo file. (you can see its even already opened on the left tab)  such as 
 
` unl:gnx://testmd1.leo#felix.20231224140524.1` or `unl://testmd1.leo#@button test2`

The ctrl+click only yield 
**Not found: 'unl:gnx://testmd1.leo#felix.20231224140524.1'** 
or 
**Not found: 'unl://testmd1.leo#@button test2'** in the terminal

_(Although they do work if I'm already in testmd1.leo, like the ones you've already fixed that do not contain the optional 'outline' part.)_

image

edreamleo:
@boltex "even if it's already opened." OMG! You've said the magic word!

`g.findAnyUnl` should search *all open commanders, but *only* if the unl gives no file part at all!!!

And now it does! Moreover, the clarity of purpose simplifies the code even though code is a bit longer.

`test.leo` contains new tests and improved comments. All the unls in that file work as expected.

edreamleo:
@boltex Please review this PR again. The recent changes are a big step forward.

boltex:
I've tried (I've made sure to pull your changes) and it does not work. I've tried both with your test file and with my test file, and none of the 'short' UNL links work to goto another file, **even if the file is already opened**.

To be more specific and make the precise problem more apparent: The ones that do not work now are the ones with the 'short' outline path. (The ones that give the filename only in the outline path, but not the full path) 

Here is how to reproduce the bug: 
Make 'short' UNLs **making sure that they work well inside their own document**, then, SAVE-AS a new name (I've replaced '1' with '3' in the filename, as seen in the tabs) and now, clicking on the 'short' link from the testmd3.leo that points to a node in testmd1.leo, **which is already opened on the left**, fails and prints either 
**Not found: 'unl:gnx://testmd1.leo#felix.20231224140524.1'** for 'short' gnx UNLs
or
**Not found: 'unl://testmd1.leo#@button test2'** for 'short' headline / old style UNLs

image

The logic should be that If a 'short' outline path is detected, just loop all opened commanders and try to match the **filename** and **extension** (not the full path), selecting that commander if found, and then finally trying to select a headline or GNX match... this should do the trick 😆 

(_Leaving the commander selected and active, even if no node match is found. making the UNL then be a 'commander' link only instead of a full node link as a fallback_)

English is not my first language, so there's also the possibility that I'm explaining the opposite of what I'm trying to say, hehe!  If you'd like me to illustrate more precisely a concrete example to see the problem more directly, I'll gladly share my screen in a zoom session. :)

Hope this helps ! 😄

edreamleo:
@boltex Thanks for your comments. There may be an `@data` setting involved. I'll recheck soon.

tbpassin:
> The logic should be that If a 'short' outline path is detected, just loop all opened commanders and try to match the **filename** and **extension** (not the full path), selecting that commander if found, and then finally trying to select a headline or GNX match... this should do the trick

If the file name is not an absolute path, we need to remember to check alternative locations if the UNL's file name is not among the open outlines.  Alternative locations to check:

1. The current directory;
2. ~/.leo;
3. The paths in the `@data` setting that Edward introduced;
4. The Recent Files list;
5. Certain well-known Leo outlines, such as LeoPyRef.leo, the Cheat Sheet, etc.

boltex:
historical Note: This is the very last feature/bug completing the rewrite/code review of Leo while making LeoJS. (A few weeks back I thought we were done, but I had yet to implement UNLS ! haha! Now we're really at the last part - for real this time! 😆 )

tbpassin:
Nothing like re-implementing to unearth weaknesses!

edreamleo:
@tbpassin @boltex

Rev 1145ef1 adds an `@data unl-path-prefixes` to `test.leo` and adds some comments to `g.openUNLFile`. Imo, that's all that is needed.

@tbpassin I disagree with you about default paths. The proposed change is too complex for the dubious potential gain. Yes, `@data unl-path-prefixes` is an obscure setting, but it works. There is no need for anything better.

@boltex Changing `@data unl-path-prefixes` in `test.leo` should make all the unls in that file work for you, as they do for me.

**Summary**: Please approve this PR :-)

edreamleo:
@boltex Congratulations on your milestone!

boltex (approved):
Looks good. 
I'll probably create a new issue in the near future to get UNLs to resolve short unls with the current commander's cwd (current working dir) as a base path to check for possible existing file there.

edreamleo:
@boltex Thanks for the approval. I agree: we should discuss any other changes in a new issue.

edreamleo:
@boltex @tbpassin Tracking this issue in Leo 6.7.8 is tricky because this PR was merged into Leo as part of 6.7.7.

Please look at the diffs for  `test_leoGlobals.py`. The new tests will pass *provided* `@data unl-path-prefixes` is set correctly.

I now also remember adding tests to `test.leo`, as shown in the diffs. Those hand tests also pass on my machine.

boltex:
@edreamleo Sorry, I'm completely lost here: what changed? the last modification on this was on december 31st... I dont get it..?

tbpassin:
I think that @edward thinks that your problem of not finding the files to open is because their path is not specified in the `@data` setting.  So nothing has changed, and there is no issue.

What I thought I understood from your post, though, didn't sound like that.  I read it as saying that clicking on a relative unl didn't navigate to another outline **in the same directory**, which it should and which I seem to remember that I had tested at some point.

My understanding of the current intended behavior is this:

1. If the path-based unl does not include a file name, try to navigate there within the current outline;
2. If the unl includes a relative file name, try to go to the right place in that outline if it's open or try to open an outline of that name in the current outline's directory;
3. If the unl includes a relative file name and that file is not in the current directory, try to find it on one of the paths specified in the `@data` setting.
4. I didn't realize this but @felix reported it, if the unl includes a complete path, open that outline if not already open and navigate to the right  node.

boltex:
@tbpassin Who's @felix? I'm @boltex 😄 

Haha! Thanks for your clarifications. This supports my theory that perhaps my explanations was not understood by @edreamleo . (sorry I'm french! I often mush up my explanations!)

It is indeed the # 2 in the list you've laid out that does not seem to work among all those choices.

It is easy to test by the way, just make a dummy leo file somewhere on your harddisk (outside the Leo's home folder or expected path in some other @data setting) create a couple random nodes, and copy the unl shown status bar at the bottom of the Leo window into the body of another sibling node, _test it by ctrl+clicking on it... to see if it at least work within it's own document,_ 

Then, the real test is to  save the file as a new file with a different name **in the same folder**, close Leo, and reopen Leo by clicking on that new file which contains the unl link pointing to the original file, **and then to test it by ctrl+clicking that unl.**

I hope it is (relatively) easy to fix, as other systems in Leo (external files, etc. such as nodes with headlines like : **@clean myFile.txt** ) do find the existence of relative files easily with `g.os_path_exists` ... 

This 'short' unl standard is the most apparent and discoverable features of unls as it is copy/pastable easily by newbies because it's shown by default in the status bar. And so it would be expected that such unl links in Leo files residing in the same folder will resolve to finding the expected node in that other Leo file.

boltex:
@edreamleo Sorry about my confusion - I've re-read the whole thread again a couple times, trying to wrap my head around the expected behavior, and mostly also what "unl-path-prefixes" is, and how to use it... I think i'm slowly starting to get it... 

But I keep coming back to thinking that perhaps it Should default **in leoSettings.leo** at least to "." (a dot) meaning the current commander's file directory?  if possible? (So that the default for new users, and people who don't know about those subtle details, would at least be that an UNL with a simple filename would at least be found naturally, if resting in the same folder)

I think this would be a killer feature of Leo!  I'd like to demo this cool UNL copy-from-status bar and use it quickly and easily to reference nodes in other Leo file of a same project folder in my upcoming Video Tutorial 😄  

Hope this makes sense (again sorry for my lack of experience with UNLs, I've just discovered those in the last month or so!)

Félix

edreamleo:
@boltex No need to apologize. This is a confusing topic.

Some thoughts:

- A default for `@data unl-path-prefixes` would be less useful than it might seem.
  A default might cause more confusion.
  This setting is one place where explicit is better than implicit.

- Most users will likely only need one `@data unl-path-prefixes` node, in `myLeoSettings.leo`,
  but individual `.leo` files can also define `@data unl-path-prefixes`.

We can discuss this further in a *new* issue if you like.

boltex:
@edreamleo Thanks! Will do! :)

Issue #3738, opened 2023-12-28
Wrong folder/filename givin in title bar when showing files from the .leo folder


boltex:
To reproduce, open the workbook.leo in the .leo folder of your user on your windows machine (usually in C:\Users\YOURNAME\.leo )

I think it grabs the first ".leo" it finds instead of the last and therefore thinks the filename is the last part of the path as shown [here](https://github.com/leo-editor/leo-editor/assets/847716/e348ba0a-9e08-4cab-88cb-cbed992913d0).

tbpassin:
I just tried this and got the same ".leo" in the title bar after I opened another Leo outline in the .leo folder.  But then I opened another instance of Leo and repeated the test and did not get the glitch.

So the problem may be a little more subtle, but I definitely see it sometimes.

boltex:
@tbpassin Thanks for this reproduction experiment! 

Sometimes, subtle details like the one you've just spotted help a lot in pin-pointing the actual bug & fixing it.

boltex:
@tbpassin @edreamleo I've done more tests to illustrate the problem more precisely and reproduce the bug reliably:

Here, After doing 'save as' on testMd1 renaming it testMd2, and then re-opening testMd1 from the recent files menu, I get those two files opened from the same folder, but with title bar somewhat tangled/wrong for testMd2, ( for comparison, the second tab for testMd1 below is ok)
![image](https://github.com/leo-editor/leo-editor/assets/847716/e9373748-7a39-437a-be29-bf776087e5aa)

![image](https://github.com/leo-editor/leo-editor/assets/847716/b9e72b72-5395-4f8a-ab99-dab3aa87949a)

edreamleo:
@boltex @tbpassin This issue is a tiny nit.

Are either of you interested in handling it? I'm not.

edreamleo:
@tbpassin Thanks for volunteering!

tbpassin:
Very strange.  I reproduced my case where the filename in the title was only ".leo", then ran c.computeWindowTitle() in that outline and it returned the correct name.

tbpassin:
So far the only case where I can reproduce this bug is in a particular Leo window I've had open for days.  I have several outlines open in it.  If I open another outline, then I see the bug.  But when I open another Leo window and open the same outlines, then I don't get the bug.  I'm going to close that Leo window and see if re-opening it will still exhibit the same bug.

tbpassin:
[...]

PR #3742, opened 2023-12-31
Fix incorrect Window title in some edge cases.

tbpassin:
A change to one line.

### DIFF
--- a/leo/plugins/qt_frame.py
+++ b/leo/plugins/qt_frame.py
@@ -4637,7 +4637,7 @@ class TabbedFrameFactory:
         if not f:
             return
         c = f.c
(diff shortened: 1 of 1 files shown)


boltex:
@tbpassin Hmm, this 'title' parameter was added just to be used at this specific line... I'll let @edreamleo clear this up.

 (Maybe something else was changed since that addition 3 months ago which interferes with this call to computeWindowTitle)

edreamleo:
@tbpassin Is this PR a fix for #3738?

tbpassin:
> @tbpassin Is this PR a fix for #3738?

Yes it is.

Issue #3744, opened 2024-01-01
Create a tokens-only version of Leo's beautifier


edreamleo:
See PR #3757.

This work (in Python) will is a preliminary project (prototype) for #3662: Rewrite Leo's beautifier in Nim:
- Writing a Python parser in Nim is out of the question.
- Transliterating Python *tokenizer* to Nim should be straightforward.

Furthermore, a token-only version of `Orange` should be easier to maintain.

Finally, LeoJS could use a token-only beautifier, but there is no urgent need to do so.

**New Files**

- I wish to maintain `leoAst.py` and `test_leoAst.py` as they are (except for slight improvements.
  The `Orange` and `Fstringify` classes demonstrate the power of the `TokenOrderGenerator` class.

- The new `leoTokens.py` and `test_leoTokens.py` files demonstrate a tokens-only version of Leo's beautifier.
  Creating completely separate files simplifies annotations and eliminates unions.
  Doing so is a prerequisite to transliterating Leo's beautifier to Nim.

**Background**

The `TokenOrderGenerator` class is fragile. New versions of Python often break it.
As a result, the `Orange` class is also fragile.

The revised `Orange` class will use a **token-based scanner** to resolve context questions.
This scanner will be a recursive-descent parser, but limited to parsing function definitions and (maybe) function calls.

**Testing**

Leo's existing unit tests cover all common PEP8 cases.
The stakes are small: bugs in the beautifier are mostly harmless.

tbpassin:
I think I remember reading that the Python PEG parser can be used on its own.  Or if not, here's a Python only implementation of a PEG parser:

https://github.com/erikrose/parsimonious

In the long run it might be worth using a PEG parser for the language modes.  Each mode would need a grammar file, then the parser could generate the mode file.  If this works it would simplify adding new languages.

edreamleo:
@tbpassin Thanks for this link. It's just in time :-)

edreamleo:
Completed (at last!) via PR #3757.

PR #3745, opened 2024-01-02
PR for #3744: token-only beautifier

edreamleo:
See #3744.

- [x] Add `use_ast` module-level switch and honor it.
- [x] Remove unused `Token.level` ivar!
- [x] Don't split lines unless `use_ast` is true.
- [x] Suppress several unit tests when `use_ast` is false.

**Status**

- When `use_ast` is false the beautifier always puts spaces around equal signs.
  This causes two unit tests to fail and incorrectly reformats many lines in `leoAst.py`.


edreamleo:
I am going to close this PR. I'll try again with code based on ekr-3751-split-token-classes branch, PR #3751.

Issue #3767, opened 2024-01-20
Easy install on GitHub


edreamleo:
See PR #3775.

`pip install -r requirements.txt` (from inside the cloned repo) should install all of Leo's requirements.

edreamleo:
Closed via PR #3775.

PR #3775, opened 2024-02-02
PR for #3767: easy install from GitHub

edreamleo:
See #3767.

- [x] Delete `setup.py` and `new-requirements.txt`.
- [x] Add dependency data to `pyproject.toml` for `python -m build`.
- [x] Update `requirements.txt` *by hand* instead of using `pip freeze`.
- [x] Tweak `setup.cfg` for `pip show leo`.
- [x] Tweak `MANIFEST.in` to reduce warnings.
- [x] Update syntax coloring and other files to support `.toml` files.
- [x] Add `pip-install-r.cmd` and `pip-uninstall-all.cmd` to `leo/scripts`.
   I used private versions of  `pip-install-r.cmd` and `pip-uninstall-all.cmd` to test the install/uninstall process.
   The documentation will tell users to complete installation using `pip install -r requirements.txt`.
   Users should *not* use the equivalent of `pip-uninstall-all` !
- [x] Add annotations for several files to keep the updated mypy happy.
- [x] **Extra**: Fix a crasher when using `--gui=null`.

**Won't Do**: I have no plans to allow installation w/o the Qt gui.

The following be part of new documentation, which is *not* part of this PR:

**When installing from `PyPi`**

`pip install leo` and `pip uninstall leo` should work as before.

**When installing from GitHub**

Users will do the following *once*:

- `git clone https://github.com/leo-editor/leo-editor.git`.
- `pip install -r requirements.txt`

This does *not* install Leo, so `pip uninstall leo` will *not* work.

Users can use `pip uninstall -r temp_requirements.txt` to undo `pip install -r requirements.txt`.


edreamleo:
@tbpassin @maphew @boltex requirements.txt now specifies PyQt6.

I know of no way to fall back to PyQt5. I don't think that matters for GitHub-based installs, but I could be wrong.

edreamleo:
@tbpassin @maphew @boltex 

Further testing after `pip cache purge` indicates that `pip install -r requirements.txt` installs *both* PyQt5 and PyQt6.

Presumably one of Leo's requirements pulls in PyQt5. So there seems to be no need to worry about PyQt.

tbpassin:
> Presumably one of Leo's requirements pulls in PyQt5

I don't think so.  I just pip-installed Leo 6.7.7 into a new venv, then ran `pip show pyqt5':
```
(leo) C:\Tom\venvs\leo\Scripts>pip show pyqt5
Name: PyQt5
Version: 5.15.10
Summary: Python bindings for the Qt cross platform application toolkit
Home-page: https://www.riverbankcomputing.com/software/pyqt/
Author: Riverbank Computing Limited
Author-email: info@riverbankcomputing.com
License: GPL v3
Location: C:\Tom\venvs\leo\Lib\site-packages
Requires: PyQt5-Qt5, PyQt5-sip
Required-by: leo, PyQtWebEngine
```
So it's only used by Leo.  Maybe some plugin asks for it?

tbpassin (approved):

tbpassin:
Ah, yes, of course.  The equivalent of PyQtWebEngine for pyqt6 has "qt6" in its name.

Issue #3777, opened 2024-02-04
Failed unit test re leoTokens.py


edreamleo:
See PR #3779.

Viktor and Thomas report similar errors:
```Console

======================================================================
FAIL: test_blank_lines_after_function_3 (unittests.core.test_leoTokens.TestTokenBasedOrange.test_blank_lines_after_function_3)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/user/PyVE/GitHub/Leo/leo-editor/leo/unittests/core/test_leoTokens.py", line 327, in test_blank_lines_after_function_3
    self.assertEqual(results, expected)
AssertionError: "\\\n[122 chars]omment 1.\n        # Trailing comment 2.\n    print('3')\n" != "\\\n[122 chars]omment 1.\n        # Trailing comment 2.\n        print('3')\n"
  \
      def writeAsisNode(self, p):
          print('1')

          def put(s):
              print('2')

          # Trailing comment 1.
          # Trailing comment 2.
-     print('3')
+         print('3')
? ++++


----------------------------------------------------------------------
Ran 931 tests in 7.797s

FAILED (failures=1, skipped=5)
(Leo) [user@fedora-leo-study-vm leo]$
```

edreamleo:
Closed via PR #3779.

PR #3779, opened 2024-02-04
PR for #3777: failing unit test re leoTokens.py (linux only)

edreamleo:
See #3777.  This PR tested on Windows 11 and Ubuntu.

This failing unit tests revealed several problems with existing unit tests:

**Milestones**

- [x] Remove all backslashes from `test_leoAst.py` and `test_leoTokens.py`.
   Many `"""\` instandes remain in other test files, but that will be another PR :-)
- [x] *All* tests files continue to work after being beautified.
- [x] Create `leo/unittests/misc_tests` folder containing all top-level tests *except* `py3_test_grammar.py`.
   That file must *never* be beautified.
- [x] `tbo.cmd` beautifies all test files (except `py3_test_grammar.py`).


tbpassin (commented):
test_leoAst.py line 1957:
```
expected = textwrap.dedent(contents.strip() + '\n')
```

Is `contents` expected to be a multi-line string?  `contents.strip()` removes all leading whitespace so `textwrap.dedent()` won't do anything for a single-line string.  OTOH, if it's a multi-line string, stripping the leading whitespace will throw off the relationship between the first and following lines.

Line 2010 - same comment

The idiom `contents = r"""` places a newline at the start of the first line. This will be fine if the string is going to be executed as code , but if it is going to be used in a string comparison it may not be.  I just want to be sure that the extra line has been accounted for everywhere this idiom is used.

edreamleo:
@tbpassin Yes, typically the "expected" string is multi-line, but the pattern works regardless.

PR #3790, opened 2024-02-14
Added support for a 'get UNL' command to the leo server.

### LINKED ISSUE #3789: Support a 'get-unl' command in Leo's server
This would be nice to display the current node's unl like in the original Leo's status bar of any client connected to a leoserver instance.

(This should have been added in the last server revision. Why didn't I think of that? Let's just say I discovered UNLs very late in the process :) hehe! )

boltex:
Fixes #3789

### DIFF
--- a/leo/core/leoserver.py
+++ b/leo/core/leoserver.py
@@ -2596,19 +2596,40 @@ class LeoServer:
         return self._make_minimal_response(response)
     #@+node:felix.20240213234032.1: *5* server.get_unl
     def get_unl(self, param: Param) -> Response:
-        """Return UNL for specific position, or currently selected node"""
+        """
+        Return UNL for specific position, or currently selected node.
+        This defaults to using the normal status bar UNL indicator method 
+        unless 'short' or 'legacy' boolean parameters are used.
+        """
         c = self._check_c(param)
         p = self._get_p(param)
         unl = ""
-        try:
-            if p and p.v:
+        if p and p.v:
+            # Set a method to get an UNL: either specific, or the default status-bar method.
+            if (hasattr(param, 'short') or hasattr(param, 'legacy')):
+                # Parameter given: Specific UNL method
+                method = p.get_short_gnx_UNL  # Default to short gnx UNL.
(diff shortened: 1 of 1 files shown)


edreamleo:
@boltex I didn't see this PR because there was no 6.7.8 milestone.

edreamleo (changes requested):
Using `c.config.getString('unl-status-kind')` seems like a bad idea. 

The following is my version from PR #3793 (now closed in favor of *this* PR):

```python

 def get_unl(self, param: Param) -> Response:
        """
        Return the unl as it appears in the status line for given position.
        
        By default, this is the full legacy (headline-based) unl.
        """
        tag = 'get_unl'
        p = self._get_p(param)
        valid_kinds = (
            'full-gnx',  #    'unl:gnx:' + f"//{c.fileName()}#{self.gnx}"
            'full-legacy',  # 'unl:'     + f"//{c.fileName()}#{'-->'.join(<parents_headline_list>)}"
        )
        kind = getattr(param, "unl-kind", None) or 'full-legacy'
        if kind not in valid_kinds:
            raise ServerError(f"{tag}: invalid unl-kind: {kind!r}")
        if kind == 'full-gnx':
            unl = p.get_full_legacy_UNL()
        else:
            unl = p.get_full_gnx_UNL()
        return self._make_minimal_response({"unl": unl})
```
What do you think?

edreamleo (changes requested):
Also, I recommend using either ` p.get_full_legacy_UNL()` or `p.get_full_gnx_UNL()`.

That explicitly returns (to LeoInteg) the full file name, not the directory name.

edreamleo:
@boltex This PR clearly affects LeoInteg, but unless I am mistaken LeoJS is likely to want to use similar logic.

Within LeoJS, `c.config.whatever` is irrelevant. The local (LeoJS) settings seem like they would determine the format of the unls.  Do you agree?

edreamleo (approved):
Thanks for making these changes.

boltex:
oops this was not my final commit on this, its still full of bugs!

PR #3797, opened 2024-02-15
Fix leo server bugs with 'get unl' and other 'hasattr' misuses,

### LINKED ISSUE #3796: Fix leo server bugs with 'get unl' and other 'hasattr' misuses,
little bugs to fix in recent UNL feature of leoserveer.

boltex:
Fixes #3796

### DIFF
--- a/leo/core/leoserver.py
+++ b/leo/core/leoserver.py
@@ -2603,34 +2603,34 @@ class LeoServer:
         """
         c = self._check_c(param)
         p = self._get_p(param)
-        unl = ""
-        if p and p.v:
-            # Set a method to get an UNL: either specific, or the default status-bar method.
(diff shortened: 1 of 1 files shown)


boltex (commented):
My bad: I should have turned the other pull request into 'draft' state when i started working on it earlier , This fixes the early merge of 3790.

boltex:
I used your #3798 as starting point and fixed the hasattr(dict, prop) => prop 'in' dict to check properly for attribute existence in a dict. 

(deformation from javascript web development where everything is an object and there is only one single way to check if an object has an attribute which I tend to forget when I code in python - why did guido not unify those two concepts ? lol )

edreamleo (approved):

Issue #3800, opened 2024-02-16
Back Arrow Does Not Always Return To The Last Position


tbpassin:
See PR #3806. This PR changes the logic of *all* the methods of the `NodeHistory` class.

In the forward/back arrows, clicking the back arrow does not always return to the last position.  This seems to be a change to me, probably a fairly recent change or I would have noticed.

To verify:
Use an open outline, such as the Workbook, that has a good number of nodes.
1. Navigate to the last node (this is for convenience; the behavior does not depend on it);
2. Pick some node higher up the screen, and in the Nav tab, type part or all of its headline.  This will navigate to the chosen node;
3. Use the back arrow to move back to the starting node.
4. Pick some other node and repeat steps 2 and 3. Notice that the back arrow did not return you to the starting node.  Instead, it first visited the 2nd node.

You can keep navigating (using the Nav tab) to other nodes, and the back arrow will navigate to all of them in turn before getting back to the previous node.  

Expected behavior:  The back arrow will actually take you back to the previous position.

The current behavior is very inconvenient.

edreamleo:
@tbpassin Thanks for this report.

Somehow I have been putting up with this annoying hangnail. Thanks for reminding me of it.

tbpassin:
I had been suffering with it without really noticing.  Then one day I was doing a lot of hops back to the starting point but they didn't go right there. It became really annoying.

PR #3806, opened 2024-02-19
PR for #3800: goto prev/back

edreamleo:
See #3800. 

This PR removes the *extremely evil* effects of kwargs and switch-like ivars.

- [x] Eliminate the evil `skipBeadUpdate` ivar.
- [x] Rewrite `NodeHistory.goPrev/goNext`.
   These methods delete entries in the node list if the position no longer exits.
- [x] Rewrite `NodeHistory.update`, eliminating the evil `changed` kwarg.
  Fix the **central bug** of the old code:
  The code must insert new beads in the *middle* of the bead list, *not* the end!
- [x] Rewrite `NodeHistory.select` w/o the evil `skipBeadUpdate` ivar.
- [x] Simplify the `go-back` and `go-forward` commands in `commanderOutlineCommands.py`.

### DIFF
--- a/leo/commands/commanderOutlineCommands.py
+++ b/leo/commands/commanderOutlineCommands.py
@@ -689,27 +689,13 @@ def findNextClone(self: Cmdr, event: Event = None) -> None:
 def goNextVisitedNode(self: Cmdr, event: Event = None) -> None:
     """Select the next visited node."""
     c = self
-    p = c.nodeHistory.goNext()
-    if p:
-        c.nodeHistory.skipBeadUpdate = True
-        try:
-            c.selectPosition(p)
-        finally:
-            c.nodeHistory.skipBeadUpdate = False
-            c.redraw_after_select(p)
+    c.nodeHistory.goNext()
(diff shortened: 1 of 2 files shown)


edreamleo:
@tbpassin @boltex Please review this PR. All my hand tests pass. The new code is **much** better than the old!

This PR will remain a draft you comment.

@boltex This PR will probably affect LeoJS.

tbpassin (approved):
Wow, was I naive!  I didn't realize how involved this history navigation is.  I thought that every position change that wasn't caused by traversing the history list would get appended to its end.  I suppose that would be too simple because it doesn't take into account potential branching in the middle.

I didn't notice anything to remark on in the new code, but the proof will be in using it in practice.

boltex (approved):
Looks good to me. I'll implement this in leojs tomorrow and will test & experiment with those changes :)

edreamleo:
@boltex @tbpassin Thanks for your review. I'll merge this PR now so that we can all test it.

As you say, more testing is required, but I'm confident that the new code is way better than the old.

PR #3808, opened 2024-02-20
Leo babel 2

SegundoBob:
This revision does the following:

* Implements a new option "Leo-Babel-Sudo" to run the script using sudo.  The default setting for this option is off.

* Implements a new option "Leo-Babel-Tab-Babel" to use a "Babel" tab in the Log pane for all Leo-Babel output.

* Implements line prefix options to indicate the type of each line in the "Babel" tab of the Log pane.

* Implements a new option "Leo-Babel-Poll-Delay" to specify the output polling delay in milliseconds.  Previously the delay was 1000.  Now the default is 1.

* Inserts a shebang line as the first line in every script file.  This isn't needed by Leo-Babel which specifies the interpreter for the script file, but it allows the user to safely run the script without specifying the interpreter.

* Instead of the script file always having the extension (if any) specified by the Leo-Editor script-file-path parameter, a shell script now has extension .sh and a Python script has extension .py.

* Implements the display of the following information just before executing a script:

    * The "Current Working Directory" for the script.
    * The list of command line arguments to be executed.

* Changes the default color for "Leo-Babel-stdout"

* Changes the default color for "Leo-Babel-stderr"

* Deprecates Babel Parameter "Leo-Babel-Node-Creation-Default" in favor of "Leo-Babel-Node-Creation".

* Deprecates Babel Parameter "Leo-Babel-completion" in favor of "Leo-Babel-information".

* Changes the default color for "Leo-Babel-information"

* Changes the default for Babel Parameter "Leo-Babel-Python".

* Changes the default for Babel Parameter "Leo-Babel-Shell".


SegundoBob:
EKR or TBP,

Is there something wrong with this pull request or will you eventually get around to it?

Respectfully,
SegundoBob

tbpassin:
I don't think I'm involved here...

tbpassin:
Looking through the PRs just now, I checked on this one and I noticed that the code for babel.py isn't in LeoPyRef.leo.  There is only a node for `__init__.py` (which is an empty file anyway).  I don't know why that is, not even if it's intentional or a mistake.

I declined to work on this PR in the past because I didn't know what it does or how it is intended to work.  I hate to see it languishing, though. I suggest that the PR be given a label of "Code" and a milestone of 6.7.9.  Edward is likely not to look at anything without a milestone.

SegundoBob:
Thomas,

Thanks for the information and advice.

I have never touched leo-editor/leo/core/LeoPyRef.leo.  I don't know who added the references to Leo-Babel.  They are odd and probably of no use to anyone, but I don't know the purpose of referencing Leo-Babel in leo-editor/leo/core/LeoPyRef.leo.

I do all my work on Leo-Babel using the .leo files in the subtree leo-editor/leo/plugins/leo_babel.

I would recommend only referencing the code the implements the Leo-Babel plugin---not the examples, results, or tests. That is, I recommend only 4 or 5 nodes to leo-editor/leo/core/LeoPyRef.leo that reference Leo-Babel:

@file ../plugins/leo_babel/babel_api.py
@file ../plugins/leo_babel/babel_kill.py
@file ../plugins/leo_babel/babel_lib.py
@file ../plugins/leo_babel/babel.py
@nosent ../plugins/leo_babel/__init__.py

Since ../plugins/leo_babel/__init__.py is an empty file required by Python import, the reference to it may not be needed?

---
SegundoBob

SegundoBob:
Thomas,

As far as I can determine, I am not allowed to add a tag or a milestone to my pull request.  If you think the tag and the milestone are appropriate, I would appreciate it if you would add them.

---
SegundoBob

tbpassin:
Oh, I see. All right, I've added the "Code" label and the milestone.  I'll come back and do a review in the next few days.  It won't be very throrough since I am not familiar with the code, as I said, but I can scan the changes and see if I happen to notice something useful.

tbpassin (commented):
babel_lib.py
- lines 92ff - don't need six() any more, since Leo is Python3 only.  After removing these methods, the "six" package won't need importing either.
- Lines 139ff, unl2pos: the form of UNLs has been changed; there is a "legacy" form which is mostly like the old path-based UNLs, and the new one which uses the gnx.  Code that uses UNLs probably should be checked to make sure that the UNL changes don't cause trouble.
- Line 416 - could move the method call inside the f-string in braces
- Lines 456, 457 - could make the lambdas more readable by ending each line with the line continuation token (i.e., a backslash).  This would allow the members of the tuple to be easier to identify and read.  Lambdas work with the line continuations (I tried it once). For example:
```
 itOut = leoG.IdleTime(lambda ito,\
            prefix=babelCmdr.babel_prefix_stdout,\ 
            color=babelCmdr.babel_color_stdout,\
            fdr=reo,\
            babelCmdr=babelCmdr: itf(prefix,\
                                   color, fdr, babelCmdr),\
            delay=babelCmdr.babel_polling_delay)
```

Also, is this syntax really what is intended?
```babelCmdr=babelCmdr: itf(prefix, color, fdr, babelCmdr)```

It looks like a type specification but I don't understand the `babelCmdr=babelCmdr:` part or the dynamic specification.  I don't have to understand but please double-check it's what is intended.

Line 629: `cmdrRes: Leo-Editor command for resultsRoot` should "command" be "commander" (cf line 673)?

Otherwise, I didn't review any of the doc files.

**Comment by SegundoBob** [comment]
Thomas,

Thanks.   I hope to address the issues you raise within 7 days.

---
SegundoBob

**Comment by tbpassin** [comment]
One of my comments was about `babelCmdr=babelCmdr: itf(prefix, color, fdr, babelCmdr)`. I just realized that it's the end of the argument list for the lambda.  So never mind that comment.

This mistake of mine does highlight how helpful good visual organization of the code can be. I think this change to the format would make things clearer:
```
itOut = leoG.IdleTime(lambda ito,\
            prefix=babelCmdr.babel_prefix_stdout,\ 
            color=babelCmdr.babel_color_stdout,\
            fdr=reo,\
            babelCmdr=babelCmdr:\
                itf(prefix,color, fdr, babelCmdr),\
                delay=babelCmdr.babel_polling_delay)
```

**Comment by SegundoBob** [comment]
Thomas,

Thanks for your improvement.  Is this tweak to your improvement acceptable?

```
    itOut = leoG.IdleTime(lambda
        ito,
        prefix=babelCmdr.babel_prefix_stdout,
        color=babelCmdr.babel_color_stdout,
        fdr=reo, babelCmdr=babelCmdr
        :
            itf(prefix, color, fdr, babelCmdr), delay=babelCmdr.babel_polling_delay)
    itErr = leoG.IdleTime(lambda
        ito,
        prefix=babelCmdr.babel_prefix_stderr,
        color=babelCmdr.babel_color_stderr,
        fdr=ree,
        babelCmdr=babelCmdr
        :
            itf(prefix, color, fdr, babelCmdr), delay=babelCmdr.babel_polling_delay)

```
---
SegundoBob

**Comment by tbpassin** [comment]
Looks good to me.  Don't forget to include the backslashes at the ends of the lines, though. They are essential with the lambda, because the lambda is only allowed to be a one-line expression.  Without the line-continuation symbol you would have multiple lines.

There is an alternative to backslashes in the body of the lambda (the part after the colon).  You can parenthesize the expression, like this:
```
def f(x, y):
    z = lambda x, y: (x, 
                      y)
    return z(x, y)

g.es(f(1, 2))  # prints "(1, 2)"
```
I don't see this helping readability in this case, although it would make clear that you intend to return a tuple.

SegundoBob:
Thomas,

The whole lambda definition is in the parentheses of a function call.   Backslashes are not needed.  I made the changes and ran tests of Leo-Babel before posting my suggestion.  My "git diff" even flagged the changes as "No syntactic changes".

---
SegundoBob

tbpassin:
Thanks, that's good to know.  I forget exactly what my code block was like when I learned that backslashes could be used to continue a line in a lambda, but it wasn't in the parentheses of a function call.

SegundoBob:
Thomas,

You can parenthesize any expression .  A lambda definition is an expression.   I can't think of any Python code where backslashes can't be eliminated by parenthesizing the  expression containing them.

---
SegundoBob

edreamleo (approved):

edreamleo:
@SegundoBob My apologies for the delay in reviewing this PR. I was not aware of it until a few days ago.

edreamleo:
@SegundoBob Rev 1363417 (in devel) adds one line to keep ruff happy.

PR #3810, opened 2024-02-22
PR for Leo 6.7.8

edreamleo:
The PR for Leo 6.7.8.

- [x] Update version in `conf.py`.
- [x] Update `leoPyRef.leo`.
- [x] Add `leoTokens.py`.
- [x] Add `leo/modes/nim.py`.
- [x] Beautify all core files with `leoTokens.py`.
- [x] Simplify all test files.


tbpassin (commented):
FAQ.html lines 8, 36 , 2142- shouldn't this be for 6.7.8, not 6.7.7?
Same for IPythonBridge, lines 36, 193.
Same for acknowledgements.html lines 36, 158.
... and other files!  There seems to be a lot of redundancy here!
OTOH, if version 6.7.7 is intended, never mind!
Hmm, these html files were probably all generated by Sphinx, right?  So the version number changes need to be in the RsT source, not here in the html files.

installing.html - I think the formating is all wrong around lines 162.  It looks like the RsT source should have been using a "::" to create a preformatted block.  The actual html here is nuts.  Also, the backslashes are doubled, which is not what should be shown to the user.  Please review to see how much of this material should be a preformatted block.

Otherwise, I decided not to review all those python files from PyLeoRef.  A quick look seems OK, and there's no way to do a detailled review of all those code changes.

edreamleo:
On Thu, Feb 22, 2024 at 4:04 PM Thomas Passin ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
>
> FAQ.html lines 8, 36 , 2142- shouldn't this be for 6.7.8, not 6.7.7?
>
Thanks for your comments. This is a draft.

I'll fix the version by changing conf.py, per the checklist.

> Otherwise, I decided not to review all those python files from PyLeoRef. A
> quick look seems OK, and there's no way to do a detailled review of all
> those code changes.
>
LeoPyRef.leo is another file that will be updated later. I have reviewed
all the changed files. They all seem correct to me. There are only a few
significant changes. The rest are formatting and routine editing.

Edward

Issue #3811, opened 2024-02-23
UNLs with short file parts should use all open commanders


boltex:
**Background (EKR)**

See PR #3812. Experiments (in test.leo) show the following:

- UNLs with *empty* file parts use `@data unl-path-prefixes` and then all open commanders.
- UNLs with *short* file parts **fail** unless the short file part is in `@data unl-path-prefixes`.
- UNLs with *long (absolute)* file parts fail unless the exact give file exists.

**To do**

- UNLs with *short* file parts should work much like UNLs with *empty* file parts.

**Investigation**

- An improved unit test shows that `g.openUNLFile` works as described above.
- **Legacy** (path-based) UNLs with short file names *do* search all open commanders.
- Oops! **New** (gnx-based) UNLs with short file names search only c.

The **culprit** is `g.findAnyUnl`. The bug is here:
```python
# If there is a file part, search *only* the given commander!
if file_part:
    c2 = g.openUNLFile(c, file_part)
    if not c2:
        return None
    p = g.findGnx(tail, c2)
    return p  # May be None.
```
The proper code is:
```python
# First, search the open commander.
# #3811: Do *not* fail if this search fails.
if file_part:
    c2 = g.openUNLFile(c, file_part)
    if c2:
        p = g.findGnx(tail, c2)
        if p:
            return p
```

tbpassin:
I definitely think that relative UNLs should resolve relative to the directory of the invoking outline.

I also proposed that if a UNL could not be found, then the recent files list should be checked to see if a outline's name could be matched there.  @Edward didn't like that, and I suppose he's probably right because the search locations wouldn't be stable over time.

I am perhaps more sensitive to most about being able to locate files because I run a lot of Linux VMs, and I'm always copying files and directories from, say, my main  Windows machine to one or another of the VMs.  Having unls (or gnxs with paths) that would keep working could be very important, and the only way that could happen easily is with relative files resolving the same way on all the VMs.

boltex:
@edreamleo Thank you ! I'll assign you to this issue instead, and I'll focus on finishing the few issues that remain for LeoJS 1.0 and the next LeoInteg version ! :)

edreamleo:
@boltex You're welcome :-)

boltex:
@edreamleo Thanks for your work on this, but I see that this fixes another bug that was not my intended proposed enhancement. 

[...]

PR #3812, opened 2024-02-23
PR for #3811: handle short unls

edreamleo:
See #3811.

- [x] Improve `test_g_openUNLFile` to show that `g.openUNLFile` handles short file names correctly.
- [x] Disable `@data unl-path-prefixes` in `test.leo` to simplify testing.
- [x] Fix `g.findAnyUnl` as described in [#3811](https://github.com/leo-editor/leo-editor/issues/3811).

### DIFF
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -6717,8 +6717,17 @@ def run_unit_tests(tests: str = None, verbose: bool = False) -> None:
 #    For example, the link: `unl:gnx://#ekr.20031218072017.2406` refers to this
 #    outline's "Code" node. Try it. The link works in this outline.
 #
-#    *Note*: `{outline}` is optional. It can be an absolute path name or a relative
-#    path name resolved using `@data unl-path-prefixes`.
+#    *Note*: `{outline}` can be:
+#
+#    - An absolute path to a .leo file.
+#      The link fails unless the given file exits.
+#
+#    - A relative path to a .leo file.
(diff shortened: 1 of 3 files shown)


edreamleo:
@boltex @tbpassin This issue is ready for review.

Please read carefully the comments in @3811 :-)

tbpassin (approved):

boltex (approved):
This indeed fixes the bug that short UNL didn't find their targets **in already opened files**. It's also clean code that works, so I'll gladly approve this pull request...

But my intended enhancement was that short UNLs could find **and open** outlines if the short UNL could be resolved relative to the invoking commander's current path. without needing an entry in @data unl-path-prefixes (see my added comment in https://github.com/leo-editor/leo-editor/issues/3811 for precise explanation)

tbpassin:
I fully agree that ctrl-clicking (or invoking the underlying command)  UNL that resolves to an outline in the directory of the invoking outline should open that outline and navigate to the node.

When I reviewed the changes, I assumed that this line would both open the outline and navigate to the right position:
`p = g.findGnx(tail, c2)`

If that is not correct then this PR needs more work.

edreamleo:
@boltex @tbpassin I have merged this PR because it fixes #3811.

I have created #3814 for discussion of Félix's additional proposal.

As indicated in that issue, I am not interested in the proposed enhancement.
Imo, there is no use case that would justify changing Leo's existing behavior.

Issue #3814, opened 2024-02-25
Resolve UNLs with short paths relative to c


edreamleo:
See PR #3815. In the discussion of PR #3812, @boltex said:

[My] intended enhancement was that short UNLs could find and *open* outlines if the short UNL could be resolved relative to the invoking commander's current path, without needing an entry in `@data unl-path-prefixes`.

**Discussion (EKR)**

On second reading, I agree that this proposal makes sense.  It involves a change only to `g.openUNLFile`.

**My mistaken first impression**

My first (erroneous) thought was that this issue suggested that Leo should support **relative paths** in UNLs.
These paths would be resolved using the directory containing `c`, the presently selected outline.

This would be a significant change to Leo's UNLs. It would affect Leo's gui and Leo's UNLs.
Imo, this proposal is not worth doing for the following reasons:

At present, legacy (path-based) and new (gnx-based) UNLs can have the following kinds of file parts:

- Empty. Resolves to `c`, the presently selected outline.
- Short. Resolves using `c`, the outlines listed in `@data unl-path-prefixes`, and all other open outlines.
- Long (absolute). Resolves *only* the the given outline.

Imo, this existing behaviour is adequate to cover all reasonable use cases.
In particular, sessions are an easy way to open any outlines needed to resolve UNLs with short file parts.

This proposal's **fatal flaw** it that creates a *less convenient* work flow for users.

Indeed, Leo *never* generates relative file parts because *all UNLs in Leo's status area refer to c*.
In other words, users would have to create relative file parts **by hand**.

Yes, we could imagine new (complex, never used) machinery for specifying relative file parts, but that would be pure featuritis.

boltex:
@edreamleo @tbpassin **Thank you dear friends! IT WORKS!** 🥳 🎺 🎵 Hurray!!

Now, I'm off to transliterate all that UNL good stuff to LeoJS!!  (And finish my second video tutorial, where the opening scene is me using a UNL in a 'table of contents Leo document' to open another Leo file that has some demonstrations of using the VSCode UI from within leo scripts! - point being to obviously also demonstrating the existence and usage of UNLs among all the other features I present in the video! )

[...]

PR #3815, opened 2024-02-25
PR for #3814: Better defaults for UNLs with short file names

edreamleo:
See #3814.

- [x] Rewrite `g.openUNLFile`.
    Require exact matches for absolute file paths or paths found in `@data unl-path-prefixes`.
    Otherwise, resolve paths using the `c's` directory, but see the code for the details.
- [x] Revise unit tests accordingly, moving no-long-used code to the attic.
- [x] In `test.leo`, activate entries `@data unl-path-prefixes`.

### DIFF
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -7319,14 +7319,19 @@ def getUNLFilePart(s: str) -> str:
 def openUNLFile(c: Cmdr, s: str) -> Cmdr:
     """
     Open the commander for filename s, the file part of an unl.
-
-    Use `@data unl-path-prefixes` to convert to relative to absolute paths.
-
     Return None if the file can not be found.
     """
+    # Aliases.
     base = os.path.basename
+    dirname = os.path.dirname
+    exists = os.path.exists
+    isabs = os.path.isabs
+    join = g.os_path_finalize_join  # Not os.path.join
     norm = os.path.normpath
+
+    # c's name and directory.
     c_name = c.fileName()
+    c_dir = dirname(c_name)
 
     def standard(path: str) -> str:
         """Standardize the path for easy comparison."""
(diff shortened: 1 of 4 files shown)


edreamleo:
@boltex @tbpassin This PR is ready for review:

- Revised unit tests pass.
- More importantly, all the by-hand tests in `test.leo` work as expected.

tbpassin (commented):
test.leo lines 1479, 1480 - These paths make the test unable to be run on someone else's machine.

Otherwise, the code changes look OK.

boltex (approved):
I'm smiling ear-to-ear right now!

tbpassin:
I just tried this branch out (I didn't do that before my code review) and I think it's not quite there yet.  Neither a _unl:gnx_-style or a path-based unl that do not include a node opens to the root of the outline.  I think they should, and that is how historically UNLs were handled, as best as I remember.

Example for _flexx_test.leo_ in same outline as _workbook.leo_ and the UNLs to click on are in the workbook (the referenced nodes exist in the actual outline):

`unl:gnx://flexx_test.leo#TomP.20200402105706.1`   (works)
`unl:gnx://flexx_test.leo`   (does not open or navigate to the outline)
`unl://flexx_test.leo#Demo-->@file flexx-demo.py`  (works)
`unl://flexx_test.leo`  (does not work)

boltex:
@tbpassin @edreamleo Thomas makes a good point (I never thought of UNLs without the node part before, hehe! )

But now considering this possibility, I also acknowledge that in the case an UNL without the 'node' part, (i.e. that simply pointing to a Leo file that exists, without the '#' suffix part... ), it should simply open it. (if the file already exist). 

**(As always: If this is not trivial & simple to code, then it should have it's own separate issue & merge request)**

Either way, I want to thank again both of you for the attention you are porting to the (very cool and useful) UNL related features! 😄

edreamleo:
@tbpassin @boltex Thanks for your comments. I am going to merge this PR as it is.

Nothing about UNLs is easy to implement, so any further requests/suggestions should have their own separate issues.

Some specific comments:

- Afaik, Leo has never opened UNLs w/o a node specifier. Maybe it has, but that would be news to me.
- I don't understand why creating relative paths *by hand* is in any way preferable to using the `@data` node.
- `test.leo` is a *semi-official* test file that I use for by-hand testing.
  Imo, it's fine for `@data` node to refer to paths on my machine. No unit tests use `test.leo`.

tbpassin:
> * I don't understand why creating relative paths _by hand_ is preferable to using the `@data` node.

Because
1. You have to remember about the setting and the right setting name;
2. You have to open the settings file, navigate to the right place, and add a new line with the new path;
3. Then copy the UNL into the node where you want to place the navigation link. (even though that UNL may not be on the clipboard at this point);
4. You (may) have to restart  Leo for the new setting to take effect.

Whereas what you really want to do is:
1. Copy the node UNL either from the status line or using the minibuffer command (I have the command in a custom local menu) - or by just typing the unl as "unl://&lt;filename>";
2. Paste it into the node where you want to place the navigation link.

IOW, using the `@data` node is a lot of awkward boiler plate that is not closely related to what you want to do.  Right now you may think that it's easy, but wait until you haven't thought about it in two or three months and then try to figure out why your new links aren't working.

In addition, if you want to share your outlines with someone else, having to also share the right `@data` settings will be easy to forget, making the sharing activity harder to get working.

One way I'd like to use these new UNL links is to point into places in the documentation outlines, like leoDocs.leo or the cheatsheet.  Preferably I'd like those links to work automatically.  But I wouldn't mind using the `@data` node for those as much because they would probably be fairly static and permanent.

So yes, let's discuss this further by using a new issue in the future.

boltex:
@tbpassin @edreamleo 
Heh! I agree that nothing about UNLs is easy to implement! and that further requests/suggestions should have their own separate issues.

> I don't understand why creating relative paths by hand is preferable to using the @data node.

In somewhat the same vein as @tbpassin point of view: from my point of view, using the ` @data node` is an advanced, optional feature. Where as copying and using short relative UNLs directly, (_or by also needing to 'craft' it by inserting a relative foldername+slash, at the start of the file-part_) is easy and intuitive. 

Most importantly, this would allow to share a folder project, or a repository, via zip, github, etc, to another person to collaborate.  (that other person opens it on its own hard drive, where no absolute file path would work, neither from `@data node` setting nor full-path UNLs... Whereas relative, short UNLs, would be easily and out-of-the-box navigable experience for that other user.)

Crafting an UNL to be relative would (most of the time) be trivial: it's grabbing from the status bar, and adding a relative folder name and a slash in front of the file path to make it relative.

 (for instance adding "docs/" or "../docs/" as prefix to the file-path of the UNL to re-use @tbpassin 's example.)

I also would think overall, that users don't know about the settings and the `@data` type of setting, etc... But since UNLs look and intuitively feel like URLs on a webpage, most people, experienced leonistas and newbies alike, would expect the default behavior of short UNLs paths to be resolved relative to the commander's outline file path. _(still allowing for settings like `@data unl-path-prefixes` to also be used as possible paths to look into)_

Like Thomas said, let's discuss this further by using a new issue in the future! 😄 

**Regardless, I just want to underline that I'm very grateful for the added behavior that #3814 adds to the handling of UNLs being clicked! Thanks again!**

Félix

edreamleo:
@boltex @tbpassin When I awoke this morning I saw why I was having trouble understanding you both.

I have just created #3816.

The `@data` setting works (and saves work) in some situations involving (say) writing a book. The entries in the dict correspond `.leo` files holding parts of the book. In such situations, the initial work of setting up the `@data` node pays off.

But in *other* situations relative paths (and omitted node specifiers) will be amazingly useful!
We are going to say to ourselves, "how did we ever live without these features?"

Issue #3816, opened 2024-02-26
Enhance UNLs with relative paths and optional node specifiers


edreamleo:
See PR #3817.

**The proposal**

- Allow **relative paths** in UNLs.
- Allow **file-only** UNLs by making the node specifier optional.

Both features would be important additions to Leo:

- Relative paths will allow people to share `.leo` files more effectively.
- File-only UNLs are nifty way to refer to other outlines!

**Notes**

- **UNLs** with relative paths are *completely different* from `file://` **URLs**.
- Users will have to create UNLs with relative paths by hand.
  Leo's status area can never create such UNLs.

tbpassin:
I think that most or all of the work can take place in `g.findAnyUnl()`. There will be two slightly tricky points:
1. Checking a relative UNL whose outline is not open without visibly opening it.  We don't want the user to suddenly see a lot of opened outlines he didn't expect;
2. If there is no node specifier, opening the outline and selecting its root node.  This should be easier than (1).

There is code that could be used as a model for 1) already, since Leo opens an outline without making it visible when first looking for its setup nodes, if I'm remembering that correctly.

Code for 2) should be found in the method that creates a new outline.

I also suggest that when a Unl with a node specifier turns out to point to a non-existent node, that it be treated as if there were no node specifier.  IOW, it would open the right outline anyway.  Currently no outline is opened and an error message is printed to the terminal.

edreamleo:
Completed via PR #3817.

PR #3817, opened 2024-02-26
PR for #3816: improved UNLs

edreamleo:
See #3816. This PR was easier to do than I imagined. It makes minor, easily understood, changes in the expected places, namely `g.findAnyUnl`,  `g.openUNLFile` and the regex in `g.getUNLFilePart`.

- [x] Add easy special case to `g.findAnyUnl`.
- [x] Make the gnx part optional in the `file_part_pattern` used by `g.getUNLFilePart`.
- [x] Add easy special case to `g.openUNLFile` for file parts containing `os.sep`.
   In addition, improve the `standard` function.
- [x] Add new test cases to `test.leo`. Everything appears to work as expected.
   All unit tests pass.
- [x] Update the documentation in the section `<< About clickable links >>`.

The new code uses `@data unl-path-prefixes` only for *short* (not relative) paths.
Imo, this behavior is both reasonable and compatible with all existing UNLs.

### DIFF
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -6740,8 +6740,11 @@ def run_unit_tests(tests: str = None, verbose: bool = False) -> None:
 #
 #    This link works: `unl://#Code-->About this file`.
 #
-#    *Note*: `{outline}` is optional. It can be an absolute path name or a relative
-#    path name resolved using `@data unl-path-prefixes`.
+#    As in point 2 above, `{outline}` or `{gnx}` may be empty, but at least
+#    one must exist.
+#
+#    `{outline}` may be an absolute path name or a *short* name resolved
+#    using `@data unl-path-prefixes`.
 #
 # 4. Web URLs: file, ftp, gopher, http, https, mailto, news, nntp, prospero, telnet, wais.
 #
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -6800,6 +6800,11 @@ def findAnyUnl(unl_s: str, c: Cmdr) -> Optional[Position]:
         file_part = g.getUNLFilePart(unl)
         tail = unl[3 + len(file_part) :]  # 3: Skip the '//' and '#'
 
+        # #3816: Just open the file if there is no tail.
+        if not tail:
+            c2 = g.openUNLFile(c, file_part)
+            return c2.p if c2 else None
+
         # First, search the open commander.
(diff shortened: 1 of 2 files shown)


tbpassin (changes requested):
This PR does open the file containing a relative UNL in the same directory if it includes a node part (gnx or path-based) but it does not open the outline if there is no node part.

Without a node specifier, this new code (line 6803ff in leoGlobals.py) is not called:
```
        if not tail:
            c2 = g.openUNLFile(c, file_part)
            return c2.p if c2 else None
```
This is probably because the tail part is not empty as expected (and it's also missing its first character.  I.e., "flexx_test.leo" becomes "lexx_test.leo".

tbpassin:
Aha, I hadn't realized a plain file name needs to have a '#' character appended.  I looked at the regexes and then I realized it.  So now both these forms work for me:

`unl:gnx://flexx_test.leo#`
`unl://flexx_test.leo#`

For both, the target outline opens if it isn't open already.
Also, an outline one level up opened with

`unl://../<filename>`

Based on this testing, I think this PR is ready.

edreamleo:
@tbpassin Thanks for your comments. I'll merge this PR after updating some comments.

boltex (approved):
I've also tested with `../myFile.leo` to see if the relative path work with `../` to go 'UP' a folder. Even `../anotherSideFolder/otherLeoFile.leo` worked to go up a folder, and back down into ANOTHER folder. Impressive. 

UNLs Works as expected -> resolving relative chunks of paths just like a regular web URL! 

Thanks for this feature! This rocks! 😄

boltex:
@edreamleo @tbpassin To celebrate this achievement for the UNLs to honor any relative path given, I thought you'd like to see that once converted line-by-line into javascript, it even works (relative paths that go up with `../../' and back down into other directories) in vscode on the web with LeoJS ! (in the next beta version coming out in less than a week)

Screenshot as proof below:
image

tbpassin:
Great going, @boltex! 

I think of these cross-outline navigable links as letting us easily create webs of documents and navigate around them, without web servers or needing to run them through a system like Sphinx.

I mentioned in an earlier comment (maybe on another comment thread) wanting to link to places in documents like _LeoDocs.py_. I have been learning that many times it's easier for me to find something in the outline, using the Nav pane, than in the rendered Sphinx set of HTML pages.  If I want a rendered view of a section, I can always view it in VR3. Now I will be able to link to those places easily.

PR #3825, opened 2024-03-02
Fixed set_name_and_title to reset the db 'key' used for its entries when 'save-as' changes the filename.

### LINKED ISSUE #3822: "save-as" does not change it's db key for its entries, so db related data still saved in old name's key in db when saving, (until reopened.)
To reproduce: open a .leo file, change it's window size and dimension in a notable way, set some node expanded, some opened, and save file.

The settings about the window position and size, along the state of opened/closed nodes is saved in db. (you can test this works so far as intended by closing leo and double clicking on the file in window's explorer to open that single file.)

So far so good. reopen the file, perform a save-as command to save it as another filename. From this point, changing window size and position along with node expanded state are still saved in db with the original filename's keys for db indexing.  

It should be done with the new given path+filename when the save-as was performed instead.

(as stated at the end of this issue's title: closing and re-opening the file will make the saving in db work fine for all normal settings such as position, size, node's states, etc.)

boltex:
Fixes #3822

I've tested it manually, and ran unit tests: all pass.

### DIFF
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -39,7 +39,6 @@ def set_name_and_title(c: Cmdr, fileName: str) -> str:
 
     Return the finalized name.
     """
-    oldFilename = c.mFileName
     # Finalize fileName.
     if fileName.endswith(('.leo', '.db', '.leojs')):
         c.mFileName = fileName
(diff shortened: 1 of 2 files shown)


boltex:
This could also have been fixed by removing key from CommanderWrapper entirely, and just using its self.c.mFilename instead of its self.key property... 

I'm going to change this pull request for this more elegant solution.

edreamleo (approved):

PR #3827, opened 2024-03-03
Boltex/issue3826

### LINKED ISSUE #3826: Make sure .leojs file format has no state bits in file, and instead all saved in db.
**Mirror of LeoJS's issue https://github.com/boltex/leojs/issues/73** -> When saving to a .leojs file, the expanded, marked, etc. bits should only be put in the db, like for standard .xml leo file.

Meaning that they simply should be saved in, and pulled from, the db. Just like for regular .leo Files.

boltex:
Make sure .leojs file format has no state bits in file, and instead all saved in db.  Making the .leojs JSON file format a pure equivalent of the .leo xml file format.

Fixes #3826

I also added a 'getrecentfileslist' method to the server.

I've tested manually and ran unit test to make sure they all pass.

### DIFF
--- a/leo/core/leoFileCommands.py
+++ b/leo/core/leoFileCommands.py
@@ -546,7 +546,6 @@ class FastRead:
 
                     v._headString = v_dict.get('vh', '')
                     v._bodyString = gnx2body.get(gnx, '')
-                    v.statusBits = v_dict.get('status', 0)  # Needed ?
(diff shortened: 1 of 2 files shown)


edreamleo (approved):
The changes involve only code that you wrote, namely `scanJsonVnodes` and `leojs_vnode`, so I don't feel qualified to comment :-)

PR #2883, opened 2022-09-20
PR for #2882: require qt6.1+

edreamleo:
See #2882.

Closed in favor of PR #3828. Merging devel into this PR proved difficult/impossible.

**Summary**

Alas, more specific annotations are not very useful. 

- [x] Use contents of leoQt6.py as basis of leoQt.py.
- [x] Delete leoQt5.py and leoQt6.py.
- [x] Remove all isQt5 and isQt6 constants.
- [x] Use code formerly enabled by isQt6 constant.
- [x] Fix pylint complaints. Run pylint as follows:
    `pylint leo --extension-pkg-allow-list=PyQt6.QtCore,PyQt6.QtGui,PyQt6.QtWidgets`
- [x] **Double check**: Change plugins.attrib_edit, plugins.bookmarks, and plugins.todo.
- [x] (Won't do now) Fix mypy complaints.
    Fixing mypy complaints did reveal some potential (minor) bugs,
    but most annotation "fixes" were more lax annotations.
    Imo, it would be better to fix annotations when we are actually going to use the code!

PR #3828, opened 2024-03-03
PR 2 fixes #2882: Qt6 only

### LINKED ISSUE #2882: Require Qt6.6+
See PR #2883. Closed  via #3828.

Supporting only Qt6 allows better mypy annotations.

edreamleo:
See #2882. Replaces PR #2883.

This PR focuses on removing Qt5 code while keeping mypy happy.

A *later* PR will attempt to strengthen the annotations for Qt widgets and Leo wrappers.

- [x] Use contents of leoQt6.py as basis of leoQt.py.
- [x] All unit tests pass.
- [x] Fix pylint complaints. Run pylint as follows:
    `pylint leo --extension-pkg-allow-list=PyQt6.QtCore,PyQt6.QtGui,PyQt6.QtWidgets`
- [x] Delete leoQt5.py and leoQt6.py.
- [x] Remove all isQt5 and isQt6 constants.
- [x] Replace `x.exec_()` by `x.exec`.
- [x] Delete Qt5-only modules: phonon, QtDeclarative, QtWebKit, and QtWebKitWidgets.
- [x] Retire five of Terry Brown's plugins. There is no obvious way to make these plugins work with Qt6:
   `notebook.py`, `richtext.py`, and three "editpane" plugins: `pandownview.py`, `webengineview.py`, and webkitview.py.\
   These plugins are no great loss. They do not work with the typical installation of Leo 6.7.8.

**Real bugs**

This PR fixes the following real bugs:

- `LeoEditPane`: change `update` to `update_flag` to not shadow `QWidget.update`!
- `eventFilter` method in `qt_text.py`: Add self argument.
- `paintBlock` method in `qt_text.py': Use int values instead of float.


tbpassin (commented):
LeoPrinting.py, line 294:
`doc.print_(dialog.printer())`
Is `print_()` available in PyQt6?  Or is it a Leo method?

freewin.py,  deleted lines 256ff:
```
    from OpenGL import GL
    assert GL  # To keep pyflakes happy.
```
Do we know that we don't need to import GL in Ubuntu anymore for PyQt6?

test_gui.py, line 70:
`        # Skip this test when using PyQt5.`
Isn't this comment superfluous now?

edreamleo:
@tbpassin Many thanks for your careful review. I'll deal with all the issues you before merging this PR.

edreamleo:
> LeoPrinting.py, line 294: `doc.print_(dialog.printer())`

Good catch: `doc.print` is correct in Qt6.

> freewin.py, deleted lines 256ff: `from OpenGL import GL`

A cff shows that neither OpenGL nor GL appears anywhere in Leo. I see no reason to worry :-)

> test_gui.py, line 70: ` # Skip this test when using PyQt5.` Isn't this comment superfluous now?

Yes. I'll remove the comment and the try/except block.

PR #3847, opened 2024-03-26
PR: Move leoAst.py commands and test classes to the attic

edreamleo:
This PR fixes a serious bug in leoAst.py by deleting all its commands and test classes.

I will merge this PR only as a last resort.


edreamleo:
I am going to close this PR. The proper course of action is to fix the bug!

Issue #3845, opened 2024-03-25
Improve mypy annotations


edreamleo:
It's time to significantly strengthen Leo's mypy annotations:

- PR [#3828](https://github.com/leo-editor/leo-editor/pull/3828) converted Leo's codebase to Qt6.6+,
  but the PR weakened the mypy annotations.
- PR [#3844](https://github.com/leo-editor/leo-editor/pull/3844) added `g.isUniqueClass`, the perfect tool for validating annotations.

edreamleo:
I am closing this issue. No further work on annotations is planned.

PR #3850, opened 2024-03-26
PR for #3845: Improve mypy annotations

edreamleo:
See #3845.

**Phase one: Disambiguate Event annotations**

This has been a spectacular success. These changes affect the files in `leo/core` and `leo/commands`.

Define QEvent and LeoKeyEvent as follows:
```python
from leo.core.leoGui import LeoKeyEvent
QEvent: TypeAlias = QtCore.QEvent
```
- Change Event to LeoKeyEvent for Leo commands.
- Change Event to QEvent everywhere else in Qt gui code.
- (Maybe) Use LeoKeyEvent instead of KeyEvent in cursesGui2.py.

**Phase two:  Disambiguate Widget annotations**

A partial success. These changes mostly affect the Qt gui files.

- Some minor annotations remain `Any`. It's not worth futzing with them.
- mypy has trouble annotating Leo's wrapper classes.
  Apparently, mypy doesn't know enough to accept `Union` annotations.
  In such cases I have changed `Any` to `Wrapper` (defined as `Any`).
  I am **not** going to change Leo's wrapper classes to accommodate mypy!!!

**Phase three: Remove mypy and pylint suppressions where possible.**

Many `type:ignore` suppressions have been removed. Many still remain.
Many pylint suppressions are no longer necessary.

**Bug fixed**

Changing code that *isn't* an annotation can prevent Leo from starting!

However, better annotations revealed real (or potential) bugs:

- In various Gui base classes, change the `str` annotation to `Any`.
- `checkerCommands.py: check_nodes` and `check`: The ctor for the `CheckNodes` class should have a `c` args.
- `VimCommands.shell_command`: Apparently, this method should pass a `LeoKeyEvent` to the `vc.do` method.


edreamleo:
@boltex @tbpassin This PR is ready for review. I'll wait for your approvals :-)

tbpassin (commented):
1. I notice that some usages of the Event/LeoKeyEvent argument have defaults and some do not, e.g:
`def insertIcon(self, event: LeoKeyEvent = None) -> None:`
`def deleteIndentation(self, event: LeoKeyEvent) -> None:`

I wonder if they should all default to None, or if any of these cases should *not* default to None.

2. leoGui.py. lines 19ff: I don't see LeoKeyEvent getting imported, though it is used later.
3. contextmenu.py, lines 54ff - same comment.
4. pyplot_backend.py, lines 52-53. We don't need these comments about qt5 any more:
```
# matplotlib.backends.backend_qt5.FigureManager probably does exist. See:
# https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/backends/backend_qt5.py
```

edreamleo:
@tbpassin

Point 1: Excellent catch!  All of Leo's commands (with `@g.command` or other decorators) *should* take a single (positional) argument `LeoKeyEvent`. All other events are *gui-related* events.  In effect, the `LeoKeyEvent` annotation is the "knife" that separates Leo's de-factor core from Leo's gui code.

Let's look at some cffs:
- A cff of `event: LeoKeyEvent` yields 715 matches. 
- A cff of `event: LeoKeyEvent =` yields 398 matches.
- A cff of `event: LeoKeyEvent,` yields 79 matches.
- A cff of `event: LeoKeyEvent)` yields 266 matches.

These number need not "add up" because not all annotations corresponds to `def` statements.

However, these cffs do indicate that Leo is not consistent in its *annotations*. That by itself is not too serious.

But there are also inconsistencies in how methods (or functions) *use* the event arg:

- Many methods use `self.c` instead of `c = event.get('c')`.
- Most other methods use this pattern:
```python
c = event.get('c')
if not c:
    return
```
- Some other methods just pass the event arg to a *gui-related* helper.
  This seems dubious as far as annotations go.

**Summary**

@tbpassin Thanks for your important comment.

Substantial work will be required to regularize the code with its annotations.

I *might* do some of that work in this PR, but I'll probably leave this issue to another PR. I definitely will *not* make widespread code changes in *this* PR merely on account of the annotation of the `event` arg.

PR #3857, opened 2024-04-09
Improve annotations in leoGlobals.py

edreamleo:
This PR is a milestone in Leo's history. All annotations in `leoGlobals.py` are as accurate as they can be.

The only minor exception is the annotation of `*args` and `**kwargs`. Imo, the typing module should provide help.

- [x] Annotate `g.app` as `LeoApp`. 
    **Breakthrough**: This annotation allowed mypy to check all other Leo files more thoroughly!
- [x] Remove evil `multiple` kwarg from `runOpenFileDialog`. Add `runOpenFilesDialog`.
    This is a (minor) breaking change to Leo's API.
- [x] Fix all resulting mypy complaints.

**Significantly improved annotations**

- `leoApp.py`: Improve annotations of `g.app.db` and `g.app.global_cacher`.
- `leoConfig.py`: Improve annotations of ivars and methods related to buttons and commands.
- `leoCache.py`: Annotate `GlobalCacher.db` as `Union[dict, SqlitePickleShare]`.
- `commands/editCommands.py: dump_caches`: Add a guard.
- `leoGlobals.py`:  It is *thrilling* to be able to disambiguate the following:
  - Use `LeoKeyEvent`, `QEvent`, `QMouseEvent`, `LeoFrame`, `QFrame` as `QWidget` as appropriate.
  - Use `IO`,  `Module` and `Union` annotations as appropriate.
 
**Bugs found and fixed**

Most of the diffs involve only annotations.  Here are real bugs:

- `leoApp.py: LeoApp.finishQuit`: Add a guard.
- `checkerCommands.py: MypyCommand.check_file`: pass a string, not a list, to `bpm.start_process`.
- `spellCommands.py: DefaultWrapper. __init__`: Set `g.app.spellDict = DefaultDict().d`.
- `leo.plugins.leo_babel.tests: test.py`: Don't patch `g.IdleTime`.

### DIFF
--- a/leo/commands/checkerCommands.py
+++ b/leo/commands/checkerCommands.py
@@ -435,7 +435,7 @@ class MypyCommand:
         if not mypy:
             print('install mypy with `pip install mypy`')
             return
-        command = f"{sys.executable} -m mypy {fn}".split()
+        command = f"{sys.executable} -m mypy {fn}"
         bpm = g.app.backgroundProcessManager
         bpm.start_process(c, command, fn=fn, kind='mypy')
     #@+node:ekr.20210302111935.7: *3* mypy.run (entry)
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -151,11 +151,11 @@ def importAnyFile(self: Self, event: LeoKeyEvent = None) -> None:
         ("Python files", "*.py"),
         ("Text files", "*.txt"),
     ]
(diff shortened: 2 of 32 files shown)


edreamleo:
@boltex @tbpassin I'm going to merge this, but please do review and test devel.

I've tested Leo's spell checking with the new code, which was important because of the major change to g.app.spellDict.

I've also tested Leo's caching. That's not so important because only an annotation changed. However, this changed annotation might be useful to LeoJS.

Issue #3871, opened 2024-04-19
Searching for the definition should not alter the user's search options


edreamleo:
See PR #3872. This is an irritating hangnail.

boltex:
@edreamleo Leo and LeoInteg codebases are structured enough that it took me only about an hour to implement this in LeoInteg with https://github.com/boltex/leointeg/commit/0480148115d291fe1c812b156a1a151b3b1f5c56 , and in leoserver.py with https://github.com/leo-editor/leo-editor/pull/3943 

🦁  Leo Rules!

PR #3872, opened 2024-04-19
PR for #3871: searching for definitions should not change search settings

edreamleo:
See #3871.

**Breaking change to Leo's API**:  This PR changes the signatures of `find.find_def` and `find.find_var`.

- [x] Rewrite `find-def` and `find-var` commands. These commands are now synonymous.
- [x] **Aha**: Don't use Leo's find commands.  Search only body text,  honoring `@nosearch`.
- [x] As before, for unique matches, select the node that matches.
  `do_find_def` must honor `@bool reverse-find-defs`.
- [x] Set the find text if there are multiple matches. This is convenient and should never be annoying.
- [x] Support `@bool prefer-nav-pane`, with default set to `True` to match most Leonistas' preferences.
  - True: Show *multiple* search results in the `Nav` pane if the `quicksearch` plugin is enabled.
  - False: Always show *multiple* matches using clones as in the `clone-find` commands.
  Note: regardless of this setting, Leo selects *unique* search results by selecting the node.

**Minor changes**

- [x] Init `c.quicksearch_controller = None`.
  `QuickSearchController.__init__` sets `c.quicksearch_controller = self`.
- [x] Improve unit tests.


tbpassin (commented):
Line 890 if leoFind.py - I assume that at some point the `if 1:` will be replaced by a real test.

Why can't we just save the old settings and replace them after the search?  Notionally -
```
old_settings = settings
def_settings = calculate_def_settings()
result = self.do_find_var(def_settings)
ftm.set_widgets_from_dict(old_settings)
```

edreamleo:
> Why can't we just save the old settings and replace them after the search? Notionally -

Sure, the code can do that, and that's what this PR does.  However, if the the search *required* (say), the 'ignore-case' option, then the *next* search won't find what the user is looking for.

Otoh, ctrl-clicking the *actual* name of a class or def will find the name *without* the need for 'ignore-case'. The problem comes when the user doesn't remember the real name.

As I write this, perhaps the solution is to force the user to specify 'ignore-case' (somehow) when they want a broader search.

Thanks for your comment. It has helped.

edreamleo:
@tbpassin I forgot to ping you. See the above comment.

tbpassin:
I didn't even realize that there is a way to search for a def or variable name without double-clicking on it in the body.  Now that I've looked, I see them in the tab-completion list. I probably didn't fully realize what the issue is that the PR tries to solve.

Used as a minibuffer command, I'm not sure I understand why there is an issue about _ignore-case_.  In fact, I don't see why the command should be using _ignore-case_ at all.  It picks up the word under the cursor to search for, and that word in the code should be capitalized correctly already, isn't it?  Or is the expected use that the user will type a name but he doesn't remember how or if it's capitalized, and run a `find_def` command on that?

I think this is minor - I mean changing the case setting - in comparison with the command finding the wrong definition line when there is more than one def by that name in the outline.  When this happens, I don't know how to keep looking for the next one.

edreamleo:

As you imply, the way forward may be to treat the minibuffer command a bit differently than the Ctrl-click version.

I'll say more after I've run some experiments.

edreamleo:

I am thrilled with the result. Control-clicks now take noticeably longer, but the new code is much better.

edreamleo:
@boltex This PR will probably affect both LeoInteg and LeoJS.

boltex:
@edreamleo When reading about this, I also thought about suggesting re-using the nav pane! If you make results use the same objects-structure as nav pane clickable results, I wont even have to modify LeoInteg! :) (it mirrors those nav pane results in it's 'goto' pane)

 I'll review and try it out as soon as I finish debugging and polishing 'detached body panes' in leoJS and LeoInteg, in a day or two.

edreamleo:
@boltex Recent revs put the results in the `Nav` if and *only* if the `quicksearch` plugin is enabled.

Will that plan be good enough for you?

edreamleo:
@tbpassin @boltex Recent revs contain a significant optimization. The code tries to find possible alternate spellings only if searches for the primary spelling (the string that is control-clicked) fail.

edreamleo:
@boltex @tbpassin Please approve this PR. Imo, `@bool prefer-nav-pane` is the last piece of the puzzle.

tbpassin:
I just checked out this branch to try before reviewing.  When Leo started I got this error message (Leo seems to be working all right):
```
  File "c:\Tom\git\leo-editor\leo\core\leoConfig.py", line 2109, in visitNode
    return f(p, kind, name, val)  # type:ignore
           ^^^^^^^^^^^^^^^^^^^^^

  File "c:\Tom\git\leo-editor\leo\core\leoConfig.py", line 404, in doMenuat
    ans = self.patchMenuTree(mlist, targetPath)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "c:\Tom\git\leo-editor\leo\core\leoConfig.py", line 478, in patchMenuTree
    ans = self.patchMenuTree(val, targetPath, path=path + '/' + name)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "c:\Tom\git\leo-editor\leo\core\leoConfig.py", line 478, in patchMenuTree
    ans = self.patchMenuTree(val, targetPath, path=path + '/' + name)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "c:\Tom\git\leo-editor\leo\core\leoConfig.py", line 467, in patchMenuTree
    kind, val, val2 = z
    ^^^^^^^^^^^^^^^

ValueError: not enough values to unpack (expected 3, got 1)
```
```
Leo 6.7.9-devel, ekr-3871-search-settings branch, build bfec235001
2024-05-27 05:06:16 -0500
```

edreamleo:

The crashes you report have nothing to do with this issue. You may file a separate issue if you like.

tbpassin:
> The crashes you report have nothing to do with this issue

: ) I knew that but mentioned it in case you recognized having made a change recently.

edreamleo:
@tbpassin I haven't touched any menu code in awhile.

My guess: your issue might be a lapse in detecting user errors.

tbpassin (approved):
## leoFind.py
#### Lines 572 ff   `       # Look for alternate matches only if there are no exact matches.`

I wonder if this is worth the extra time it takes when there is no match. Of course, no-match shouldn't occur very often, so maybe it's a moot point.

#### Lines 627 ff -
```
        # The Nav pane can show only one match, so issue a warning.
        if len(unique_matches) > 1:
            g.es_print(f"Multiple matches for {word}", color='red')
            for z in unique_matches[1:]:
                g.es_print(z)
        # Put the first match in the Nav pane's edit widget and update.
        x.clear()
        e.setText(unique_matches[0])
        c.frame.log.selectTab('Nav')
```
I suppose this is to handle the case when more than one kind of hit happens (e.g., def vs Class), is that right?  I can understand why the code is in here, but it's going to be weird to have the log pane flash the message and then the view will switch right back to the Nav pane.  I don't have a solution to offer right  now but I think we should keep trying to find one.

Otherwise I didn't notice anything to comment on.  Very nice work! I'm elated to see the results in the Nav pane.  The very first thing I tried is to CTRL-click on `c.free_layout.get_top_splitter()`.  The two places I knew about from earlier showed up  in the Nav pane as desired.

edreamleo:

Not searching for alternates makes a noticeable difference.

Issue #3873, opened 2024-04-21
@language isn't inherited in .txt files


edreamleo:
See PR #3877.

Within `prototype: leoTokens.rs` in the attic (in a branch) `@language` rust does not apply to its descendants.

PR #3877, opened 2024-04-22
PR for #3873: @language isn't inherited properly in .txt files

edreamleo:
See #3873.

- [x] Rewrite `g. getLanguageFromAncestorAtFileNode` to use four passes.
  The first two passes search body text for unambiguous `@language` directives.
  The second two search for `@<file> nodes` and use the file's extension.

In practice the second two passes will rarely happen. The new code will be as fast as the old.

Imo, the new search order will never cause problems. It certainly fixes the original issue.

### DIFF
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -2652,7 +2652,7 @@ def getLanguageFromAncestorAtFileNode(p: Position) -> Optional[str]:
         if len(languages) == 1:  # An unambiguous language
             return languages[0]
 
-    # Passes 3 & 4: Use file extension in @<file> nodes.
+    # Passes 3 & 4: Use the file extension in @<file> nodes.
 
     def get_language_from_headline(v: VNode) -> Optional[str]:
         """Return the extension for @<file> nodes."""


tbpassin (approved):
I don't see anything to object to.  I would like to see the two phases be given symbolic names.  E.g.,
```
ATFILE_ONLY = 1
ALL_NODES = 2

#...
for phase in (ATFILE_ONLY, ALL_NODES):
    #...
    if phase == ATFILE_ONLY and not v.isAnyAtFileNode():
    # etc.
```

edreamleo:
@tbpassin Thanks for your comments. I am going to rewrite the code as follows:

- Pass 1: Search body text for `@language` directives. This is the natural way.
- Pass 2: Use the file extension of `@<file>` nodes only pass 1 fails.
  This pass will almost never happen, and there is no need to special case `.txt`: the present default (Python) suffices.

edreamleo:

The new code is much better than the old.

gatesphere:
So... what exactly is the implication for .txt at-file nodes, given 5989f3f?  I currently have code living in .txt files, and demarcated by an `@language` directive, and things seem to work in this context for me already, in current `devel`.  Will that behavior change, because the file happens to be `.txt`, some magic string that the user is somehow supposed to be aware of?  Or am I misreading this, and this fix removes such an instance?

tbpassin (commented):
The new code looks all right.  But I now realize that I don't understand what problem is being tackled here.

The PR title says that the language directive isn't being inherited correctly in .txt files.  What .txt files are we talking about?  External files that are represented by an `@file` tree?  What language would a .txt file have except _text_ or _plain_?  And why would it change mid-file?  Or do we expect a file with a .txt extension to actually contain multiple language directives? 

How could we be dealing with a .txt file without it being an at-file?  The only ones I know about could be nodes used by the rst3 command to create text files (that might or might not have a .txt extension) but those don't need to be processed for a language directive.

Apparently we don't expect to handle multiple language directives because if there are more than one we are going to ignore them all and go with the file extension.  So why don't we start with the top node of the file's tree and look down through all the nodes, rather than work up through parents and extended parents?

IOW, the purpose of this code is unclear to me, and also I don't see why a text file with a language directive needs to be treated any differently from any other external file with a language directive.  I realize that Edward found some problem looking at files in the attic, but I don't grasp what that problem was or why it happened.

I need more information to be able to do an effective review.

edreamleo:
@gatesphere Good question. Imo, you have nothing to worry about :-)

The new code will work *better* than before for any `@<file>` node containing an `@language` directive:
- Previously, under unusual circumstances, the `.txt` extension would cause override the `@language` directive, thereby forcing Python coloring.
- Now, `@language` directives *always* take precedence over file extensions.

edreamleo:
@tbpassin As you say, a `.txt` extension implies an `@<file>`.

The issue arose from the complex nature of the legacy code. The new code is dead simple.  It's also correct.

Issue #3809, opened 2024-02-22
Leo refreshes the wrap state but not the language syntax coloring when moving node outside of parent's directive effect.


boltex:
See PR #3878.

When moving a node outside of a node that had `@language` directive, the syntax coloring does not change. 

However, the wrap state does change, if an `@wrap` or `@nowrap` was in effect.

boltex:
I've discovered this after I implemented this for the next version of both LeoJS and LeoInteg, and trying it out in the original Leo itself! 😄

PR #3878, opened 2024-04-23
PR for #3809: recolor body text when moving nodes

edreamleo:
See #3809.

- [x] Add `jedit.colorize`(!!). It forces `QSyntaxHighlighter` to do a full redraw.
   However, it should not cause another huge performance bug.
- [x] Add a guard to `LeoBody.updateSyntaxColorer`.
- [x] Call `c.recolor` in `c_oc.moveOutlineRight`.
- [x] Add comments to `c_oc.moveOutlineToFirst/LastChild`.


edreamleo:
@boltex @tbpassin I'm going to merge this PR now so that we can all test it further.

Issue #3141, opened 2023-02-10
Leo Crashes On Automatic Restart


tbpassin:
See PR #3879.

The `restart-leo` command does not work when Leo was started from a script.

**Original report**

This has happened to me twice in a row now.  I update my git repo and get a notice from Leo that one of the open outlines has changed.  It asks if I want to restart,  and I click "yes".  Leo restarts, but never gets beyond the splash screen.  CTRL-C brings back control to the console but does not clear the spash screen and Leo won't restart.  If instead I forcibly close the terminal, the splash screen vanishes and the next time I launch Leo from a new terminal it launches normally.

```
Leo 6.7.2-devel, devel branch, build 7d0eb2c4b0
2023-02-10 04:07:08 -0600
Python 3.10.9, PyQt version 6.4.1
Windows 10 AMD64 (build 10.0.19044) SP0
```

I have not tried to find out the cause as yet.  The Windows Event Viewer shows no relevant events.

tbpassin:
This issue occurs running the _restart-leo_ command too (not a surprise). 
It only happens when I start Leo from my batch file that sets the PYTHONPATH to point to my Leo clone.  It does not happen when I cd to the git\leo-editor directory and then start Leo from the console with `py -m leo.core.runLeo`.

So it must be some kind of interaction between the `execve()`  that restarts Leo, and the evvironment set up by the batch file I'm running it from, since the `execve()` call invokes python.exe directly, rather than the batch file I originally started it with.

tbpassin:
Quasi-solved.  While the relaunched Leo is hung displaying its splash screen, if I hit <ENTER> in the console then Leo continues on to complete its startup.  Apparently it gets into some state where it is waiting for the console to read a line.

I have no idea why, but at least it's nothing too serious. it only happens when I launch Leo from my batch file.

edreamleo:
@tbpassin Good work, Thomas. Many thanks for your comments and investigations.

tbpassin:
I'm not sure Leo is completely off the hook yet, because I now remember that when I have shut it down in the past (I mean without restart), that its launching console sometimes also needed me to press \<ENTER\> before it would display the prompt and be usable again.  That's probably the same phenomena, somehow related to my batch file.

The batch file is fairly simple.  

```
setlocal
set PYTHONPATH=....
py -m leo.core.runLeo %*
endlocal
```

[...]

PR #3879, opened 2024-04-23
PR for #3141: restart command crashes

edreamleo:
See #3141.

- [x] Use `Popen.run` to do the restart.
- [x] Add print statements showing the arguments to `Popen.run`.

**Note**: The revised code calculates the files to be reloaded from the open outlines, not the original invocation of Leo.

### DIFF
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -124,17 +124,10 @@ def restartLeo(self: Self, event: LeoKeyEvent = None) -> None:
     # Create the command to restart Leo.
     leo_editor_dir = os.path.normpath(os.path.join(g.app.loadDir, '..', '..'))
     launchLeo_s = fr"{leo_editor_dir}{os.sep}launchLeo.py"
-    if 1:  # Use os.execv.
-        args = ['leo', launchLeo_s] + restart_paths + ['--no-splash']
-        command = fr"{sys.executable} {' '.join(args)}"
-        print(f"{'\n'}Restarting Leo with this command:\n\n{command}")
-        os.execv(sys.executable, args)
-    else:  # Use os.system.
-        python = 'py' if g.isWindows else 'python'
-        files_s = ' '.join(restart_paths)
-        command = fr'{python} {launchLeo_s} {files_s} --no-splash'
-        print(f"{'\n'}Restarting Leo with this command:\n\n{command}\n")
-        os.system(command)
+    args = ['leo', launchLeo_s] + restart_paths + ['--no-splash']
+    command = fr"{sys.executable} {' '.join(args)}"
+    print(f"{'\n'}Restarting Leo with this command:\n\n{command}")
+    os.execv(sys.executable, args)
 #@+node:ekr.20031218072017.2820: ** c_file.top level
 #@+node:ekr.20031218072017.2833: *3* c_file.close
 @g.commander_command('close-window')
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -93,7 +93,7 @@ def restartLeo(self: Self, event: LeoKeyEvent = None) -> None:
             if veto:
                 g.es_print('Cancelling restart-leo command')
                 return
-    # Officially begin the restart process. A flag for efc.ask.
+    # Officially begin the restart process. A flag for efc.ask and efc.on_idle.
     g.app.restarting = True
     # Save session data.
     g.app.saveSession()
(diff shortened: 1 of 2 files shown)


edreamleo:
@tbpassin This PR is ready for your testing.

tbpassin (changes requested):
Line 126:
`python = 'py' if g.isWindows else 'python'`

This is plain wrong.  On Linux the executable name is usually "python3" - I don't know about the Mac - but on any platform the user might be using some other version of Python (e.g., python3.12 vs the system's default python3.11).

Always use `sys.executable` instead.

Line 129 - I know that the option _--no-splash_ can come after the file list, but I'd rather follow normal useage and put all the options first.

tbpassin (changes requested):
This solution does not solve the problem. After restart begins, Leo emits the following and then hangs:
```
C:\Tom\git\leo-editor>Leo 6.7.9-devel, ekr-3141-restart-crash branch, build 9edd07a887
2024-04-23 15:24:50 -0500
```
Pressing &lt;ENTER> allows the launch to continue.  
After manually closing a Leo session launched by `restart-leo`, the terminal window also hangs until &lt;ENTER> is pressed.

Also, Leo should remember and use the _--theme=_ option that was used for the original startup. (Perhaps all the command-line options be remembered and used?)

Tested on Windows with 9edd07a8874912e85931e21310fa83360a8f936f.

edreamleo:
@tbpassin Thanks for your comments. I'll investigate further and do more thorough testing.

tbpassin:
I'll test your branch first.  I will also test it on Linux.  About the differences, I don't know what could be different between our two systems.  I always get my terminals using `<WIN_KEY>-R cmd /k aliases`.  _aliases_ is a batch file that loads some _doskey_ shortcuts.  Maybe that's involved somehow.  I will check that out by using a pristine cmd session.

Oh, I wonder if the terminal is the difference.  There is a newish "terminal" program which can be run instead of the older cmd window we all used to use.  If you are running it (possibly without realizing it's different) that could be a difference.

tbpassin:
The current version of this branch on Linux (XUbuntu) failed to start with this message:
```
    from leo.commands import commanderFileCommands
  File "/home/tom/git/leo-editor/leo/commands/commanderFileCommands.py", line 131
    command = fr"  {executable_s}{'\n'}  {'\n  '.join(args)}"
                                                             ^
SyntaxError: f-string expression part cannot include a backslash
```
I've got Python 3.11 on this VM, vs 3.12 on Windows.  Didn't 3.12 change the syntax for f-strings a bit?

edreamleo:
@tbpassin Good catch, Thomas. I'll test a fix with 3.9.

tbpassin:
I don't think any major Linux distro uses Python higher than 3.11 yet - at least, none of my usual VM distros do.

tbpassin:
In Linux, using the current version of devel, things are even worse - a `restart-leo` command leaves you at a working command line `<ENTER>` doesn't cause Leo's restart to continue.  I haven't found a way to have it continue as yet.

edreamleo:
@tbpassin I'm using `command.exe` with the (Vanilla, I think) `Command Prompt` configuration.

tbpassin:
ChatGPT suggests starting the new Leo session using POpen and then closing the existing instance.  I haven't tried it.  Of course, ChatGPT doesn''t know all the complications involved in closing Leo.
```
    def restart_program(self):
        # Get the Python executable path and the script path
        python_executable = sys.executable
        script_path = os.path.abspath(__file__)

        # Launch a new instance of the program using subprocess
        subprocess.Popen([python_executable, script_path])

        # Close the current instance of the program
        sys.exit()
```

edreamleo:
@tbpassin Thanks for the hint from ChatGPT.  I'll play with it.

edreamleo:
@tbpassin The latest rev uses Popen. On my machine I need to hit `<return>` only after exiting the *restarted* process.

Overall, I think Popen is better. It feels snappier and there is no need to quote everything.

What do you think?

tbpassin:
I tried it and still have the same problems.  I do agree the code is a little easier.

But **Success** (at least on Windows) - I replaced Popen with run and now it works without needing the `<ENTER>`s -

`subprocess.run(popen_args)`

edreamleo:
@tbpassin Excellent. I'll commit your version next.

tbpassin:
I take it back - on Linux using `subprocess.run(popen_args)` did not require a final ENTER after manually closing the re-started session. I had forgotten to manually change the code last time.

However, I noticed that the restarted session did not open an outline I had opened while the original session was running.  I saw this on Linux but I suppose it is probably the same on Windows.

edreamleo:
@tbpassin You are correct. The `restart-leo` command only reopens the outlines that were opened when the `restart-leo` command is executed.

Please approve this PR. We can discuss enhancing it later, but that's a low priority for me.

tbpassin (approved):

edreamleo:
@tbpassin Thanks for your approval. The first comment in this PR mentions that the code only opens outlines that are open at the time the user executes `restart-leo`.

PR #3886, opened 2024-04-25
PR: restart-leo should not call sys.exit after Popen.run

edreamleo:
`Popen.run` *replaces` the running process, so the later call to `sys.exit` seems confusing and redundant.

However, this PR requires thorough testing.


edreamleo:
@tbpassin Thomas, I'll wait for your review :-)

tbpassin:
I just tested on Linux (XUbuntu) and it worked without requiring any ENTERs.  Windows coming up next...

tbpassin (approved):
This change works, as tested on both Linux and Windows 10.  I approve the PR.

But I do suggest a few naming changes before the PR is finalized.  All names like _popen_args_s_ should replace the _popen_ with _run_, since we are now using `run()` instead of `popen()`.  Perhaps _cmd_ would be even better in case we change away from `run()` in the future.

I don't think these changes should require my approval again.

edreamleo:
@Thanks for your testing.

Hah! The command echo said `Popen.run` rather than `subprocess.run`. 

I'll change the echo and remove the prefix entirely. There is now only one flavor of command.

Issue #3895, opened 2024-04-29
Freewin plugin - TypeError: setHtml(self, text: Optional[str]): too many arguments


LewisNeal:
Leo 6.7.9-devel, devel branch, build 15f6b5fa8d
Python 3.12.2, PyQt version 6.7.0
Windows 10 AMD64 (build 10.0.19045) SP0

Running z-open-freewin command then pressing the [Rendered<-->Plain] toggle gives error:

_Uncaught exception in Leo...
Traceback (most recent call last):
  File "N:\git\leo-editor\leo\plugins\freewin.py", line 967, in switch_and_render
    self.set_and_render(True)
  File "N:\git\leo-editor\leo\plugins\freewin.py", line 960, in set_and_render
    self.browser.setHtml(html, self.url_base)
TypeError: setHtml(self, text: Optional[str]): too many arguments_

**dummy** then displays in the Freewin rendered window.

tbpassin:
Hmmm, I just opened Freewin for that branch and didn't get an error.  I'm using 
PyQt version 6.6.2.  You are using 6.7.0. The docs for 6.7.1 say that the url_base parameter is an allowed optional argument for that method.

I wonder if PyQt 6.7.0 has a bug here.  I'll look into it.

tbpassin:
If you keep getting the error, there is an alternative way I can set the base url, so I could change the code to use it.  It would still be extremely strange, though, to get that error.

LewisNeal:
py -m pip list |find /i "pyqt6"
gives:
PyQt6                         6.7.0
PyQt6-Qt6                     6.7.0
PyQt6-sip                     13.6.0
PyQt6-WebEngine               6.7.0
PyQt6-WebEngine-Qt6           6.7.0

I then reinstalled pyqt6 and pyqt6_webengine with no improvement. Also reinstalled PyQt6-sip.
Same result with:
PyQt6                         6.6.0
PyQt6-Qt6                     6.6.0
PyQt6-sip                     13.6.0
PyQt6-WebEngine               6.6.0
PyQt6-WebEngine-Qt6           6.6.0

and same result with:
PyQt6                         6.6.1
PyQt6-Qt6                     6.6.1
PyQt6-sip                     13.6.0
PyQt6-WebEngine               6.6.0
PyQt6-WebEngine-Qt6           6.6.0

I even downgraded docutils 0.21.2 to 0.21.1 but no change to error.

tbpassin:
[...]

PR #3897, opened 2024-04-30
FreeWin Fixes - Better Display Of Local Images

tbpassin:
- Renders local relative image files.
- Remove code for alternate rendering widget because qt6 now has a QWebEngineView.
- Don't render node if rendering widget is hidden.
- If node is empty render view shows a blank screen. with stylesheet background color.
Fixes Issue #3895.


edreamleo (approved):
Thanks, Thomas.

Issue #3898, opened 2024-05-05
LeoServer.py: Add current commander info in get_ui_states command


boltex:
For leoserver : Add the current commander info to the get_ui_states command return data.  This will enable more flexible UI interaction in LeoInteg or any other clients connected to it.

edreamleo:
@boltex I'll do this issue immediately.

boltex:
@edreamleo I'll make some more additions to the server. I spoke to soon in thinking those modifications were enough to implement detached body panes in LeoInteg. (or other external clients) This will take me a day or so. 

(I'll make a new branch and pull request.)

edreamleo:
@boltex No problem. Take as much time as you want.

PR #3899, opened 2024-05-05
adding more info in 'get_ui_states' command.

boltex:
Adds more info to the get_ui_states data

PR #3900, opened 2024-05-05
PR: Add commander entry in value returned by get_ui_states

edreamleo:
See #3898.

- [x] Add "commander" key to the returned value. Use the same value as in `get_all_open_commanders`.


edreamleo:
Closed in favor of PR #3899.

Issue #3901, opened 2024-05-05
Add toggle-unl-view command


edreamleo:
See PR #3902. This command toggles between path-based and gnx-based UNLs in the status area.

The command will affect *no* settings related to UNLs or gnxs.

edreamleo:
Closed via PR #3902.

PR #3902, opened 2024-05-05
PR for #3901: Add toggle-unl-view

edreamleo:
See #3901.

### DIFF
--- a/leo/core/leoFrame.py
+++ b/leo/core/leoFrame.py
@@ -898,9 +898,9 @@ class LeoFrame:
         if self.statusLine:
             self.statusLine.clear()
 
-    def computeStatusLine(self, p: Position) -> str:
+    def computeStatusUnl(self, p: Position) -> str:
         if self.statusLine and p:
-            return self.statusLine.computeStatusLine(p)
+            return self.statusLine.computeStatusUnl(p)
         return ''
 
     def disableStatusLine(self, background: str = None) -> None:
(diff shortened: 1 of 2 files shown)


edreamleo:
@tbpassin I think you will like this new command. It sidesteps all issues related to settings.

The new command came to mind because of a remark you made on another issue.

tbpassin (changes requested):
1. As a general comment, it seems to me that some of the new code in qt_frame.py could be moved to a more general location, since it has nothing Qt-specific in it.  This way it would be easier to port to some other GUI framework (not that it's likely, but there is a precedent ...).
2. leoFrame.py, lines 2122 ff - I'd like to see a docstring saying what this method is for and that it is expected to be overridden by derived classes.  Many other empty methods are doc-stringed like this.
3. qt_frame.py, line 4379 - _Compute the UNL part of the status line_ - this docstring conflicts with the method name (_computeStatusLine_) since the method name says it computes the whole status line, not just part.  Please make these compatible.
4. qt_frame.py, line 4304 - This attribute name (`self.toggle_view`) should say what kind of view is being toggled.  Perhaps `self.toggle_unl_style`?
5. `p.get_UNL()` could be changed to have an optional boolean argument for returning gnx- or legacy-based UNLs.  This would simplify the code here, would be helpful to authors of other scripts, and would not require any other changes to existing Leo code.

PR #3903, opened 2024-05-07
Correcting stickynotes.py

XCaminhante:
QMdiArea.SubWindowView became QMdiArea.ViewMode.SubWindowView at PyQt6
QMdiArea.TabbedView also became QMdiArea.ViewMode.TabbedView

### DIFF
--- a/leo/plugins/stickynotes.py
+++ b/leo/plugins/stickynotes.py
@@ -652,18 +652,18 @@ class Tabula(QtWidgets.QMainWindow):  # type:ignore
         self.tb.setObjectName("toolbar")
 
         def do_tile():
-            self.mdi.setViewMode(QMdiArea.SubWindowView)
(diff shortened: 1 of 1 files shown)


XCaminhante:
Qt.WA_DeleteOnClose became Qt.WidgetAttribute.WA_DeleteOnClose
I flipped DeleteOnClose:False with DeleteOnClose:True, then added try-catches at the relevant places so the widget is recreated when necessary.

Also replaced self.mdi.removeSubWindow(i) with i.close(), so the widget frame don't became a ghost.

XCaminhante:
I removed the function blocking onClose() event, I didn't see the need of it.
I also modified the stickynote command so it can steal notes from tabula when is not visible, but it's glitching.

edreamleo (approved):
Thanks for this work.

Issue #3905, opened 2024-05-10
Fix problems with c.frame.ratio and c.frame.secondary_ratio properties


edreamleo:
See PR #3906. A *later* PR will handle bugs in the `compute_ratio` method.

edreamleo:
Closed via PR #3906.

PR #3906, opened 2024-05-10
PR for #3905: ratio properties

edreamleo:
See #3905.

This PR contains the minimal changes required to deprecate the unwanted (deprecated) proprieties.

A *later* PR will tackle bugs in `c.frame.compute_ratio` and `c.frame.compute_secondary_ratio`.

- [x] Add deprecation message to `LeoQtFrame.ratio` and `LeoQtFrame.secondary_ratio` properties.
- [x] Replace these properties with `c.frame.compute_ratio` and `c.frame.compute_secondary_ratio` methods everywhere.
- [x] Add corresponding do-nothing methods to LeoFrame class.
- [x] Remove all `ratio` and `secondary_ratio` ivars.

**Extras**

- [x] Fix recent crasher in `cursesGui2.py` by defining a do-nothing `computeStatusUnl` method.
- [x] Fix several pylint complaints.
- [x] Remove dead code in `fc.leojs_globals` and change the annotated return type to `None`.
- [x] Simplify `LeoOPML.py` by not computing the `ratio` ivar.

### DIFF
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -133,7 +133,7 @@ def restartLeo(self: Self, event: LeoKeyEvent = None) -> None:
     print('Restarting Leo with:\n')
     print(args_s)
     print('')
-    subprocess.run(args)
+    subprocess.run(args)  # pylint: disable=subprocess-run-check
 #@+node:ekr.20031218072017.2820: ** c_file.top level
 #@+node:ekr.20031218072017.2833: *3* c_file.close
 @g.commander_command('close-window')
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -286,8 +286,8 @@ def new(self: Self, event: LeoKeyEvent = None, gui: LeoGui = None) -> Cmdr:
     frame.lift()
 
     # Resize the _new_ frame.
-    frame.resizePanesToRatio(frame.ratio, frame.secondary_ratio)
-    c.frame.createFirstTreeNode()
+    frame.resizePanesToRatio(frame.compute_ratio(), frame.compute_secondary_ratio())
+    frame.createFirstTreeNode()
 
     # Finish.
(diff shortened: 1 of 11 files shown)


edreamleo:

- A later PR will tackle any bugs in `c.frame.compute_ratio` and `c.frame.compute_secondary_ratio`.
- Imo, this PR contains the unavoidable price to be paid to make Leo's core truly compatible with the `nested_splitter` plugin!

tbpassin (commented):
qt_frame.py, lines 4775, 4776 - The usual idiom in Leo code has been to create a separate variable if it's going to be used more than once.  Here that idiom is removed:
```
# old
    f = c.frame
    r = max(0.0, f.ratio - 0.1)
    f.divideLeoSplitter1(r)
# new
    r = max(0.0, c.frame.compute_ratio() - 0.1)
    c.frame.divideLeoSplitter1(r)
```
Thus _c.frame_ is now retrieved twice.  It's not wrong but it's not consistent with Leo's normal code practice. 

Otherwise I didn't notice anything to comment on.

edreamleo:
@tbpassin Thanks for your comment.

What do you think of removing the two properties instead of deprecating them?

I would like to do that because apparently the properties interfere with pylint and mypy checks. Furthermore, there is no obvious use case for the properties. Yes, technically removing the properties would be a breaking change, but almost surely nobody would notice.

tbpassin:
I'd rather see them stay for a while, since we can't be sure no one is using them for some tricky purpose.  Can't the pylint and mypy warnings be suppressed with #ignore type comment lines?

I do appreciate that new methods with different names were introduced rather than reusing the older methods.

edreamleo:
The problem is that the properties *suppress* messages.

Anyway, I can foresee no "tricky purpose" that justifies the properties.

edreamleo:
@tbpassin I've moved the properties to the attic so it will be easy to restore them in the unlikely event that someone wants them.

It's time to merge this PR and use it in the VR project.

Issue #3910, opened 2024-05-15
Deprecate the nested_splitter and free_layout plugins


edreamleo:
This is an *experimental* issue. #3909 must be completed first.

The `nested_splitter` and `free_layout` plugins are problematic. See [this post](https://groups.google.com/g/leo-editor/c/1C4_AzG1www).

This issue contemplates deprecating these two plugins.

If possible, Leo would *not* enable the two plugins by default.

PR #3911, opened 2024-05-15
PR for #3910: deprecate free_layout and nested_splitter plugins

edreamleo:
See #3910.

Leo's core will never load these two plugins, but other plugins should be allowed to do so.
**Warning**: The free_layout may not work as expected because of changes to `dw.createMainLayout`.

- [x] `LeoQtGui`: Add `find_widget_by_name`, `get_top_splitter`, and `_self_and_subtree`.
- [x] Set `c.free_layout = None` (permanently!) in `c.init_objects`.
- [x] `dw.createMainLayout` creates `QSplitter`s, not `NestedSplitters`s.
- [x] Similarly, revise several `QtFrame` methods, especially `toggle-split-direction`.
- [x] Revise the VR plugin. The `viewrendered` method places the VR pane depending on settings.
- [x] Remove the `ViewRenderedProvider` and `ViewRenderedProvider3` classes.
- [x] Revise the VR3 plugin as a *prototype*. Thomas is entitled to change what I've done.
- [x] Revise the `bookmarks.py` plugin so that it doesn't use `free_layout`.
- [x] Similarly, revise the `editpane` plugin and `pyplot_backend` plugins.
- [ ] Create an info item that contains tips and examples for converting layouts.

**Extras**

- [x] Add `NestedSplitter.dump_layout`. This is a non-trivial method!

### DIFF
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -128,6 +128,9 @@ def restartLeo(self: Self, event: LeoKeyEvent = None) -> None:
     leo_editor_dir = os.path.normpath(os.path.join(g.app.loadDir, '..', '..'))
     launchLeo_s = fr'{leo_editor_dir}{os.sep}launchLeo.py'
     args = [sys.executable, launchLeo_s] + restart_paths + ['--no-splash']
+    # Add any --trace args.
+    for z in g.app.debug:
+        args.append(f"--trace={z}")
     args_s = 'subprocess.run([\n  ' + ',\n  '.join(args) + '\n])'
     print('')
     print('Restarting Leo with:\n')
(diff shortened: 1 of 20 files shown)


tbpassin (approved):
I didn't see anything to suggest.  It wouldn't be a surprise if there are a few unexpected effects!

edreamleo:
@tbpassin Thanks for your quick review. There is a lot more work to do.

tbpassin:
I just worked out how to get VR3 (or any widget, really) to share the stacked widget which contains the body editor:

```
    body_stacked_widget = gui.find_widget_by_name(c, 'bodyStackedWidget')
    stacked_layout = body_stacked_widget.children()[0]

    h = c.hash()
    if stacked_layout.count() == 1:
        v3.controllers[h] = vr3 = v3.ViewRenderedController3(c)
        stacked_layout.addWidget(vr3)
```
Then you can select the stacked layout's index for the body editor or for VR3.  For future-proofing it would be a good idea to select by widget instead of index number in case some other widget wants to share to stacked layout too.  But for now, my toggling script, with this change, works fine.

If the stacked layout were given a name as well, we could find it in one step instead of two.  Right now it is nameless.

edreamleo:
@tbpassin This PR is ready for review. There is no rush to merge it.

PR #3913, opened 2024-05-15
PR for #3912: crash on MacOs

edreamleo:
See #3912.


tbpassin:
I was surprised that we didn't see the message from

```
        except Exception:
            g.error(f"unexpected error creating: {fn}")
```
The actual error message was better...

PR #3914, opened 2024-05-15
Fix erroneous command if pyenchant library isn't present.

tbpassin:
This bug showed up on a Mac but it would apply to any OS if the pyenchant library isn't available.


edreamleo (approved):
Thanks Thomas. I have no idea where the erroneous `.d` came from.

edreamleo:
Rev 1c56b75 in devel fixes two annotations.  All is well.

tbpassin:
I think it must have been from some much earlier version.  Maybe it used to be correct.  The branch of code is never hit if pyenchant isn't present, so no one has seen it now that pyenchant is available again.

Issue #3916, opened 2024-05-18
Restart-leo should support all command-line args


edreamleo:
Should be straightforward. Just supporting `--trace=whatever` is useful.

edreamleo:
Closed via PR #3922.

PR #3922, opened 2024-05-18
PR for #3916: Improve `restart-leo` command

edreamleo:
See #3916.

- [x] Save `os.getcwd()` for `restart-leo`.
- [x] Use the original `sys.argv`.

This PR needs thorough testing, despite its apparent simplicity.

### DIFF
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -99,12 +99,6 @@ def restartLeo(self: Self, event: LeoKeyEvent = None) -> None:
     # Save session data.
     g.app.saveSession()
     g.app.setLog(None)  # Kill the log.
-    # #3141: Remember all open outlines.
-    restart_paths: list[str] = [
-        c.fileName() for c in g.app.commanders() if c.fileName()
-    ]
-    if g.isWindows:
-        restart_paths = [z.replace('/', os.sep) for z in restart_paths]
     # Close all unsaved outlines.
     for c in g.app.commanders():
         frame = c.frame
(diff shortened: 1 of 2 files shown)


edreamleo:
@boltex @tbpassin This PR is ready for review. I have merged it into the branch I use for the nested-splitter work.

tbpassin (changes requested):
CommanderFileCommands, lines 121ff -

1. I don't think this will work reliably.  The reason is that outlines given on the command line may have relative paths, but the effective directory for `run()` may have changed.  So the relative paths would be wrong.

2. I think someone having Leo restart in the middle of a session would expect to see their open outlines re-opened, but not necessarily all the outlines that were opened during the original startup.  This makes it an interesting exercise - to keep all the other command line options but change the outlines in the restart args.

edreamleo:
@tbpassin Thanks for your comments. I agree that something must be done.

But what? The only way to convert a relative to an absolute that preserves the meaning is to know the cwd.

Perhaps Leo should remember the cwd on startup.

tbpassin:
> Perhaps Leo should remember the cwd on startup.

I thought that too. It seems feasible, doesn't it? The run() command could just cd to that directory before invoking Leo.

For outlines that were opened after Leo started and were open at restart, their commander already knows its absolute path.  So nothing special needs to be done for them.

edreamleo:
@tbpassin Saving and restoring the cwd was straightforward. What do you think?

tbpassin (approved):
This part seems straightforward enough.

edreamleo:
@tbpassin Thanks for your help, Thomas.

PR #3928, opened 2024-05-22
PR for #3915: Require slots

edreamleo:
See #3915.

- Classes with slots can not be monkey-patched!
- All classes in `leoNodes.py` already have slots.

- [x] Add slots to the `Commands` class.

**Changes related to leoApp.py**

- [x] Require slots for all classes in `leoApp.py`:
    - [x] Add slots to the `LeoApp` class.
    - [x] Add slots to the `IdleTimeManager` class.
    - [x] Add slots to the `RecentFilesManager` class.
    - [x] Add slots to the `PreviousSettings` class.
    - [x] Add slots to the `LoadManager` class.
- [x] Avoid monkey-patching methods by testing `g.leoServer` and calling the corresponding methods.
- [x] `test_leoServer.py` sets `g.leoServer`.

Issue #3915, opened 2024-05-17
Use slots for most of Leo's classes


edreamleo:
PR #3928 went too far. PR #3929 adds a few slots to small classes.

This issue was a disappointment. It did not improve Leo's ability to catch references to non-existent ivars.

edreamleo:
Closed via PR #3929.

PR #3929, opened 2024-05-22
PR 2 for #3915

edreamleo:
See #3915.

This PR replaces PR #3928. It contains only harmless changes.


tbpassin (commented):
[leoapp.py]
-   Line 3376 (not part of the PR)
    `self.edit_headline = 'Recent files. Do not change this headline!'`

   This could be a class level attribute or a module level symbolic constant.  It's not really an instance variable.

Everything else seems all right to me.

edreamleo:
@tbpassin Thanks for your comment. The `edit_headline` ivar and its logic is mysterious to me. It's part of a dubious feature that I never use.

tbpassin:
I looked for uses and found three of them.  It seems to be used as a target for a search so it shouldn't have the text changed (unless it becomes a symbolic name at a higher level).  Why one might want to search for it I haven't looked into.

edreamleo:
@tbpassin I don't understand and I don't want to investigate :-)

tbpassin:
With the latest devel branch I don't see those three usages anyway.

edreamleo:
@tbpassin :-)

Issue #3935, opened 2024-05-27
 test_most_public_server_methods fails after PR #3919


edreamleo:
PR #3919 added the `get_is_valid` method. The unit test fails because method will throw an exception if `ap` is None.

I'll submit a new unit PR with the obvious fix.

edreamleo:
Fixed via PR #3936.  There is no need for PR #3941!

PR #3936, opened 2024-05-27
PR for #3935: fix server.is_valid

edreamleo:
See #3935. This PR passes all tests, including pyflakes, pylint and beautifier.

- [x] `server.is_valid`: Guard `c` and  `ap` by enclosing the code in a `try/except` block.
   The guard on `gnx` is still needed to avoid incorrectly returning `self._make_minimal_response({'valid': True})`
- [x] `server._get_p`: Fix a valid pylint complaint.


edreamleo:
@boltex This PR is ready for review.  Please review it asap.

boltex (approved):
Thanks for those adjustments! They are very good! (Why didn't I see that originally? Ha! )

edreamleo:
@boltex Thanks for the approval :-)

Issue #3934, opened 2024-05-26
the cloneToAtSpot command doesnt undo the 'spot' node when undoing


boltex:
The `clone-to-at-spot` command doesn't undo the 'spot' node when undoing: It undoes the 'clone', but forgets the 'spot' node it may have created.

The undo code should use `u.beforeUndoGroup` and `u.afterUndoGroup`.

edreamleo:
@boltex Thanks for your report. I'll fix this immediately.

edreamleo:
Closed via PR #3937.

PR #3937, opened 2024-05-27
PR for #3934: undo for clone-to-at-spot

edreamleo:
See #3934.

- [x] Undo the operation using `u.beforeUndoGroup` and u.`afterUndoGroup`.
- [x] As before, revert to initial state if the clone would be invalid.

### DIFF
--- a/leo/commands/commanderOutlineCommands.py
+++ b/leo/commands/commanderOutlineCommands.py
@@ -1011,37 +1011,48 @@ def cloneToAtSpot(self: Cmdr, event: LeoKeyEvent = None) -> None:
     of the outline. Create the @spot node if necessary.
     """
     c, p, u = self, self.p, self.undoer
+    tag = 'clone-to-at-spot'
     if not p:
         return
     # 2015/12/27: fix bug 220: do not allow clone-to-at-spot on @spot node.
     if p.h.startswith('@spot'):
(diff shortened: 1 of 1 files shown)


edreamleo:
@boltex I'm not sure this PR does what you want. However, I think *in this case only* a two-step undo is convenient.

boltex:
@edreamleo Could an undo group not be favorable so as to be undone / redone in one step?

edreamleo:
@boltex Yes, the command could be undone in one step. Would a two-step undo would be inconvenient?

edreamleo:
@boltex Ready for review.  Undo/redo work (in one step) properly regardless of whether an `@spot` node exists.

The tricky part is undoing and redoing `c.p`.  All my hand tests now work as expected.

boltex:
Thank you for making it a one-step undo. 

 I work in front-end, UI, and interface design and interaction. In that field, the concept of a "more than one step" undo is an unthinkable cardinal sin! Hahah! :)  (an undo action should be a one-shot action that reverts the last command, if undoable)

Many thanks again for re-considering and making the command!! 😄

boltex (approved):
This looks good!! Many thanks!

edreamleo:
@boltex Thanks for your comments and approval.

PR #3938, opened 2024-05-27
PR for Leo 6.8.0

edreamleo:
This PR describes all the diffs between 6.8.0 and 6.7.8.

This release must be named 6.8.0 because of breaking changes to Leo's API.

See the distribution checklist for a list of tasks.


tbpassin:
I already found one serious problem.  The new CTRL-click doesn't work right.  In the _ekr-3871-search-settings_ branch a CTRL-click on `logTabWidget = self.get_top_splitter(` found and navigated to a definition.  There are actually two definitions and they both were displayed in the Nav pane.

But in the current _6.8.0 branch_, the search phrase didn't show up in the Nav pane and the Nav pane did not show both of the two matches that exist in LeoPyRef.leo.

edreamleo:
@tbpassin Thanks for your testing. Everything should work once I merge the new ctrl-click stuff into devel and from devel into the 6.8.0 branch.

tbpassin:
I'm working on the review... there is a lot to cover.

tbpassin:
There are quite a few lines like this still in the codebase:
`splitter = c.free_layout.get_top_splitter()`

Can the method be changed to return _main_splitter_?  Or maybe that wouldn't be a good idea because old user code would expect it to be a Nested Splitter.  But maybe deprecate it to print a message and return None?

edreamleo:
@tbpassin Wow. Thanks for even considering a review :-)

This is not the place to worry about the free_layout changes. Anyway, the code should stay as it is unless somebody screams.

tbpassin (dismissed):
### mypy.ini
#### `python_version = 3.12` 
Used to be 3.9.  What will happen for users with versions < 3.12?

#### Lines 12, 13 - Are these compatible with each other?
```
# mypyc: strict_optional must be true: use--strict-optional
strict_optional = False
```
### PKG-INFO.TXT
Update `Version: 6.7.9.a1` to `Version: 6.8.0a1`

### abbrevCommands.py
#### Lines 174ff:
```
                self.subst_env = []
            else:
                self.subst_env = c.config.getData(key, strip_data=False)
```
Won't the _else_ branch return a string? The new code just above returns a list.  Shouldn't they both return the same kind of object?

### commanderEditCommands
#### Lines around 531:
```
    for pat_s in c.config.getData('extract-patterns') or []:
        try:
            pat = re.compile(pat_s)
```
Instead of names like `pat_s`/`pat = re.compile(pat_s)`, I like to use a suffix of "re" for one of them, e.g., `pat`/`pat_re` = `re.compile(pat)`,

It is a little more mnemonic and a little less to keep in mind while reading the code.

#### Line 524:
`def rp_get_leading_ws(c: Cmdr, lines: list[str], tabWidth: int) -> tuple[list[int], list[str]]:`
A welcome improvement.

### leoAtFile.py
#### line 3292: `self.root.gnx == gnx`
This should be `self.root.gnx = gnx`, shouldn't it?

I get nervous about changes in at-file handling.  I hope the unit tests are **very** reliable!

#### Line 3625: `f"code? {int(self.in_code)} "`
Why is this printing the int value of the boolean instead of just the boolean?  I would think we'd want to see "True" or "False" here.

###leoCache.py
#### Lines 215 ff
```
        result = [
            '\n',
            f"SqlitePickleShare('{self.root}')\n",
        ]
        for key_tuple in self.keys():
            key = key_tuple[0]
            result.append(f"{key} {self.get(key)}\n")
        return ''.join(result)
```
I suggest removing the "\n" from each entry, and using 
`return '\n'.join(result)` at the end.  This is simpler and more readable. A trailing "\n" can always be added if desired.

### leoFileCommands
#### Line 1857:
`aList2: list[tuple[VNode, dict]] = self.createUaList(vnode_list)`
A few lines earlier we got rid of the old _aList_ in favor of _vnode_list_.  Here, we should likewise change _aList2_ to _vnode_list2_.

### leoTokens.leo
I did not review the new methods,

### leoVersion.py
#### Line 55 - update version to 6.8.0.
#### Line 56 - update date.

### leoServer.py
#### Line 2381 - ` Return p.b's length in bytes, where p is c.p if param["ap"] is missing.`
This method now returns a Response, not a length.  Please fix either the docstring or the return type.

### leoDist.py
#### Line 201 - Update the version.
#### Line 1984 - Update the version.

### leoDocs.leo
I did not review this outline.

### Plugins
I did not review any plugins.

edreamleo:
@tbpassin This PR is not the place to request changes. Working through the distribution checklist will change version numbers, etc.

However, the question about mypy version deserves an answer. Leo itself should run on Python 3.9, but Leo's annotations assume that Python 3.12 features are in effect. In other words, I only run mypy using Python 3.12.

edreamleo:
@tbpassin

Yikes. You found a serious bug in leoAtFile.py.  See #3957.

Other comments
- `c.config.getData` returns a list of lines.
- `f"code? {int(self.in_code)}"`: This is a trick to make lines line up.
- `leoFileCommands`: I'm not going to change the names of ivars unless there is a real bug.
  Furthermore, aList2 contains tuples, not VNodes, so the name change would not be appropriate.

Issue #3946, opened 2024-05-28
Leo fails to load if previously optional Qt modules do not exist


edreamleo:
See PR #3947. Leo should run if optional Qt modules do not exist.

**Background**

Leo 6.7.8 eliminated conditional  (try/except) imports for optional Qt modules.
The assumption was that `pip install leo` would install those module.

But *this assumption is false* when using GitHub unless users remember to run `pip install -r requirements.txt`.
In that case, Leo won't run because of missing optional Qt modules, especially `uic`.

edreamleo:
Closed via PR #3947.

PR #3947, opened 2024-05-28
PR for #3946: optional Qt modules

edreamleo:
See #3946.

Most of the work in this PR involves making `leoQt.py` more robust:

- [x] Add testing section that sets all optional modules to None.
   Leo loads, but after loading throws exceptions because `uic` is None.
- [x] Import all optional Qt modules in `try/except` blocks.
  Set the modules to `None` if they fail to load, and add their name to the *public* `_missing_modules` list.
- [x] Add disabled code reports the missing modules.

**Other files**

- [x] runLeo.py always give a traceback on failure-to-load.
  This traceback is to important to omit!
- [x] The `backlink`, `graphcanvas` and `todo` plugins load only if the `uic` module exists.

### DIFF
--- a/leo/core/leoQt.py
+++ b/leo/core/leoQt.py
@@ -74,20 +74,9 @@ else:
         from PyQt6 import uic
     except Exception:
         uic = None
-
-    # Give a hint if any optional module does not exist.
-    if (
-        not Qsci or not QtDesigner or not QtMultimedia or not QtNetwork
-        or not QtOpenGL or not printsupport or not QtWebEngineCore
-        or not QtWebEngineWidgets or not QtSvg or not uic
-    ):
-        print('')
-        print('leoQt.py: one or more optional Qt modules do not exist.')
-        print('Please run `pip install -r requirements.txt`')
-        print('')
     #@-<< import optional Qt modules >>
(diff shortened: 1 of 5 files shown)


edreamleo:
@tbpassin This PR allows me to run Leo again on Ubuntu.

I shall merge this PR later today so that we can all test it more thoroughly.

tbpassin (approved):
I'm glad to see the uic has been decoupled from the main Leo code.  I didn't see why it had been made mandatory.

### leoQt.py
#### Lines 73ff:
```
 try:
        from PyQt6 import QtWebEngineWidgets
    except Exception:
        QtWebEngineWidgets = None
        _missing_modules.append('PyQt6.QtWebEngineWidgets')
```
These imports all have the same structure and could be replaced by a table-driven loop.  This might be a more reliable way to make sure all the import modules, missing module list entries, etc. remain coordinated in the future.

### runLeo.py
#### Lines around 47 -
```
 message = (
        '*** Leo could not be started ***\n'
        'Error starting Leo!\n'
```
When do we print the list of missing Qt modules?  Or if we don't print it, how is it supposed to be used?

Issue #3957, opened 2024-05-30
Fix blunder in leoAtFile.py


edreamleo:
See PR #3958.

At line 3292 the statement `self.root.gnx == gnx` has no effect.

**Update**

I am going to remove the "ReleaseBlocker" label. The bug was apparently less serious than I thought at first.
However, an extra week of testing is still justified.

edreamleo:
Closed via PR #3958.

PR #3960, opened 2024-06-03
PR 2 for #3957: Fix Leo's read code

edreamleo:
See #3957.

These changes must be *carefully* reviewed by as many Leonine experts as possible.

- [x] `refresh-from-disk` when the file's gnx overrides `p.v.gnx`.
- [x] Rewrite case 1 of the section << handle node start >>`.
   The new code removes all traces of `root_v.gnx` if it conflicts from the file's gnx.

### DIFF
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -373,7 +373,7 @@ def refreshFromDisk(self: Self, event: LeoKeyEvent = None) -> None:
         g.es_print(f"Unknown @<file> node: {p.h!r}")
         return
     if p.v.gnx != old_gnx and not g.unitTesting:
-        g.trace(f"Changing gnx! old: {old_gnx} new: {p.v.gnx} in {p.h}")
+        print(f"refresh-from-disk: gnx for `{p.h}` changed from: `{old_gnx}` to: `{p.v.gnx}`")
     c.selectPosition(p)
     # Create the 'Recovered Nodes' tree.
     c.fileCommands.handleNodeConflicts()
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -352,6 +352,7 @@ def refreshFromDisk(self: Self, event: LeoKeyEvent = None) -> None:
     at = c.atFileCommands
     c.nodeConflictList = []
     c.recreateGnxDict()
+    old_gnx = p.v.gnx
     if p.isAtAutoNode() or p.isAtAutoRstNode():
         p.v._deleteAllChildren()
         p = at.readOneAtAutoNode(p)  # Changes p!
@@ -371,6 +372,8 @@ def refreshFromDisk(self: Self, event: LeoKeyEvent = None) -> None:
     else:
         g.es_print(f"Unknown @<file> node: {p.h!r}")
         return
+    if p.v.gnx != old_gnx and not g.unitTesting:
+        g.trace(f"Changing gnx! old: {old_gnx} new: {p.v.gnx} in {p.h}")
     c.selectPosition(p)
     # Create the 'Recovered Nodes' tree.
     c.fileCommands.handleNodeConflicts()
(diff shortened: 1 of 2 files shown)


tbpassin:
I don't understand what this code is supposed to do, so I don't know how to review it.  It seems to me that there are two potential cases:

1. The user asks to refresh an external file, because he has messed up the one in the outline.  In this case the file from disk should be read in like any external file would be when an outline gets opened - I assume that the import keeps the gnx's that are contained in the external file;
2. The user asks to refresh an external file but for some reason, unknown to the user, that file has been changed or corrupted.  If it has, there must be some policy for what should happen.  It's also possible that it has been changed in such a way that it could not be successfully recreated.  In that place, I suppose we'd like to keep the version in the outline.

Either way, one wants to know if the external file was changed before the refresh - after all, if it had been changed, a "refresh" would not recreate the file the way the user thought it had been.  Reading the external file in and seeing if nodes or gnx's had changed wouldn't assure that the file on disk was actually correct.

This makes me think that every time an external file gets read into the outline, some kind of hash or checksum should be stored to compare the external file against as the first step in a "refresh".  If we don't care about trailing line endings then they could be removed before taking the hash.

At any rate, I need to know more about what is supposed to be accomplished before I can do a meaningful review.

edreamleo:
@tbpassin The changed code handles `@+node` sentinels. In particular, the changed code handles the `@+node` sentinels for the `@file` node itself.

Perhaps the new comment is a bit misleading. The code has nothing *directly* to do with `refresh-from-disk`. But suppose (as recently happened to me) that I discover that my personal `leoPy.leo` file is missing an `@file` node for, say, `leoBackground.py`. I create the node: `@file leoBackground.py`.  This new node will have a *different* gnx than the corresponding `@+node` sentinel in the external file.

All the recent changes were intended to *change* the gnx in the root node so that it uses the gnx in the external file.  I think today's changes work at last.

edreamleo:
@tbpassin Interesting question. With the *new* code importing the file does work. However, I had to change the root's headline from `@file <full-path-to>leoBackground.py` to `@file leoBackground.py`. There a lots of *user* test cases to consider, but the new code probably "just works" because there is just one path through the code.

tbpassin:
I just tried this out on the PR's branch.  The external file I used was an `@file`, not an `@clean`.

1. I have an external file I created back in January.  I created a new outline and imported it.  I pasted GNX of the top-level node of this imported file into a node.  Then I went to the January outline and copied the file's GNX from there.  I pasted it into the same node I had pasted the other GNX so I could compare them easily.

The imported external file had a different GNX than the original one.

2. I went back to the original outline and added some text to the body of the external file's top node.  Then I refreshed from disk.  The top node of the refreshed external file stayed the same.

I guess this is all OK, although I would have expected the imported file to be a clone and not a copy with new GNX's.   As I think more on it, there is no way an imported `@clean` file could have the same GNXs, because it doesn't contain any sentinels.  Having the same (non-cloning) behavior for `@file` nodes is consistent, so that's good.

edreamleo:

tbpassin (approved):
leoAtFile.py. line 3275 `v = root_v`

I find this part a little confusing because `v` is getting reused.  Earlier, at line 3268, we have 
`v = gnx2vnode.get(gnx)`
but then we get 
`if root_v.gnx != gnx:`

This leaves me thinking "What?? `v` is the node for `gnx` but now its gnx might be different?" I know this is only a matter of temporary names but for some reason it seems unduly confusing to me.  I think it would be better to choose a different name instead of `v` for the downstream code, maybe just `v1` or `v_corrected`..

Also, the only place I see this later `v` used is at line 3287, where it could just as well be `root_v` - and in fact `root_v` is actually used just above at line  3285. I know this is only a nit or matter of taste, but it would help to reduce cognitive dissonance a bit and that's always good.

edreamleo:
@tbpassin Thanks, Thomas, for your review.  Yes, the business with 'v' is confusing, but I only wanted to touch case 1.

There is now only one path through case 1. Previously, there were three paths, all kinda alike, but each different.

PR #3971, opened 2024-06-16
PR: Fix various quirps, mostly reported by Félix

### LINKED ISSUE #3972: "save changes before closing?" doesn't handle 'yes'
Part of PR #3971.

edreamleo:
This PR concerns only desktop Leo. 

Félix and I now agree that LeoJS should support all Leo commands that make sense in vs-code. The reason: to maintain as much compatibility as possible with existing scripting practice.

**Fix mypy complaints**

- [x] cursesGui2.py: Fix a mypy complaint. Change 0 to True(!)

**EKR: Fix #3972.**

- [x] Revise `qt_gui.runAskYesNoCancelDialog`.

**EKR: remove unused code**

- [x] Remove the `renameBuffer` node. None of this code has ever been used.
- [x] Move the `sort-fields` command to the attic.
   This Emacs-style command has no unit test. It may never have worked.
   Original report: The `sort-field` code contains `w.insert(f"{int1}.0", f"{z[1]}\n")`, but the first arg should be an int.

**Félix reported all of the following in emails on June 15 and 16, 2024**

- [x] Remove dubious print statement in the `clean-all-lines` command.
- [x] Remove duplicate `@cmd('yank') decorator.
- [x] In the 'kill-kill-ring' command, the globalKillbuffer is spelled incorrectly with a lowercase 'b'. Thus this clears nothing.
- [x] Remove unused `_addPrefix` method and `_useRegex` ivar.
- [x] In `reformatSelection`, the `undoType` parameter is force-set weirdly at the start of that method.
- [x] Make `toggle-angle-brackets` undoable.

**Won't do**

- In `toggle-angle-brackets`, replace the `c.redrawAndEdit(p, selectAll=True)` at the end for a simpler redraw().
  Selecting all text is correct, regardless of undo.

**Improve testing scripts**

Trying to save time in the test scripts was foolish!

- [x] `full_test_leo.py` runs `beautify_all_leo.py` instead of `beautify_leo.py`.
- [x] `beautify_leo.py` checks *all* significant Leo files for changes.

### DIFF
--- a/leo/commands/bufferCommands.py
+++ b/leo/commands/bufferCommands.py
@@ -166,27 +166,6 @@ class BufferCommandsClass(BaseEditCommandsClass):
             w.seeInsertPoint()
             self.endCommand()
             c.recolor()
-    #@+node:ekr.20150514045829.12: *4* renameBuffer (not ready)
-    def renameBuffer(self, event: LeoKeyEvent) -> None:
-        """Rename a buffer, i.e., change a node's headline."""
-        g.es('rename-buffer not ready yet')
-        if 0:
-            self.c.k.setLabelBlue('Rename buffer from: ')
-            self.getBufferName(event, self.renameBufferFinisher1)
-
-    def renameBufferFinisher1(self, name: str) -> None:
-        self.fromName = name
-        self.c.k.setLabelBlue(f"Rename buffer from: {name} to: ")
-        event = None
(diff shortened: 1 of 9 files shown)


tbpassin (approved):

edreamleo:
@tbpassin Thanks for your review!

PR #3987, opened 2024-07-07
Do not use polyfill scripts from polyfill.io

LeSuisse:
The `polyfill.io` domain is under the control of a bad actor, see https://sansec.io/research/polyfill-supply-chain-attack

Removing it completely should be fine, it is no more needed for browsers that are severely outdated.

### DIFF
--- a/leo/plugins/viewrendered3.py
+++ b/leo/plugins/viewrendered3.py
@@ -1089,7 +1089,6 @@ REST = 'rest'
 RST = 'rst'
 
 ASCIIDOC_CONF_FILENAME = 'html5.conf'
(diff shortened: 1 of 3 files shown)


tbpassin:
This is only used in `@language latex` nodes.  I didn't know anyone even used them and I haven't ever tested them. After the backlog of PRs gets worked through, I'll make this polyfill change.

In the meantime, anyone who needs to can delete the line

`    <script src="{MATHJAX_POLYFILL_URL}"></script>`

from viewrendered3.py.

tbpassin (approved):

PR #3995, opened 2024-07-13
PR for 6.8.0-1

edreamleo:
The culprit: `pyproject.toml` *must* contain duplicates of the dependencies in `requirements.txt`.

**Notes**

- PyPi requires that the new file be `leo-6.8.0-1-py3-none-any.whl`
  PyPi only allows **build numbers** of the form `-1`.  `.post1` is **not** allowed.

**Required changes**

- [x] `PKG-INFO.txt`: Update version to `6.8.0.post1'.
- [x] `pyproject.toml`: Duplicate dependencies from `requirements.txt`.

**Other changes**

- [x] `MANIFEST.in`: Add comment.
- [x] `install_leo_locally.py`: Add `version` string. Don't clear pip caches.
- [x] Test that `install_leo_locally.py` installs all of Leo's dependencies.
- [x] Add pylint to both sets of dependencies.

This PR will change *only* those files that affect which files are included in the distribution.

All other changes will become part of Leo 6.8.1.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -12,9 +12,9 @@ global-exclude *.mypy_cache, *.mypy_stubs, *.gitattributes, *.htmlcov
 include *.py
 include *.TXT
 
-# Required: Do not omit!
+# No longer required.
 # obscure pip bug fix https://github.com/pypa/setuptools/issues/1694
-### include pyproject.toml
+#include pyproject.toml
 
 # Tweaks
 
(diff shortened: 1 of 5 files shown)


edreamleo:
@tbpassin @gatesphere The culprit was the missing requirements in `pyproject.toml`, not the "obscure pip bug".

Local testing is complete. The next step will be to upload to testpypi.

tbpassin (approved):
How about if pyproject.toml gets generated automatically from requirements.txt, using a new script to do so?  That would guarantee that the two requirements listings are in sync.

edreamleo:
@tbpassin I agree. A script to keep the two files in sync would be a good idea.

Issue #3988, opened 2024-07-10
Traceback leoColorizer.py and pygments\lexer.py


LewisNeal:
Running Leo 6.8.1-devel, build eb4ce4f641
Python 3.12.4, PyQt version 6.7.2
Windows 10 AMD64 (build 10.0.19045) SP0

This is definitely a Leo problem, as the reply to [pygments issue #2753](https://github.com/pygments/pygments/issues/2753) makes clear.

When searching with Find/F3 I received this traceback:
```console
Uncaught exception in Leo...
Traceback (most recent call last):
  File "N:\git\leo-editor\leo\core\leoColorizer.py", line 2823, in highlightBlock
    self.colorizer.recolor(s)  # Highlight just one line.
    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "N:\git\leo-editor\leo\core\leoColorizer.py", line 3243, in recolor
    self.mainLoop(s)
  File "N:\git\leo-editor\leo\core\leoColorizer.py", line 3104, in mainLoop
    for token, text in lexer.get_tokens(s):
  File "C:\Users\lewisneal\AppData\Local\Programs\Python\Python3\Lib\site-packages\pygments\lexer.py", line 270, in streamer
    for _, t, v in self.get_tokens_unprocessed(text):
  File "C:\Users\lewisneal\AppData\Local\Programs\Python\Python3\Lib\site-packages\pygments\lexer.py", line 804, in get_tokens_unprocessed
    yield from action(self, m, ctx)
               ^^^^^^^^^^^^^^^^^^^^
TypeError: PygmentsColorizer.at_language_callback() takes 3 positional arguments but 4 were given

_get_format_from_style KeyError(Token.Punctuation.Indicator)
_get_format_from_style KeyError(Token.Literal.Scalar.Plain)
```

edreamleo:
@LewisNeal What language was Leo trying to color when pygments crashed?

edreamleo:
Closed via PR #3999.

PR #3999, opened 2024-07-14
PR for #3988: crash in pygments colorizer

edreamleo:
See #3988.

**Long standing bugs that affect LeoJS**

- [x] Fix long-standing bug in `BaseColorizer.useSyntaxColoring`.

**Newly-discovered bugs that do not affect LeoJS**

- `@nocolor` could crash pygments.
- The pygments colorizer does not recolor when an `@language` directive changes.
  The `JEditColorizer` class works as desired via the Leo-specific matcher `match_at_language`.
  Similarly, for matchers for `@killcolor`, `@color`, etc.

**Other changes**

- [x] Make `PatchedLexer` a subclass of `pygments.DelegatingLexer`.
   This change *might* be the only change needed to fix the original issue.
- [x] Add a guard to `BaseColorizer.setTag`.
- [x] Leo's `recolor` command now calls `colorizer.force-recolor` as a hook for the pygments colorizer.
   This forces a full recolor. It's not as smooth as the jEdit colorizer, but it's an improvement.
- [x] All coloring unit tests now test both the jEdit and pygments colorizers.

### DIFF
--- a/leo/core/leoColorizer.py
+++ b/leo/core/leoColorizer.py
@@ -18,6 +18,7 @@ import warnings
 # Third-part tools.
 try:
     import pygments  # type:ignore
+    from pygments.lexer import DelegatingLexer
 except ImportError:
     pygments = None  # type:ignore
 #
@@ -3145,6 +3146,7 @@ class PygmentsColorizer(BaseColorizer):
         """Colorize the name only if the language has a lexer."""
         from pygments.token import Name
         language = match.group(2)
+        g.trace('match', repr(match))
         # #2484:  The language is known if there is a lexer for it.
         if self.pygments_isValidLanguage(language):
             self.language = language
@@ -3175,9 +3177,10 @@ class PygmentsColorizer(BaseColorizer):
 
         from pygments.token import Comment  # type:ignore
         from pygments.lexer import inherit  # type:ignore
+        
+        g.trace('language', language, lexer.__class__.__name__)
 
-
-        class PatchedLexer(lexer.__class__):  # type:ignore
+        class PatchedLexer(DelegatingLexer, lexer.__class__):  # type:ignore
 
             leo_sec_ref_pat = r'(?-m:\<\<(.*?)\>\>)'
             tokens = {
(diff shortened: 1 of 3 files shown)


tbpassin (changes requested):
[I read the issue that ekr filed with the pygments projects so I know that some of these changes are about.]
_leoColorizer.py_
1. Line 18, typo in comment. Should read `# Third-party tools`.
2. Line 828, typo in comment.  Change _enabgled_ to read _enabled_.
3. Line 3187: `class PatchedLexer(DelegatingLexer, lexer.__class__):`
This is probably just my ignorance about multiple inheritance, but now calls that used to be handled by _lexer_ are now supposed to be handled by _DelegatingLexer_. What makes that happen? Is it just because _DelegatingLexer_ comes first in the constructor?

edreamleo:
@tbpassin The last rev fixes both typos.

Point 3: The code just follows the pattern [here](https://pygments.org/docs/lexerdevelopment/#delegating-lexer). I know nothing more :-) The revised unit tests cover all languages known to pygments w/o crashes, so that probably suffices.

tbpassin (commented):
_leoColorizer.py_, line 3040 :
 `def force_recolor(self)`

I don't see how this method actually performs a recolor.  It picks up any language or color directive changes, but where does it actually do recoloring?  Or does it just re-initialize the colorizer for the next time it gets triggered? If it doesn't actually recolor, it should be named more clearly.  

I presume that a recolor will happen on any change to  a node's body.  So changing an `@language` or `@color` directive would trigger a recolor.  Therefore I can understand that the method only needs to reset the language or color states of the colorizer.  But in that case it's not actually doing any recoloring.

Also, if the language were to change, wouldn't that require the colorizer to use a different lexer?  I don't see where that happens.

edreamleo:
@tbpassin I don't know exactly why `force_recolor` works, but it does. It's a slight modification of `pyg_c.recolor`.

The pygments colorizer is tricky. It contains numerous hacks, including code borrowed (with attribution) from the Jupyter project. Please don't ask me to explain the details :-)

edreamleo:
@tbpassin Please approve the PR. Merging it will help further testing.

tbpassin (approved):

edreamleo:
@boltex The changes to `BaseColorizer.useSyntaxColoring` are the only changes that should affect LeoJS.

Issue #3998, opened 2024-07-14
Add script to sync requirements.txt and pyproject.toml


edreamleo:
The script would update `pyproject.toml` using `requirements.txt`.

edreamleo:
Closed via PR #4000.

PR #4000, opened 2024-07-14
PR for #3998: use script to sync requirements files.

edreamleo:
See #3998.

- [x] Indent two lines in `requirements.txt`.
- [x] Use script to update `pyproject.toml` from `requirements.txt`.

Here is the Leonine script that updated `pyproject.toml`.
```python
"""Update the section `pyproject.toml: dependencies` from requirements.txt."""
g.cls()
req_h = '@file ../../requirements.txt'
req_p = g.findNodeAnywhere(c, req_h)
assert req_p, req_h
proj_h = g.angleBrackets(' pyproject.toml: dependencies ')
proj_p = g.findNodeAnywhere(c, proj_h)
assert proj_p, proj_h

# Indent all non-blank lines with 2-blanks.
results = ['dependencies = [']
for s in g.splitLines(req_p.b):
    ss = s.strip()
    if not ss:
        results.append('')
    elif ss.startswith('@'):
        continue  # Don't change directives.
    elif ss.startswith('#'):
        results.append(f"  {ss}")
    else:
        i = ss.find('#')
        if i > -1:
            s2 = ss[:i].strip().replace('"', "'")
            # Ensure spaces around '=='.
            if '==' in s2 and ' == ' not in s2:
                s2 = s2.replace('==', ' == ')
                s3 = f'  {ss[len(s2):]}'
            else:
                s3 = ss[len(s2):]
            results.append(f'  "{s2}",{s3}')
        else:
            results.append(f'  "{ss}",')
results.append(']\n\n')
# Print the changes.
for s in results:
    print(s)
# Make the changes!
new_b = '\n'.join(results)
if new_b != proj_p.b:
    print('Saving:', proj_p.h)
    proj_p.b = new_b
    c.redraw()
    c.save()
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -278,52 +278,54 @@ readme = "README.md"
 #@+node:ekr.20240713074535.1: ** << pyproject.toml: dependencies >>
 dependencies = [
 
-    # For build devs.
-    "build>=1.2.1",
-    "twine>=5.1.0",
+  # Warning: these dependencies should match those in pyproject.toml.
 
-    # For mypy...
-    "mypy",
-    "mypy-extensions",
-    "typing_extensions",
-    "types-docutils", 
-    "types-Markdown",
-    "types-paramiko",
-    "types-PyYAML",
-    "types-requests",
-    "types-six",
+  # For build devs.
+  "build>=1.2.1",
+  "twine>=5.1.0",
 
-    # General packages, including various plugins and commands...
-    "asttokens",        # For unit tests.
-    "beautifulsoup4",   # For link testing.
-    "black",            # For unit tests.
-    "docutils",         # various plugins and commands.
-    "flexx",            # leoflexx.py plugin.
-    "meta",             # livecode.py plugin.
-    "pyenchant",        # The spell tab.
-    "pytest",           # For coverage testing.
-    "pytest-cov",       # For coverage testing.
-    "pyflakes",         # pyflakes command.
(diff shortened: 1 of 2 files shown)


edreamleo:
@boltex @tbpassin @gatesphere This PR is ready for review. I was surprised by how many missing items were in `pyproject.toml`.

tbpassin (approved):
pyproject.toml and requirement.txt  seem good, but see my comments on the script in the "Conversions" thread.

tbpassin:
Please make the following improvements to the script:

1. Lines 16, 17:
```
    elif ss.startswith('@'):
        pass  # Don't change directives.
```
Change _pass_ to _continue_:
```
    elif ss.startswith('@'):
        continue  # Don't change directives.
```

2. 
a. In the `for` loop starting at line 12, omit the trailing `\n,` for all appends;
b. collect the results at line 38 with:

`proj_p.b = '\n'.join(results)`

This will guarantee that the "[" character will appear as the first character on a line without having to make sure all the appends end with the right number of `\n`s.

edreamleo:
@tbpassin What do you mean by the "Conversions" thread?

edreamleo:
@tbpassin I made the changes you requested, plus a few more.

tbpassin:
> What do you mean by the "Conversions" thread?

Sorry, the "Conversations" thread - the name of the tab we are posting on.

Issue #3994, opened 2024-07-13
A git diff script can corrupt the .leo file


edreamleo:
See PR #4001.

The script `script: diff-branches/revs (all files)` can corrupt the .leo file if it reports read errors:
```python
from leo.commands import editFileCommands as efc

efc.GitDiffController(c).diff_two_revs(
    rev1='6.7.8',  # Old
    rev2='devel',   # New
)
```
There were several hundred diffs.

PR #4001, opened 2024-07-15
PR for #3994: git-diff can corrupt an outline

edreamleo:
See #3994.  The culprit was diffing `.pdf` and `.inv` (sphinx inventory) files.

- [x] Improve 'bad .leo file' message.
- [x] Add c.validateOutlineXML.
- [x] Call c.validateOutlineXML whenever diffing files.
    This method dumps the bad outline to `~/BAD-{c.shortFileName()}.txt`.
    This dump is about the only way to figure out why the outline is corrupt.
- [x] Suppress diffing  `.pdf` and `.inv` (sphinx inventory) files.

**Discussion**

It would be safer to specify the file types that *will* be diffed, rather than those that *won't*.
However, I am willing to take the small extra risk.

### DIFF
--- a/leo/commands/editFileCommands.py
+++ b/leo/commands/editFileCommands.py
@@ -1246,7 +1246,7 @@ class GitDiffController:
     def finish(self) -> None:
         """Finish execution of this command."""
         c = self.c
-        if c.validateOutlineXML(dump=True):
+        if c.checkOutlineXML(dump=True):
             # All is well.
             c.selectPosition(self.root)
             self.root.expand()
@@ -1259,7 +1259,7 @@ class GitDiffController:
             last = c.lastTopLevel()
             c.redraw(last)
             # Re-validate.
-            if not c.validateOutlineXML(dump=False):
+            if not c.checkOutlineXML(dump=False):
                 g.es_print('The outline is *still* invalid!', color='red')
                 g.es_print('Do not save the outline!', color='red')
         c.treeWantsFocusNow()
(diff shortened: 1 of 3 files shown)


tbpassin (changes requested):
_leoCommands.py, lines 2250 ff_
1. "Validate" in XML means something specific: a) the file is well-formed, and b) it conforms with a specified schema.  This new _validateOutlineXML()_ method is basically doing a well-formedness check, not a validation. A better name would be _checkOutlineXML()_
2. Line 2254: `translate_dict = {z: None for z in range(20) if chr(z) not in '\t\r\n'}`

The allowed characters in XML 1.0 are 
```
Char | ::= | #x9 \| #xA \| #xD \| [#x20-#xD7FF] \| [#xE000-#xFFFD] \| [#x10000-#x10FFFF] | /* any Unicode character, excluding the surrogate blocks, FFFE, and FFFF. */
```
(from https://www.w3.org/TR/xml/#charsets)

3. Line 2267: Let's write the dump to the `~/.leo ` directory, NOT `~/`.

edreamleo:
@tbpassin Thanks for your comments. I'll make all requested changes.

I wonder whether the incomplete translate table might have been one of the causes of the bug.

tbpassin:
> I wonder whether the incomplete translate table might have been one of the causes of the bug.

I wondered too, but I'm still not sure.  The original version would have zeroed out more characters than necessary.  That doesn't seem like it should be harmful.  I'm wondering about character encoding when, e.g., a .pdf file is treated as text instead of binary when converted to a string.

OTOH, I'm not sure whether it's the translated version that's saved as the outline (I didn't go back and check the code and I don't quite remember).  If not, there could definitely be violations of well-formedness.

edreamleo:
@tbpassin Recent revs changed points 1 and 3 as you requested.

On second thought, point 2 opens a can of worms that is unrelated to the original bug. I would rather not mess with it, but if I must it should have its own issue :-)

Issue #3990, opened 2024-07-10
TODO icons (and possibly all uas?) are not persisted on @edit notes


gatesphere:
Adding a todo icon to an `@edit` node (via right click->task->status) adds the icon to the current working display of the file.  Saving, quitting, and reopening the .leo workbook shows that this information did not persist.

I assume this is because `@edit` nodes read their entire state from external files on disk when loading, meaning this data is simply lost because there's no specific code to save it in the outline.  Perhaps this is intended behavior, but it is inconsistent at best -- icons (and other ua info) is properly persisted on other file node types, and this presents a confusion point for the user.

("Just use `@clean`" isn't always an option, `@edit` nodes fill a niche that I have very frequent use for.)

tbpassin:
I set an "Enhancement" label on this issue, but let me know if you think it should be "Bug" instead.  If a label and milestone date aren't set, Edward might not notice a new issue for a time.

edreamleo:
Closed via PR #4005.

gatesphere:
Thank you!

boltex:
@edreamleo I applied this and all the last commits since 6.8.0 to LeoJS 😁

edreamleo:
@gatesphere You're welcome. It was good to fix this issue.

edreamleo:
@boltex Thanks for keeping LeoJS in sync. It's amazing that these bugs have stuck around in Leo for decades.

PR #4005, opened 2024-07-16
PR for #3990: uAs on @edit nodes

edreamleo:
See #3990.  This PR is a significant change to Leo's write code.

- [x] Generalize `fc.compute_attribute_bits`. There is no reason to require that `c.p` should have children!
  Hand tests confirm that to-do icons now persist on `@edit` nodes.

The new code probably affects LeoJS.

### DIFF
--- a/leo/core/leoFileCommands.py
+++ b/leo/core/leoFileCommands.py
@@ -2052,7 +2052,7 @@ class FileCommands:
         if forceWrite or self.usingClipboard:
             v.setWriteBit()  # 4.2: Indicate we wrote the body text.
 
-        attrs = fc.compute_attribute_bits(forceWrite, p)
+        attrs = fc.compute_attribute_bits(p)
         # Write the node.
         v_head = f'<v t="{gnx}"{attrs}>'
         if gnx in fc.vnodesDict:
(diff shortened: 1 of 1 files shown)


tbpassin (changes requested):
_fileCommands.py_
1. Line 2057 "
 ` v_head = f'<v t="{gnx}"{attrs}>'`

I don't see how this can produce well-formed XML.  The returned values from `fc.putDescendentVnodeUas()` look like this:

`descendentVnodeUnknownAttributes="7d710058010000003071017d7102580200000061317103580e0000007468697320697320612074657374710473732e"`

There is no leading space, so there will be no space after `t="{gnx}". Also, at around line 2087, all attributes are accumulated without spaces between them, by means of `"".join(...)`.

In XML, attributes must be whitespace-separated.

edreamleo:
Somehow you have misread the code. In a recent test file the `<v>` elements is:

`<v t="ekr.20240716112233.1" descendentVnodeUnknownAttributes="7d0[big snip]"><vh>@edit c:\test\gnx_test.py</vh></v>`

The corresponding outline node contains an attached todo icon, verifying that the bug has been fixed.

There is a space between the gnx and the attribute.  In any event:

- Leo's read code (in the node `fast.readWithElementTree & helpers`) can still read `.leo` files using `xml.etree.ElementTree`.
- This PR does not change Leo's file format in this respect, so any problem would be a separate issue.

tbpassin (approved):
OK; I still don't see where the space comes from, but the outline wouldn't load if it were not well-formed.  Anyway, good to check and confirm.

edreamleo:
@tbpassin Yes, it's always good to check :-) Thanks for your approval.

tbpassin:
Weren't we going to start calling them "User Attributes" instead of "Unknown Attributes"?

edreamleo:
@tbpassin Whatever *we* call them, the code isn't going to change :-)

Issue #4006, opened 2024-07-16
CheatSheet is still linking to leoeditor.com


tbpassin:
Change `http://leoeditor.com` to `https://leo-editor.github.io/leo-editor`.

edreamleo:
Closed via PR #4008.

PR #4008, opened 2024-07-17
PR for #4006: url's in CheatSheet.leo

edreamleo:
- [x] Fix most urls.

**Important extras**

- [x] Revise `pyplot_backend.py` so it works with Qt6. See the [Embedding in Qt](https://matplotlib.org/stable/gallery/user_interfaces/embedding_in_qt_sgskip.html) page.
- [x] Tweak pyplot-related code in `viewrendered.py` and `CheatSheet.leo`.
   The revised scripts work properly both in the VR pane and as stand-alone scripts.
 - [x] Mention that `@pyplot` has few (any?) over running a pyplot script directly in the body pane.
   Doing so shows the plot in a separate window. That's often a better alternative.

### DIFF
--- a/leo/doc/CheatSheet.leo
+++ b/leo/doc/CheatSheet.leo
@@ -1485,7 +1485,9 @@ Here is some python code::
 Here is the main page.  It contains links to individual lessons.
 https://github.com/leo-editor/leo-editor/issues/816
 </t>
-<t tx="ekr.20180405105451.1">These will work only if numpy and pyplot are installed.</t>
+<t tx="ekr.20180405105451.1">These will work only if numpy, matplotlib and pyplot are installed.
+
+The viewrendered.py plugin must also be enabled.</t>
 <t tx="ekr.20180405105738.1"></t>
 <t tx="ekr.20180503073939.1"></t>
 <t tx="ekr.20180503073954.1">@language python
--- a/leo/doc/CheatSheet.leo
+++ b/leo/doc/CheatSheet.leo
@@ -102,6 +102,7 @@
 <v t="ekr.20180405093127.32"><vh>reStructureText text example</vh></v>
 </v>
 <v t="ekr.20230716045242.1"><vh>Clickable links</vh></v>
+<v t="ekr.20180405105451.1"></v>
 </vnodes>
 <tnodes>
 <t tx="ekr.20131019184243.16683">This Cheat Sheet contains a summary of many of Leo's important features.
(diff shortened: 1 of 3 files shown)


edreamleo:
@tbpassin It's obvious that few if any people use `@pyplot` nodes.

And why should they? Running pyplot scripts as Leo scripts works well.

tbpassin:
My issues with pyplot nodes weren't about qt5 vs qt6.  It's been several years but I remember the gist,  I think. One problem was that if both VR and VR3 were in use and VR changed the pyplot backend, then it would be changed for VR3 whether or not I wanted it that way (and vice-versa).  I devised a set of flags that let either plugin know if the backend had already been changed, but that was a kind of hack.

Another thing that didn't work for me was that once the backend had been changed then rendering any pyplot code in a non-@pyplot node would go to the new canvas willy-nilly even if I didn't want that.

A third issue was that once the backend had been changed there was no way to change it back.  No doubt a method could be devised.  But then, e.g., VR wouldn't know if VR3 had changed it back.

In addition, since node's code didn't need to import pyplot, etc., you couldn't simply paste it into an external file and have it work.  You needed to add the right imports, etc.  Not hard, but just another thing to remember.

In the end I decided not to support @pyplot nodes.  It's easy enough to run them (with the right pyplot imports) with CTRL-B.

I'd be willing to reconsider but I want to come up with a new operations concept for how they should work. I especially don't want some VR3 action to change how VR renders its plots, or the other way around.

tbpassin:
> And why should they? Running pyplot scripts as Leo scripts works well

Just so.

edreamleo:
@tbpassin Thanks for your comments. I agree: `@pyplot` nodes aren't very useful.

Are there any changes you would like me to make?

tbpassin:
> Are there any changes you would like me to make?

Thanks, I appreciate the offer but I don't think so. VR3 for an @pyplot node just emits a message "@pyplot nodes are not implemented". I think the imports for it are still there, but never mind. I will leave things alone for now. 

I have a few non-VR3 scripts that do plotting.  One of them finds the data in a body and plots it. The body can look like this:
```
This is a test

1 1
2 4
3 9
4 16

blah blah

5  25
6  36
```

Another can do the same from the clipboard.  The script can use either 1-column or 2-column data.  With these scripts,  you can do a quick-and-dirty plot of tabular data, which you might have copied from some web page. Scripts like this reduce any need for @pyplot nodes.

edreamleo:
@tbpassin I'm going to merge this. We'll deal with any nits later.

PR #4011, opened 2024-07-18
'tnodeList' uncommented as attribute to exclude from being unpickeld with resolveUa. Also deleted extraneous nativeVnodeAttributes object duplicate.

### LINKED ISSUE #4010: Opening old Leo files which had 'tnodelist' entries leave 'ghosts' of empty UAs when saving
Opening old leo files works, and saving them removes the extraneous tnodelists content. But it leaves 'empty' tnodelist UAs on almost each nodes.

I think when encountering tnodelists (which are not used when loading anyways) should be specifically ignored instead of trying to unpickle them (which produces a catched exception, which defaults to an "" empty string UA) thus remaining forever haunting the Leo file with the ghosts of past tnodelists attributes 💀 👻 

Maybe that line was commented by mistake (The commit message for this is that all mentions and uses were removed, indeed tnodelist are not used, but being in that list will skip them if encountered in old Leo files) :
image

Also, that nativeVnodeAttributes tuple is also re-created, as-is, in FileCommands, but never used. It should be removed entirely in that class.

I'll make a fix and a pull request 😄

boltex:
Fixes #4010

I tested this by re-opening my old files and re-saving them under a different name and comparing the XML and making sure they also opened properly in Leo. 

(I'll fix LeoJS in the same way very soon to also support old Leo files)

I also made sure to run unit tests this time :) they all pass.

### DIFF
--- a/leo/core/leoFileCommands.py
+++ b/leo/core/leoFileCommands.py
@@ -583,7 +583,6 @@ class FileCommands:
         """Ctor for FileCommands class."""
         self.c = c
         self.frame = c.frame
-        self.nativeTnodeAttributes = ('tx',)
         self.initIvars()
     #@+node:ekr.20090218115025.5: *4* fc.initIvars
     def initIvars(self) -> None:
--- a/leo/core/leoFileCommands.py
+++ b/leo/core/leoFileCommands.py
@@ -84,12 +84,13 @@ class BadLeoFile(Exception):
 #@+node:ekr.20180602062323.1: ** class FastRead
 class FastRead:
 
+    # Used to exclude attributes from being unpickeld as UAs with resolveUa.
     nativeVnodeAttributes = (
         'a',
         'descendentTnodeUnknownAttributes',
         'descendentVnodeUnknownAttributes',
         'expanded', 'marks', 't',
-        # 'tnodeList',  # Removed in Leo 4.7.
+        'tnodeList',  # Removed in Leo 4.7.
     )
 
     def __init__(self, c: Cmdr, gnx2vnode: dict[str, VNode]) -> None:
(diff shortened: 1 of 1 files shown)


tbpassin (approved):
1. Line 93 - The trailing comment is now incomplete. Either remove it or change to read something like:
`'tnodeList',  # Removed in Leo 4.7, restored in Leo 6.8.1`

edreamleo (changes requested):
I agree with Thomas: the comment should mention the full history of `tnodeList`.

This is a significant change to Leo's code.

tbpassin:
Will this change make it impossible to read pre-4.7 outlines?

boltex:
@tbpassin @edreamleo  🤔  Hmm,... I suspected that this comment **"# Removed in Leo 4.7"** was marking the deprecation of the tnodelist attribute itself in the leo file format, and not the point where the 'tnodelist' entry in nativeVnodeAttributes was commented out.

So I verified the line's history in the past commits. And indeed, this comment was not made when that line was commented out, it was about marking when the 'tnodelist' became historically retired and irrelevant to the read algorithm. 

So the comment should stand as-is. 

@tbpassin you asked:  **"Will this change make it impossible to read pre-4.7 outlines?"**  This changes nothing in that regard: the tnodelist attribute would be totally ignored regardless.

Note: the tnodelist attribute was plain text list of gnx, not a pickled data string meant to be a UA at all.

As explained here https://github.com/leo-editor/leo-editor/issues/4010 I just make it explicit and obvious with this modification instead of relying on the un-pickle process to error out and trigger the 'catch' of this exception and produce an 'empty string' lingering ghost of a UA which would have haunted the outline forever. :)

edreamleo (approved):

edreamleo:

Issue #4009, opened 2024-07-17
Closing a tab with open VR pane causes errors on Win10


gatesphere:
See PR #4014.

My env:

```
Leo 6.8.1-devel, devel branch, build e21cd1ebe4
2024-07-16 16:46:21 -0500
Python 3.10.6, PyQt version 6.6.3
Windows 10 AMD64 (build 10.0.19045) SP0
```

Repro steps:

- Open new .leo file (doesn't need to be saved) in a new tab
- Type some stuff into the body
- Alt-0 to open vr pane
- Close the new .leo file tab while vr pane is visible, don't save
- Get repeating errors in log and console:

```
Traceback (most recent call last):

  File "<redacted>\repos\leo-editor\leo\plugins\viewrendered.py", line 805, in update
    sb = w.verticalScrollBar()

RuntimeError: wrapped C/C++ object of type QTextBrowser has been deleted
```

(This repeats several times, about once a second, and continues until Leo is closed entirely.)

gatesphere:
Leo is definitely still usable in this instance, so I hesitate to add the 'Critical' tag, but this is not a *fun* experience.

Cleanup of the VR pane object appears to be missing when a tab is closed.

gatesphere:
Confirmed this behavior also happens on same commit on my Arch Linux environment, so this is not a Windows-specific bug.

tbpassin:
I just repo'ed the bug in Win10.  Interestingly it didn't happen with VR3. Maybe that's because VR3 doesn't know how to set the vertical scrollbar yet...

There's probably a `deleteLater()` that needs to happen somewhere.  The bug may have surfaced with a particular version of Qt.  It could have been happening all along but not getting announced, perhaps.

edreamleo:
Closed via PR #4014.

PR #3984, opened 2024-06-27
Make Images Display Correctly For rest, md, and asciidoc

tbpassin:
The previous code seems to have been mangled somehow and no longer worked.  This PR fixes it.  Relative and absolute image files display and export correctly for all three languages (rest, md, asciidoc).

PR #4014, opened 2024-07-20
PR for #4009: crash in VR plugin

edreamleo:
See #4009.

- [x] Fix the crash in VR plugin.
- [x] Fix the corresponding crash in the VR3 plugin.

**Extras**

- [x] Add jedi to requirements.

### DIFF
--- a/leo/plugins/viewrendered3.py
+++ b/leo/plugins/viewrendered3.py
@@ -1558,7 +1558,7 @@ def getVr3(event):
         controllers[h] = vr3 = viewrendered(event)
     return vr3
 #@+node:TomP.20191215195433.16: ** vr3.Commands
-#@+node:TomP.20191215195433.18: *3* g.command('vr3') (**weird**)
+#@+node:TomP.20191215195433.18: *3* g.command('vr3') (**weird effect**)
 @g.command('vr3')
 def viewrendered(event):
     """Open render view for commander"""
--- a/leo/plugins/viewrendered3.py
+++ b/leo/plugins/viewrendered3.py
@@ -1469,7 +1469,7 @@ def onClose(tag, keys):
     h = c.hash()
     vr3 = controllers.get(h)
     if vr3:
-        g.trace(f"VR3: delete {vr3}")  ###
+        # g.trace(f"VR3: delete {vr3}")
         c.bodyWantsFocus()
         del controllers[h]
         vr3.deactivate()
--- a/leo/plugins/viewrendered3.py
+++ b/leo/plugins/viewrendered3.py
@@ -1469,10 +1469,10 @@ def onClose(tag, keys):
     h = c.hash()
     vr3 = controllers.get(h)
     if vr3:
+        g.trace(f"VR3: delete {vr3}")  ###
         c.bodyWantsFocus()
         del controllers[h]
         vr3.deactivate()
-        vr3.deleteLater()
 #@+node:TomP.20191215195433.13: *3* vr3.show_scrolled_message
 def show_scrolled_message(tag, kw):
     """Show "scrolled message" in VR3.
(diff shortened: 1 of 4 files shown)


edreamleo:
@tbpassin @gatesphere I am going to merge this right away.

Please feel free to comment. I can always make changes later.

tbpassin (commented):
1. I thought that deleteLater() was needed to delete Qt C++ stuff.  When will this happen without it?

2.  The Jedi requirement has nothing to do with the PR.  In the future, please don't mix unrelated changes in one PR like this.

edreamleo:

1. Experiments show that closing an outline suffices to delete all widgets, including the VR pane.
  I'm not sure why I ever thought deleteLater was needed.
  Otoh, the "starting from scratch" protocol will have to take care that everything gets deleted as desired.
2. You're right, of course, but sometimes I don't think a little bit extra is so bad when I'm going to merge a PR quickly.

tbpassin:
Will you *PLEASE* merge my pending changes for VR3 before you go changing anything else.  I already had to recreate them once because they were overlooked:

https://github.com/leo-editor/leo-editor/pull/3984

edreamleo:
@tbpassin Done.

In the future _PLEASE_ remember to assign the current milestone to PRs. Otherwise I may never know they are pending.

tbpassin:
Oops, I thought I had done that.

That reminds me, sometimes we get issues or PRs that don't have a label or milestone assigned.  Sometimes the submitter doesn't realize they should be.  But I think that ordinary users cannot add labels or milestones for themselves.  When I notice new ones I try to remember to make assignments.  But it's easy to forget.

I'm not sure how to resolve this.  If everyone can assign or change labels and milestones there could be chaos.  If they can't, their submission can get overlooked.

edreamleo:
@tbpassin I think we devs just have to remember :-)

leo/plugins/viewrendered3.py
```diff
@@ -1558,7 +1558,7 @@ def getVr3(event):
         controllers[h] = vr3 = viewrendered(event)
     return vr3
 #@+node:TomP.20191215195433.16: ** vr3.Commands
-#@+node:TomP.20191215195433.18: *3* g.command('vr3') (**weird**)
+#@+node:TomP.20191215195433.18: *3* g.command('vr3') (**weird effect**)
```
gatesphere, line 1561: Curious what 'weird effect' means here.

gatesphere (commented):
General thoughts line up with @tbpassin, but the end result is that this appears to prevent the repeating error in log & console on both platforms I'm able to test on.

If closing an outline truly destroys all widgets, then probably good to go.

Agreed that mixing unrelated things into a PR is generally not a best practice.

Thanks for the fix!

gatesphere:
> @tbpassin I think we devs just have to remember :-)

Remember what -- to occasionally scan through the list of issues without milestone tags?

I'm hesitant to apply milestones because I don't know @edreamleo's master plan for any given release.  As a result, some of my issues linger for a bit before they're noticed.  I suppose I could get over that and just be overzealous with them, and let Edward sort it out on his end.

edreamleo:
@gatesphere Hi Jake. Thanks for your comments.

I'll try to remember to scan for PRs without milestones. I wouldn't worry about assigning milestones yourself. I often change them.

edreamleo:
@gatesphere The weird effect is that on my Windows 11 machine enabling the VR3 plugin brings up the GE Force experience window!!! I have no idea how that could possibly happen.

gatesphere:
> @gatesphere The weird effect is that on my Windows 11 machine enabling the VR3 plugin brings up the GE Force experience window!!! I have no idea how that could possibly happen.

Pardon the expletive, but wtf?  That's a new one by me...   Then again, after this week, I'll believe anything.

edreamleo:
@gatesphere Whiskey Tango Foxtrot indeed.

Issue #4016, opened 2024-07-21
Create new official layouts


edreamleo:
See PR #4033.

The new layouts will include at least:
- The four layouts Jake mentioned [in these gists](https://gist.github.com/gatesphere/82c9f67ca7b65d09e85208e0b2f7eca1). 
- [This layout from Martin Liebenberg](https://mail.google.com/mail/u/0/#inbox/WhctKLbFbCtGVPQHtxWDKJmCFnmQNGgqQHplxBJXCMvsWVkGKjPznWrRpvlvVVJnKBrgSLG).

This is a significant project. As a side effect, the VR and VR3 plugins will see major changes. See [this post](https://groups.google.com/g/leo-editor/c/ZVlVBEa_-zo/m/8wc9lvZBAAAJ).

The `toggle-split-direction` command will likely apply only when Leo's original layout is in effect.

I shall not delay Leo 6.8.1 for this issue. The `ekr-4016-layouts` branch contains the work.

edreamleo:
Completed via PR #4033.

PR #4017, opened 2024-07-21
PR for #4016: new layouts

edreamleo:
See #4016.

This PR contains the first phase of the project.

**Overview**

- The VR plugin no longer knows anything about the layout in which they are embedded.
  VR calls `dw.insert_vr_frame` when the plugin creates its controller (a `QFrame`).
- At present, the `DynamicWindow` class supports only a fixed set of layouts.
- Later, an `@data` node might *describe* all layouts using a simple **layout description language** (details omitted).
- `dw.reloadSettings` calls the new `dw.recreateMainWindow` which does a `restart-leo` command if the layout has changed.

**Constructing layouts**

`dw.construct` now creates Leo's main window as follows:
```python
self.setMainWindowOptions()
self.createCentralWidget()
main_splitter, secondary_splitter = self.create_layout()
self.createMiniBuffer(self.centralwidget)
self.createMenuBar()
self.createStatusBar(self)
self.createIconBar()
```
This "boilerplate" creates all of Leo's standard widgets. `dw.create_layout` creates the "variable" part of Leo's layout:
- `.ui` files will specify only a *small part* of Leo's main window.
  The existing `qt_main.ui` and `qt_main_2.ui` files specify way too much.
- *Splitters are optional*: `dw.create_layout` may return None for either splitter.

**Generalize VR and VR plugins**

- [x] Simplify VR plugin:
  - Replace the `controllers` global with `c.vr`.
  - Use the new `getVr` function everywhere.
  - The VR plugin no longer handles layout-related code.
- [ ] Simplify VR3 plugin:
  - Replace the `controllers` global with `c.vr3`.
  -  Use `getVr3` everywhere. It calls `dw.insert_vr_frame` when creating the `ViewRenderedController3` class.
  - The VR3 plugin no longer handles layout-related code.

**Generalize layouts**

- [x] `leoSettings.leo`: add `@string qt-layout-name` setting.
- [x] `dw.createMainWindow` calls `dw.create_layout`.
- [x] `dw.create_layout` creates the layout given by `@string qt-layout-name`, defaulting to `legacy-layout`.
- [x] Add `dw.main_splitter` and `dw.secondary_splitter` as official ivars.
- [x] Replace `@bool big-outline-pane` by a `big-tree` layout
- [x] `dw.reloadSettings` changes `reload-settings` to `restart-leo` if the `@string qt-layout-name` setting has changed.
- [x] Add the VR/VR3 panes in all layouts.
- [ ] Support all layouts given in the [gists](https://gist.github.com/gatesphere/82c9f67ca7b65d09e85208e0b2f7eca1).

**Related extras**

- [x] Fix [#4028](https://github.com/leo-editor/leo-editor/issues/4028) for VR.


tbpassin (changes requested):
With only a single look at the VR3 changes I can say that the changed VR3 code is not acceptable.  Specifically, the function `getVr3` does two things where it should do one.  The function should return the vr3 object, having created it if necessary. **Putting it into a new dw is WRONG**. If the calling program wishes to do so, it can.  But the call is intended only to return an instance.  It's not even a given that I want to put it into a DW at all.

edreamleo:
I am closing this PR so as not to pollute VR3.

Issue #4012, opened 2024-07-18
RunYesNoDialog Fails


tbpassin:
See PR #4020.

[Later] This has turned out to be caused by the `RunYesNoDialog()` class failing to return the correct values.]

If you edit a node that is not an external file with an external editor and save it in that editor, Leo detects that the outline has changed and puts up the right dialog about reloading it, but does not actually reload it. This could happen if a second Leo window has been opened and the node edited in it, then saved.  It could also happen if the node has been opened for editing by a script or menu item having a body like this:
```
ext:
kind:   subprocess.Popen
arg:    "C:\Program Files\EditPlus\editplus.exe"
```
To reproduce:
Open Leo and create a new node in the workbook.  Open a second Leo session, navigate to the same node, and add some text.  Save the outline.  Click on the first Leo window.  Notice that the _File Changed, Reload It?_ dialog opens.  Click on "yes".  Observe that the node has not changed.

This functionality used to work.  I don't know when it stopped working but I noticed it today.
```
Leo 6.8.0, 6.8.0 branch, build 4599d84609
2024-06-25 03:38:07 -0500
Python 3.12.3, PyQt version 6.7.2
Windows 10 AMD64 (build 10.0.19045) SP0
```

tbpassin:
The `revert` command isn't reverting the outline to the one on disk.  This is probably the reason.  To test:

1. Save outline; 
2. Type a new line of text;
3. Without saving, issue the `revert` minibuffer command.

Actual result: the added line remains.
Expected result: the added line disappears.

tbpassin:
This is because in _c_file.revert()_ the dialog for reverting is returning "cancel" instead of "yes" when the "Yes" button is clicked:

```
    reply = g.app.gui.runAskYesNoDialog(
        c, 'Revert', f"Revert to previous version of {fn}?")

```

tbpassin:
Getting closer: the dialog is returning `2` but the code expects a value of `0`.

tbpassin:
A test script:
```
gui = g.app.gui

title = 'Test Dialog'
msg = 'Try Something'
result = gui.runAskYesNoDialog(c, title, msg)
g.es(result)
```

tbpassin:
The two dialog buttons return 2 and 3, not 1 and 2.

Presumably the optional buttons `yes_all` and `no_all` also return unexpected values but I haven't checked that.

This could be an issue with other dialogs.

tbpassin:
[...]

PR #4018, opened 2024-07-21
Make runAskYesNoDialog() work again, using direct query for which button was clicked.

tbpassin:
Fixes Issue https://github.com/leo-editor/leo-editor/issues/4012

Note that other similar dialogs may be broken the same way.  This PR only fixes this specific dialog.

### DIFF
--- a/leo/plugins/qt_gui.py
+++ b/leo/plugins/qt_gui.py
@@ -618,50 +618,50 @@ class LeoQtGui(leoGui.LeoGui):
         if g.unitTesting:
             return None
 
+        ButtonRole = QtWidgets.QMessageBox.ButtonRole
+
         # Create the dialog.
         top_frame: Optional[QWidget] = c.frame.top if c else None
         dialog = QtWidgets.QMessageBox(top_frame)
-        # Creation order determines returned value.
+        yes = no = yesall = noall = None
         yes = dialog.addButton('Yes', ButtonRole.YesRole)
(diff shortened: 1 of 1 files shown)


gatesphere (approved):
I'm not sure how the previous code ever worked -- exec returns a QDialogCode (0 or 1), and has nothing to do with the order of the button creations.  If this behavior changed, it was prior to Qt5.15, the oldest version I can find docs for.

tbpassin:
The return values were small integers - 2,3,4,5, not in that order, and I couldn't find anything on line about how to know what they would be.  Even ChatGPT let me down.  Anyway, the way in the PR is better because you can directly learn which button was clicked.  One Leo's other similar dialogs already uses a similar technique.

edreamleo:
@tbpassin Thanks for this work.

I didn't realize you had created the PR. I wasn't informed somehow.

My PR #4020 uses `dialog.clickedButton()` as does this PR, but I like my PR a bit better.

What do you think?

tbpassin:
I marked it as a critical bug, with the next available milestone.  I don't know what more I could have done.

edreamleo:
@tbpassin Why are you still pushing to this branch? Let's let PR #4020 do the work.

tbpassin:
I didn't mean to push it.  I update it to dev before deleting it and apparently Github presented that as pushing it.  Sorry.  Should be closed.

edreamleo:
@tbpassin No problem.

PR #4020, opened 2024-07-22
PR for #4012: dialog does not return proper values.

edreamleo:
See #4012.

At last I found a strategy that does not depend on enums:

- [x] Set button names to each button's desired return value.
- [x] Use `dialog.clickedButton` to get the clicked button.
- [x] Return `dialog.clickedButton.objectName() or 'cancel'`.
- [x] Fix `qt_gui.runAskYesNoDialog`.
- [x] Simplify `qt_gui.runAskYesNoCancelDialog`.
- [x] Improve a misleading comment in `qt_gui.runSaveFileDialog`.

### DIFF
--- a/leo/plugins/cursesGui.py
+++ b/leo/plugins/cursesGui.py
@@ -103,7 +103,7 @@ class textGui(leoGui.LeoGui):
     def isTextWidget(self, w):
         """Return True if w is a Text widget suitable for text-oriented commands."""
         return w and isinstance(w, leoFrame.StringTextWrapper)
-    #@+node:ekr.20150107090324.69: *3* runAskYesNoDialog
+    #@+node:ekr.20150107090324.69: *3* runAskYesNoDialog (cursesGui.py)
     def runAskYesNoDialog(self, c, title, message=None, yes_all=False, no_all=False):
         return 'yes'
     #@+node:ekr.20150107090324.15: *3* runMainLoop
--- a/leo/plugins/qt_gui.py
+++ b/leo/plugins/qt_gui.py
@@ -595,14 +595,13 @@ class LeoQtGui(leoGui.LeoGui):
             c.in_qt_dialog = True
             dialog.raise_()  # #2246.
             dialog.exec()
-            dialog.clickedButton()  # #3972.
         finally:
             c.in_qt_dialog = False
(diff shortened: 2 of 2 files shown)


edreamleo:
@tbpassin Whew. Finding a straightforward strategy took a lot of experimenting.

tbpassin:
Didn't you look at my PR? It uses button_clicked and returns the documented return values, unlike this PR.

edreamleo:

As for the ints, the Qt API is a right mess. There are at least two enums involved, but they don't cover all the cases! Imo, `dialog.buttonClicked` is the only way around the mess.

I won't merge this PR until I check all the other Qt dialogs.

tbpassin:
I checked the other one like this - `runAskYesNoCancelDialog` - and it uses dialog.clickedButton(). The return value is computed in an awkward and less clear way compared with this PR.  The dialog as is does return the right values, but would be simpler and cleaner reworked with the approach used in this PR.

edreamleo:
@tbpassin I'll simplify `runAskYesNoCancelDialog` as part of this PR.

edreamleo:
@tbpassin The changes to `qt_gui.runAskYesNoCancelDialog` are ready for your review.

tbpassin (changes requested):
_runAskYesNoCancelDialog()_
In Line 598 we do a `buttonClicked()` inside the `try` block but don't pick up th3 value.  Later in line 604 we do it __again__: ` button = dialog.clickedButton()`.  If the `try` block failed, then the second invocation will probably fail too.

I suggest adding `button = None`  before the `try` block.  Then do the ` button = dialog.clickedButton()` inside the `try` block instead of at line 604. Then line 604 becomes

`        return  button.objectName if button else 'yes'`

[Or maybe we don't need the `try` block at all]

edreamleo:
@tbpassin Rev 91f6ebaf makes the changes you suggest.

The `try/finally` block insures that the code in the `finally` *always* gets executed.

edreamleo:
@tbpassin Ready for another approval :-)

tbpassin (approved):

edreamleo:
@tbpassin Thanks for your second approval.

Issue #4015, opened 2024-07-21
'Extract' should not be disabled in the menu when there's no selection in the body pane.


boltex:
See PR #4022.

The '**Extract**' command is totally legal if no selection length in the body pane and should not be disabled.

For instance,  hitting Ctrl+Shift+D when the cursor is on a line which has text but without a selection (no selection length) still works, and is useful for creating a child node with the headline being the current line - where the cursor is. I use it all the time 😁 

(Pictured below: That the menu shows it disabled if no selection length... On the other hand, **Extract-Names** is ok to be disabled because an actual selection is required in that case)

![image](https://github.com/user-attachments/assets/bea7f1e0-d236-442c-9974-6377a862e7a8)

edreamleo:
Closed via PR #4022.

PR #4022, opened 2024-07-23
PR for #4015: enable extract menus

edreamleo:
See #4015.

- [x] Always enable 'extract' and 'extract-section' menus.


edreamleo:
@boltex This PR is ready for your review.

Issue #4037, opened 2024-08-07
reload-outline command


edreamleo:
See PR #4038.

A new `reload-outline` command will close then reopen *one* outline *without* restarting Leo.

Changing `@string qt-layout-name` will execute  `reload-outline` rather than `restart-leo`.

edreamleo:
@tbpassin This issue will be a preliminary to our work on layouts.

tbpassin:
I'm not giving up on switching to any layout without a reload, not yet.

edreamleo:
@tbpassin Fair enough. The `reload-outline` command should be straightforward and I plan to add it whatever else happens.

tbpassin:
D'accord. It should be useful aside from any layout usage.

edreamleo:
Closed via PR #4038.

PR #4038, opened 2024-08-08
PR for #4037: Add reload-outline command

edreamleo:
See #4037.

- [x] Add the `reload-outline` command to `leo/plugins/qt_frame.py`.
- [x] Add a `reload-outline` menu entry to `leoSettings.py`.
- [x] Make disambiguating headlines more accurate in several files.


tbpassin:
Don't we just about have this already with `revertCommander()`?

edreamleo:
On Thu, Aug 8, 2024 at 11:42 AM Thomas Passin ***@***.***>
wrote:

> Don't we just about have this already with revertCommander()?
>
I'll look into it. But we don't necessarily want to revert any changes.

Edward

tbpassin (changes requested):
1. Code around line 208ff uses wrong calls.  Use `insertTab()` instead.  Don't remove dw before changing index - inserting into the right index will be enough.  Also, from the Qt docs:

_Note: If you call insertTab() after [show](https://doc.qt.io/qt-6/qwidget.html#show)(), the layout system will try to adjust to the changes in its widgets hierarchy and may cause flicker. To prevent this, you can set the [QWidget::updatesEnabled](https://doc.qt.io/qt-6/qwidget.html#updatesEnabled-prop) property to false prior to changes; remember to set the property to true when the changes are done, making the widget receive paint events again._

Something like this (untested):
```
        if is_qt:
            dw = c.frame.top
            tabbed_widget = dw.parent()
            new_index = tabbed_widget.indexOf(dw)
            g.trace(new_index, old_index)
            if new_index != old_index:
                tabbed_widget.setUpdatesEnabled(False)
                tabbed_widget.insertTab(old_index, dw, <tabname>)
                tabbed_widget.setUpdatesEnabled(True)
                dw.show()
```
2. Need an alternative for when `is_qt ` is `False`.

edreamleo:
@tbpassin Thanks for the hints. I'll give them a try later today.

edreamleo:
@tbpassin `LM.revertCommander` is not what we want for this issue.

edreamleo:
@tbpassin The code you suggest does not work:

- `QStackedWidget.insertTab does not exist.
- `tabbed_widget.insertWidget(old_index, dw)` does not seem to work.

edreamleo:
@tbpassin I have removed the code that saves & restores the widget index. The suggestions you made do not work.

I'm happy enough with the latest code.

edreamleo:
@tbpassin Heh. I forgot that there is another possible way forward. A hack could tell the `LeoQtGui` class where to put the outline.

This strategy might avoid problems (including possible Qt bugs) involved with attempts to *move* tabs.

I'll try this way next.

tbpassin:
> @tbpassin The code you suggest does not work:
> 
> * `QStackedWidget.insertTab does not exist.
> * `tabbed_widget.insertWidget(old_index, dw)` does not seem to work.

It works if you use a QTabWidget.  My example was for a QTabWidget. The actual widget is supposed to be a QTabWidget, right? See attached:

[tabwidget_insert_demo.zip](https://github.com/user-attachments/files/16553764/tabwidget_insert_demo.zip)

edreamleo:
@tbpassin It seems that I did not understand Leo's widget structure :-)

The docstring for the `DynamicWindow` class contains this:
```
c.frame.top is a DynamicWindow.
c.frame.top.leo_master is a LeoTabbedTopLevel, a subclass of QTabWidget.
c.frame.top.parent() is a QStackedWidget()
```
This is *good* news, but I don't yet know how to take advantage of it.

I'll be writing an Engineering Notebook post about all this.

edreamleo:
@tbpassin This PR is ready for review. In most cases `reload-outline` is nearly instantaneous.

tbpassin (changes requested):
**qt_frame.py**
----------------
** - Line 2168** - `LeoQtFrame.destroyAllObjects`
If this is not used then let's remove it.

**- Line 2482**:
```
        if not c.fileName():
            c.save()
```
This should read
```
        if c.fileName():
            c.save()
```
The following line should be replaced with `else:`:

`if not c.fileName()` --> `else:`

**- Line 2508 ff** - `stacked_layout = stacked_widget.indexOf(QStackedLayout)` should replace the loop.

**- Lines 2541 ff ** - 
```
 # Fix all tab names.
        for i, name in enumerate(tab_names):
            tab_widget.setTabText(i, name)
```

Doesn't the insertion into the target index automatically accomplish this?

**freewin.py**
---------------
**Line 759** - Changes to freewin do not belong in this PR.  Please revert.

edreamleo:
@tbpassin Some comments:

- I've moved `destroyAllObjects` to the attic, and cleared old items from the attic.
- The strange-looking tests of `c.fileName()` are correct.
- Replacing the tab names by hand was the only approach that worked for me.
- I often add disambiguating headlines where needed.

edreamleo:
@tbpassin Oops.  PR #4045 removes `destroyAllObjects`.

PR #4044, opened 2024-08-18
PR: Add FAQ re install PyQt6-WebEngine on Linux

edreamleo:
See [this discussion](https://groups.google.com/g/leo-editor/c/PQWlN7hvmx0).


tbpassin (commented):
1. Line 32998 - specify --user:

`python3 -m pip install --user --break-system-packages -r requirements.txt`

If run without --user, and the user has admin privileges, Leo and its dependencies will be installed into Python's system site-packages instead of the user's.

2. Add a place-holder for whatever we decide to say about Linux problems that aren't strictly Python-related, like missing system libraries.

edreamleo:
@tbpassin The FAQ now suggests `--user`. If you have any more suggestions, let's add them now. I'm not interested in place-holders.

tbpassin (commented):
Lines 32994 ff:

```
`pip install leo` does not install PyQt6-WebEngine on some Linux distributions.

The solution: use pip's `--user` and `--break-system-packages` options:
```
The same will apply to any packages installed by pip, even `pip install leo`.  `--user` is not needed unless the user has admin privileges because pip in Linux goes to `--user` if installing to the primary locations fails.  But if you include `--user` you don't get an error message about it, and if you have admin authority (e.g., you ran`sudo python -m pip`) your packages won't go into `--user` where you want them.

tbpassin:
The real problems with the QWebEngineView have not been with pip-installation but with missing system libraries (those _*.so.n_ files). These are not exactly a Python/PyQt issue but have to be solved if a distro doesn't have them installed.

Sometimes the error message suggests a command line to install the missing library files and sometimes that command will let the QWebEngine run. But not always. At the least we need to tell people that if they get an error message about one of these libraries they should run the suggested command line. It's not so obvious what one should do otherwise.

tbpassin:
I've also encountered a problem on Linux with matplotlib.  I think the currently VR3 is the only thing that uses it.  The error message includes this line:

``` 
 ImportError: cannot import name 'ImageTk' from 'PIL' (/usr/lib/python3/dist-packages/PIL/__init__.py)
```

I'm not sure if VR3 uses PIL (I got this error with another program that uses matplotlib).  The fix is to use the distro's package manager to install the packages _python3-pil_ and  _python3-pil.imagetk_.  For a distro that uses apt/apt-get:

` sudo apt-get install python3-pil python3-pil.imagetk`

edreamleo:
@tbpassin Thanks for your comments. I'll add them to the FAQ entry.

Issue #4054, opened 2024-08-29
dw.change_layout() Should Add And Remove The `@string qt-layout-name` Node Itself.


tbpassin:
Currently the `dw.change_layout()` method requires the user to add a node with the headline `@string qt-layout-name` to the outline. The method adds the target layout name to the headline, and that node is supposed to be used to determine the layout when the outline gets reloaded.

This is complicated for the user to remember and do, and I for one don't want this extraneous node cluttering up my outline.

Instead, the outline's commander should know the current layout.  The `change_layout()` method should check the commander to learn whether a reload is needed.  If the layout mechanism has to be via such a node, the method should create and insert the node at the end of the outline, and the reload procedure should remove that node during reload.

tbpassin:
I realized this morning that the obvious place for the target layout setting  is in Leo's database.  The sizes and locations of the Leo window are already in it, along with a record of which nodes have been marked. The target layout would be retrieved during reload. The user wouldn't have to go hunting for a specially named setting node that may or may not exist, whose name the user may or may not remember.

edreamleo:
On Thu, Aug 29, 2024 at 11:19 AM Thomas Passin ***@***.***>
wrote:

> I realized this morning that the obvious place for the target layout
> setting is in Leo's database.
>
Thanks for this suggestion. It's worth considering. However, the setting
should exist: it indicates the preferred starting setting.

Edward

edreamleo:
@tbpassin I'm going to close this issue. The existing setting will suffice.

PR #4055, opened 2024-08-29
PR for #4054: dw.change_layout adds nodes as needed

edreamleo:
See #4054.


tbpassin (approved):
Starting at line 96, there are so many different uses of "h" that it's not so easy to keep straight what is intended.  I would like to see
```
SETTING = '@string qt-layout-name'
# ...
 p.h = f"{SETTING} = {layout_name}"
```
The idiom is used at least twice in this section of the code.

edreamleo:
Thanks for approving this PR.

Issue #4056, opened 2024-08-29
reload-outline leaves the old outline visible


edreamleo:
PR #4058 retires `reload-outline`, rendering this issue moot.

Alas, it looks like the `reload-outline` command is thoroughly broken.

I'll probably retire the code and use `restart-leo` instead.

LewisNeal:
Related issue https://github.com/leo-editor/leo-editor/issues/4049

edreamleo:
@LewisNeal @tbpassin I'm going to retire `reload-outline`, replacing it as needed with `restart-leo` and moving the old code to the attic.

I have zero desire to fix this issue, but I'll be happy to consider a PR that brings `reload-outline` back. I'll say more in an ENB.

edreamleo:
Closed via PR #4058. I would welcome a PR that successfully reinstates `reload-outline`.

PR #4058, opened 2024-08-30
PR for #4056: Retire reload-outline.

edreamleo:
See #4056.

- [x] Retire `reload-outline`, moving the code to the attic.
- [x] Use `restart-leo` instead of `reload-outline`.

### DIFF
--- a/leo/doc/leoAttic.txt
+++ b/leo/doc/leoAttic.txt
@@ -1338,6 +1338,82 @@ def updateInjectedIvars(self, w: Any, p: Position) -> None:
     else:
         w.leo_chapter = None
     w.leo_p = p.copy()
+#@+node:ekr.20240830090655.1: *3* retire reload-outline
+#@+node:ekr.20240810080628.1: *4* LeoQtFrame.reloadOutline
+@frame_cmd('reload-outline')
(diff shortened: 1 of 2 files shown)


tbpassin:
I don't understand this at all. I don't want to have to restart Leo each time I want to make a change to the layout of a single outline.  I usually have 5 or 6 outlines open at once. Each time I want to change one of them I have to wait for Leo to restart and reload all of them? It's just not reasonable nor acceptable.  And if one of the outlines is LeoPyRef, it takes a long time to get that one open again.

I won't approve this unless I see an awfully strong case.  And on the face of it, a user could close an outline and re-open it himself.  Why Leo can't do that for itself successfully is hard to understand.

How about if the rush to make changes is resisted?  Let's slow down and wait until later.

edreamleo:
@tbpassin This PR establishes a baseline of code that works. There is no other rush.

I understand you have a use case that would benefit from a `reload-outline` command. I will happy to approve a PR that fixes that command. Closing an outline and then re-opening the outline is (apparently) harder than it looks.

tbpassin:
Actually, it turns out to be almost trivial.  See my latest post on the Google Groups site (the second one in the thread):

https://groups.google.com/g/leo-editor/c/R4TeA5qqr8Y

Issue #4032, opened 2024-08-06
rst3 command improperly dirties the @rst3 node


boltex:
See PR #4059.

Consider `@rst myContent.html`:
```
====
Test
====

This is a test.
```

The rst3 command correctly generates the html, but the node becomes dirty without changing the node in any way.

edreamleo:
@boltex No calls to `c.setChanged` or `p/v.setDirty` appear in `leoRst.py`, so I think this report is mistaken.

boltex:
@edreamleo 

Maybe I made a mistake in testing this out and there's no bug... 

I'll try to see if I can actually reproduce this bug. I'll re-open if I can clearly reproduce it.

_I should have taken better notes and screenshots in the original Leo!_

boltex:
I just redid the setup. and the bug does indeed exist. Hard to tell on the screenshot, but the first node is dirty. 

![image](https://github.com/user-attachments/assets/79221d67-e843-412c-b302-4395b8c1e94d)

## How to reproduce

Create an @rst node , fill up its child like I did, save to make sure all nodes are **not dirty**, then, with either the @rst node or its child selected, hit alt+x, type rst3 and enter.

The files are generated by the rst command, and the @rst node is dirty.

edreamleo:
On Wed, Aug 28, 2024 at 7:17 PM Felix ***@***.***> wrote:

> I just redid the setup. and the bug does indeed exist. Hard to tell on the
> screenshot, but the first node is dirty.
>

Thanks for the detailed report. I'll fix this immediately.

Edward

edreamleo:
Closed via PR #4059.

PR #4059, opened 2024-09-01
PR for #4032: Don't mark @rst3 nodes dirty

edreamleo:
See #4032.

- [x] Fix `processTree` so that it never changes `p.h`.


edreamleo:

Issue #4063, opened 2024-09-21
Suppress new pylint complaints


edreamleo:
See PR #4064.

A blunder in `~/.leo/.pylintrc` prevented some crucial checks. As a result, #3950 is invalid and has been closed.

edreamleo:
Closed via PR #4064.

PR #4064, opened 2024-09-21
PR for #4063: pylint complaints

edreamleo:
See #4063.  This PR fixes all bugs reported pylint 3.2.7.

**pylint bug found and reported**

- pylint 3.3.0 did not work with Leo's original sources. See [pylint issue #9945](https://github.com/pylint-dev/pylint/issues/9945).
- The culprit was unintentionally empty format specifiers, as in `f"{x:}"`.

**Workaround pylint bug #9945**

- As an initial workaround, I checked all sources with pylint 3.2.7.
- [x] Remove empty format specifiers, allowing Leo to use pylint 3.3.0.

**Update the pylint configuration files**

- [x] Update my local copy of  `~/.leo/.pylintrc`.
    Enable all checkers except the "design" and "similarities" checkers.
- [x] Copy  `~/.leo/.pylintrc`. `leo/test/pylint-leo-rc-ref.txt`.

**Fix all resulting complaints**

- [x] Suppress pylint complaints about unused imports in leoQt.py.
- [x] Use 120 as the maximum line length.
    **Aha!** The `line-too-long` suppression must appear *before long docstrings*.
- [x] Add suppressions to the following plugins:
    demo.py, editpane/csvedit.py, leo_cloud.py, mod_http.py, richtext.py, stickynotes_plus.py.
   *These suppressions may mask real bugs*.
- [x] Use the VR3 plugin code from the tbp-vr3-docstring-rewrap branch as a starting point.
    Add a suppression line, fix four valid complaints, and fix a typo.

**Retire buggy minibuffer-based spell commands**

- [x] Move all (minibuffer-based) `spell-*` commands to the attic. These commands have probably *never* worked.
- [x] Remove all mention of the `spell-*` commands in LeoDocs.leo.

**Fix several real bugs**

- [x] `qt_tree.py`: Add a new `pattern` arg to `declutter_replace`.
- [x] `plugins/leo_babel/tests/idle_time.py`: `os.sched_yield()` exists only on some flavors of *nix.


edreamleo:
@tbpassin The changes to mostly involve VR3 reformat docstrings. I think it's important to avoid super-long lines.

edreamleo:
@tbpassin Are you good with the changes to VR3?

tbpassin (commented):
General
----------
There are a number of changes to type-checking expressions.  I'm a little concerned that some  will only work for Python 3.12+.  (I haven't tried to check this).

LeoGlobals.py
----------------
- _Lines 1971 ff_ This code uses the idiom 
`from leo.core import leoGlobals as leo_g  # pylint: disable=import-self`

Aside from not me understanding why setting `leo_g = g` doesn't do the same thing, it seems to me that these tests should be in a different module.  Then the self-import would not be needed.

qt_tree.py
----------
- _Line 237_ `f'Error in declutter: {e!r}"\n'`. I think this is missing a matching quotation character.  It isn't functional but the displayed message will look strange.  I think it should be:

`f'Error in declutter: "{e!r}"\n'`

rpcalc.py
----------
_Line 2000_:
`def addIconPath(self, potentialPaths=[]):  # pylint: disable=dangerous-default-value`

Let's just change the default `[]` to `None` instead of disabling the warning. The method hasn't been implemented anyway (almost all of the code is by the original author of the RPCalc stand-alone application.  I just adapted it to be able to live within Leo.  I'm not very familiar with most of the code base).

viewrendered3.py
---------------------
The docstring formatting is too complicated to be fixed by `@pagewidth 120`.  I have completely reformatted all the docstring sections in a branch in my own repo.  They are based on a linewidth of 65, which I think is better for a docstring. I also fixed some RsT formatting quirks for better rendering by VR3.  Just go ahead and take them from it:

https://github.com/tbpassin/leo-editor/tree/tbp-vr3-docstring-rewrap

tbpassin:
> @tbpassin Are you good with the changes to VR3?

No.  See my review comments above for a satisfactory reformatting.

edreamleo:
@tbpassin Many thanks for your excellent comments.

- General: I ran all tests using python 3.9. There were a few errors, but none concern me.
- leoGlobals: I don't remember the self imports are necessary, but removing would be a major (unnecessary) change.
- qt_tree: Good catch.
- rp_calc: Good idea.
- VR3: I'll do as you suggest. I'll start over using your latest version and fix a few minor pylint complaints.
  The diff will be two or thee non-controversial lines.

edreamleo:
@tbpassin Rev 2c46764 starts afresh from the tbp-vr3-docstring-rewrap version, with two changes:

- Add various pylint suppressions at lines 5 & 6.
- Fix a typo at line 1357.

Four (valid, easily fixed) pylint complaints remain. I'll fix those in the next push.

edreamleo:
@tbpassin Rev 1bfc324 fixes the remaining pylint complaints. I plan no further changes to VR3.

Please review and approve this PR.

tbpassin (approved):

edreamleo:
@tbpassin Many thanks for your help with this PR.

Issue #4067, opened 2024-09-24
Optionally run pylint on save


edreamleo:
A proof of principle for the pylint folks.

edreamleo:
Closed via PR #4068.

PR #4068, opened 2024-09-24
PR for #4067: run-pylint-on-write

edreamleo:
See #4067.


tbpassin (approved):
leoAtFile.py
---------------
- Line 1080ff:
```
if n > 1:
                g.es(f"finished: {n} unchanged file{g.plural(n)}")
```

We don't need g.plural(n) because n is guaranteed to be plural.  We only need:

```
 if n > 1:
                g.es(f"finished: {n} unchanged files")
```

- Line 2169: `       ok1 = ok2 = ok3 = True`

Let's have more informative names.  E.g., ok3 -> ok_lint, ok1 -> ok_flake8, ok2 -> ok_pyflakes.

edreamleo:
@tbpassin Your comment about `g.plural` is correct. I left it in just in case the code changes later.

Issue #4078, opened 2024-10-08
Fix all errors and warnings arising from PR #4077


edreamleo:
PR #4077 contains several errors.  See PR #4078.

edreamleo:
Fixed via PR #4079.

PR #4077, opened 2024-10-01
New Leo Layout System

tbpassin:
Implements a new layout system:
- [x] Adds a new _LayoutCacheWidget_ class to cache unused splitters and widgets for reuse;
- [x] Dynamic Window  hosts an instance of _LayoutCacheWidget_;
- [x]  _LayoutCacheWidget_ contains methods to create layouts from layout descriptors;
- [x] New file _plugins/qt_layout.py_ contains _LayoutCacheWidget_ and all the current built-in layouts;
- [x] All layouts can be applied without restarting the outline;
- [x] The default layout specified in the applicable @settings tree can be re-applied without restarting the outline.
- [x] The _DynamicWindow_ startup sequence that creates the initial layout is unchanged;
- [x] `g.app.gui.find_widgets_by_name()` is fixed so it no longer fails to find some widgets;
- [x] Some VR3 commands are adjusted to work better with this new system.
- [ ] For layouts that can take either VR or VR3, generalize the layout structure to allow user to choose which of these alternatives to use in a layout.
- [ ] Decide if VR/VR3 should be visible when a layout that contains them is applied.

qt_layout.py contains minibuffer commands to create all the layouts.  They all start with "layout-".  For the time being, layouts that expect to have VR or VR3 in one of the frames create the frame but do not actually open VR/VR3. They can be displayed in their frame by "vrx-toggle" or "vrx-show" commands.

A companion outline containing a custom menu with many of the new commands, and other layout commands, will be supplied separately.

This PR satisfies Issue https://github.com/leo-editor/leo-editor/issues/4076.

PR #4079, opened 2024-10-08
PR for #4078: fix errors in PR #4077

edreamleo:
See #4078 and PR #4077.

**qt_layout.py**

*Note*: I plan many stylistic improvements to `qt_layout.py`, but these will happen in a separate PR.

- [x] Beautify.
- [x] Fix the following pylint complaints:
  - W0311: Bad indentation. Found 17 spaces, expected 16 (bad-indentation)
  - W0105: String statement has no effect (pointless-string-statement)
  - Dangerous default value FALLBACK_LAYOUT (builtins.dict) as argument (dangerous-default-value)
- [x] Fix or suppress all mypy complaints.
- [x] Fix a failing unit test: add a top-level `init` function.
- [x] Fix a failing unit test: Exempt `qt_layout.py` from `test_all_qt_plugins_call_g_assertUi_qt_`.

**VR3**

- [x] Fix an unterminated docstring in `g.command('vr3-tab')`!!!
- [x] Beautify.
- [x] Fix a logic error in `update_rendering` found by mypy.
- [x] Fix or suppress all mypy complaints.
   - Add dubious annotation: ` self.rst_html: Any = ''  # bytes or str.`
     **A bug may lurk here**.

### DIFF
--- a/leo/plugins/qt_layout.py
+++ b/leo/plugins/qt_layout.py
@@ -1,10 +1,11 @@
 #@+leo-ver=5-thin
 #@+node:tom.20240923194438.1: * @file ../plugins/qt_layout.py
 """The basic machinery to support applying layouts of the main Leo panels."""
+
+#@+<< qt_layout: imports >>
+#@+node:tom.20240923194438.2: ** << qt_layout: imports >>
 from __future__ import annotations
 
-#@+others
-#@+node:tom.20240923194438.2: ** imports
 from collections import OrderedDict
 from typing import Any, Dict, TYPE_CHECKING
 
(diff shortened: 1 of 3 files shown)


edreamleo:
@tbpassin Please approve this PR *immediately*. The "devel" branch is unstable at present.

I plan no further changes to VR3. I'll make stylistic changes to `qt_layout.py` in a separate PR.

tbpassin (approved):
Quick run-through, seems OK.

edreamleo:
@tbpassin Thanks!

tbpassin:
An init() function was added to the qt_layout.py module.  **This plugin is not optional.** It is required by coordinated changes to DynamicWindow. DW imports the plugin independent of that PluginsController.

I don't know what would happen if someone enabled the module as a plugin, maybe nothing, but it's not designed that way nor intended to be used that way.

LewisNeal:
I found I had to enable qt_layout.py module in @enabled plugins. Otherwise under _Plugins>qt_layout_ the menu was not visible.

tbpassin:
The *layout-* commands are available anyway.  A menu can be built using those commands.

edreamleo:
@tbpassin The top-level `init` function always returns True, so it will have no effect except to cause a unit test to pass.

PR #4080, opened 2024-10-08
PR: Reorganize qt_layout per EKR's long-standing style.

edreamleo:
This PR reorganizes qt_layout per EKR's long-standing style:

- [x] Create separate nodes for all layouts and commands.
- [x] Group layouts together and commands together.
- [x] Add a diagram for Leo's legacy layout.


edreamleo:
@tbpassin This PR is ready for review. I strongly prefer the organization this PR creates.

PR #4085, opened 2024-10-09
New Layouts: Use VR3 If Enabled, Else VR

tbpassin:
Most but not all of the new layouts include a rendering pane.  For layouts defined by a layout structure and applied by calling `restoreFromLayout()`, the structure up until now has required either VR or VR3 be specified.

This PR optionally allows a layout writer to tell the system to use VR3 if it has been enabled else use VR.  In addition, the "big-tree" layout command, which uses code rather than `restoreFromLayout()` to create its layout, has been updated to also use VR3 if enabled.

- [x] Generalize layout structure and `restoreFromLayout()` to identify that VR3 should be used if available;
- [x] Add code to "layout-big-tree" command to accomplish the same thing.
- [x] Update the built-in layout structures to use VR3 if available.
- [x] Tested all combinations with/without VR3 enabling.
- [x] Deleted an obsolete and unused method.

### DIFF
--- a/leo/plugins/qt_layout.py
+++ b/leo/plugins/qt_layout.py
@@ -22,6 +22,13 @@ if TYPE_CHECKING:  # pragma: no cover
 #@-<< qt_layout: imports >>
 
 #@+others
+#@+node:tom.20241009141008.1: ** Declarations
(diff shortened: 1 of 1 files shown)


edreamleo (approved):
Looks good. Thanks for this work!

Issue #4083, opened 2024-10-09
Add layout descriptions for layout commands that lack them


edreamleo:

tbpassin:
Note that I will have a different PR for layouts coming up shortly.

edreamleo:
@tbpassin Thanks for the heads up. I'll wait for your PR.

tbpassin:
I just created it.  I forgot to reference the corresponding issue, so please add that link if you would.

edreamleo:
@tbpassin Done.

edreamleo:
Closed via PR #4086.

PR #4086, opened 2024-10-10
PR for #4083: Add layout diagrams to two docstrings

edreamleo:
See #4083.

- [x] Add layout diagrams to two docstrings.
- [x] (Debatable): Remove implementation details from the `layout-swap-log-panel` command.


tbpassin (changes requested):
Lines 54ff
========
` This command puts the outline in a separate area. The effect of this
    command depends on the existing layout. For example, if the legacy
    layout is in effect, this command changes the layout`

No, this layout is not conditional.  It always changes to a known starting point and then produces the layout 
```
        ┌──────────────────┐
        │  outline         │
        ├──────────┬───────┤
        │  body    │  log  │
        ├──────────┴───────┤
        │  VR              │
        └──────────────────┘
```
You may have been fooled if VR/VR3 had not already been shown.

Looking into this I found a bug on my part: in `show_vr_pane`, the `setUpdatesEnabled()` call only works for VR3, not VR.  The easiest way to make this work transparently is:
```
def show_vr_pane(c, w):
    try:
        w.setUpdatesEnabled(True)
    except:
        pass
    c.doCommandByName('vr-show')
```
Please make this change.

edreamleo:
@tbpassin rev 1f41048 incorporates all your changes and fixes a quirp in another docstring.

tbpassin (approved):

edreamleo:
@tbpassin Thanks Thomas! These layout PRs are a milestone for Leo.

tbpassin:
Thanks!

Darn it, I think that change to `show_pane()` still isn't right because it doesn't fully adjust for VR vs VR3.  Maybe I should have converted that big-tree layout to use a layout structure, but I wanted to leave it as is to someone could use it as a kind of template to build other layouts by programming instead of a layout structure.

I'll send up a fix later today.

edreamleo:
@tbpassin You're welcome.

Issue #4093, opened 2024-10-12
Lower-case @key F-key bindings don't work in @button or @comman nodes


edreamleo:
See PR #4094.

For both `@button` and `@command`, bindings such as `@key = f6` should work.  They don't, but `@key = F6` does work.

This is an ancient bug. The fix is straightforward.

PR #4094, opened 2024-10-12
PR: allow lower-case f-keys in @key bindings

edreamleo:
See #4093.

- [x]  Allow both `'f'` and `'F'` in `k.isFKey`.
- [x] Translate  `'f'` to `'F'` in `sc.getShortcut`.

My hand tests pass for both `@command` and `@button` nodes.


edreamleo:
@tbpassin Thanks for your comment and approval.

Rev f08ba4d clarifies the comment a bit.

Issue #4097, opened 2024-10-13
"execute-external-file" Command Bug Prevents Honoring Of Terminal Specification


tbpassin:
The _execute-external-file_ command is supposed to read the path to a terminal emulator program form an optional `@data` node if it exists.  However, a bug in the code that processes the `@data` node prevents that from succeeding.

This should be fixed before the next release of Leo.

edreamleo:
@tbpassin Thanks for this report. I agree that this issue should be fixed for 6.8.2.

tbpassin:
In the PR to come, I will also include code to address https://github.com/leo-editor/leo-editor/issues/4091, since it is all about the same command.

I expect to be ready for a PR sometime tomorrow.

edreamleo:
On Mon, Oct 14, 2024 at 9:47 AM Thomas Passin ***@***.***>
wrote:

> In the PR to come, I will also include code to address #4091
> <https://github.com/leo-editor/leo-editor/issues/4091>, since it is all
> about the same command.
>
> I expect to be ready for a PR sometime tomorrow.
>
Many thanks for this work. I'll look forward to the PR.

Edward

edreamleo:
Closed via PR #4100.

Issue #4091, opened 2024-10-11
"execute-external-file" command fails on some Linux systems


tbpassin:
The `execute-external-file` command launches a node or subtree in an external terminal and leaves the terminal open after the program has finished running so any printed output can be seen. Making the command work on as many Linux systems as possible was a challenge because different distros use different terminal emulators and some of them use different syntax to launch programs.

I have found a few distros for which the command failed. One common terminal emulator, gnome-terminal, may change its syntax in the future in a way that may not work.

The problem is that to discover the right syntax, the command parses the terminal's help message.  But differences in the help output can make it hard to do this correctly although it works for most cases. Even `xterm` is not universally available, and even when it is, often its appearance and readability is poor.

The determination of the correct terminal syntax to use can be improved by hard-coding the right syntax for the most common cases known. The ones I know about so far are
`('konsole', 'xfce4-terminal', 'mate-terminal', 'gnome-terminal', 'xterm')`. If anyone knows others that should be included along with the right syntax, please add the information in a comment. An example of the syntax comes from the xterm help string:
```
$ xterm -help |grep execute
    -e command args ...          command to execute
```
Here the option to use is `-e`.

 I will wait a few days to see if there are any suggestions, then go ahead with a pull request.

edreamleo:
@tbpassin Thanks for your detailed report.

edreamleo:
Completed via PR #4100.

PR #4100, opened 2024-10-14
execute-external-file Command: Fix and Enhance

tbpassin:
- [x] Linux: correct bug so that terminal to use is read correctly from `@data` file.
- [x] Linux: combine known terminals and their execution arguments into one data structure.
- [x] Linux: add more terminals and their execution arguments to known terminals.
- [x] Test on OpenSUSE/KDE and EndeavourOS/Cinnamon.
- [x] Confirm that command still works on Windows.


tbpassin:
This PR resolves Issues https://github.com/leo-editor/leo-editor/issues/4097 and https://github.com/leo-editor/leo-editor/issues/4091.

edreamleo (approved):
Excellent!

PR #4101, opened 2024-10-15
PR Fix mypy complaint by converting a generator to a list

edreamleo:
See PR #4100. mypy complains:
```console
leo\core\leoCommands.py:994: error: Argument 1 to "getCommonTerminal" has incompatible type "dict_keys[str, str]";
expected "str | list[Any] | tuple[Any, ...]"  [arg-type]
```
- [x] Change the signature of `getCommonTerminal` to:
    `def getCommonTerminal(names: Union[str, Iterable[str]]) -> str:`

I used the older `Union` syntax for compatibility with older Pythons.
This compatibility might not be needed, but it can't hurt.

### DIFF
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -751,7 +751,7 @@ class Commands:
             'xterm': '-e '
         }
 
-        PREFERRED_TERMINALS = EXECUTE_ARGS.keys()
+        PREFERRED_TERMINALS = list(EXECUTE_ARGS.keys())
         #@+node:tom.20241014154415.4: *4* SETTINGS_HELP
         SETTINGS_HELP = r'''The data in the @data node body must have a
         PROCESSORS and an EXTENSIONS section, plus an optional TERMINAL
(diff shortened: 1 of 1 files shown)


edreamleo:
@tbpassin I thought it best for you to check my work.

tbpassin:
Don't make things more complicated than needed. Don't change PREFERRED_TERMINALS. The type signature in `getCommonTerminals` should be

`str | Iterable[str]`

tbpassin (changes requested):
Don't change PREFERRED_TERMINALS. Change the argument signature of `getCommonTerminals()` to

`str | Iterable[str]'

The type `dict_keys` satisfies `Iterable`, as one can see from the expression

`for k in dict.keys()`

edreamleo:
@tbpassin Will do.

edreamleo:
@tbpassin Fixed at rev 278388f.

Thomas, in the future, *please, please, please* remember to run *all* tests, including mypy and pylint, before doing any push.

You will save us both considerable work.

tbpassin:
[grovels...] Sorry, very sorry. Will do.

edreamleo:
@tbpassin Thanks for your apology. I'm going to merge this now.

Issue #4110, opened 2024-10-19
qt_frames shows hundreds of mypy messages


tbpassin:
In testing an upcoming PR, I ran mypy on qt_frames.py. It generated hundreds of messages.  This is true of devel - it is not related to my PR.  Many if not most of the messages are "[attr-defined]".   These are usually spurious because mypy isn't smart enough.  For example,

`g.app.gui.find_widget_by_name()` causes an "[attr-defined]" message despite the method actually existing.  In this case, pymy apparently doesn't know the type of g, or maybe of g.app.gui.  Without knowing its type, mypy can't know if there is such and attribute or method.

Fixing these should not be combined with any other PR - it would be very confusing and time-consuming to work through.  Since these changes have no immediate functional effect, and new PRs that affect qt_frame are coming immediately, I suggest that the addressing the mypy issues be postponed until after 6.8.2.

edreamleo:
I am going to close this issue. These messages arose because of an environment issue.

PR #4103, opened 2024-10-16
New Layouts - Collect Layout Docstrings For Help Messages

tbpassin:
Also add type-checking annotations and improve docstrings for various methods. This PR partially completes Issue https://github.com/leo-editor/leo-editor/issues/4099.

- [x] Clean up and improve layout docstrings.
- [x] Add and improve various method docstrings.
- [x] Confirm that the method of decorating and collecting layout docstrings works.
- [x] Add typing annotations.
- [x] No pylint complaints.
- [x] No mypy complaints.
- [x] Type annotations do not interfere with correct functioning.
- [x] pytest:  `932 passed, 2 skipped, 2 warnings in 27.02s`

PR #4111, opened 2024-10-19
Use new layouts at startup

tbpassin:
This PR changes the Dynamic Window to create the initial layout using the qt-layout methods instead of the 6.8.1 code.  The user can change layouts without reloading the output and now all layouts can be defined in only one place, qt_layout.

- [x] Add new code to DW, remove obsolete layout code.
- [x] Combine "legacy" and "quadrant" layouts because they are actually the same.
- [x] Modify getVr() so it can be called correctly during DW startup.
- [x] `LayoutCacheWidget` in qt_layout makes sure VR3 is loaded during startup so that it can be placed into the right place in the initial layout. This causes an additional screen flash but only once, during Leo startup.
- [x] Minor cleanup of some layout nodes.
- [x] Added a command alias for 'layout-restore-default': 'layout-restore-to-setting'.
- [x] pylint: no errors.
- [x] pytest: `932 passed, 2 skipped, 2 warnings in 26.54s`
- [ ] mypy - mypy generates hundreds of errors for qt_frames unrelated to this PR.  It is impractical and undesirable to fix them all as part of this PR (see https://github.com/leo-editor/leo-editor/issues/4110).


tbpassin:
Please see comment to https://github.com/leo-editor/leo-editor/pull/4103 because it may be hard to merge these two PRs.  They are not incompatible but I don't see how some of the changes can be blended together without some careful hand work.  At least, I don't know how to do that in a merge tool.  This PR should have priority and then we can see how to blend in changes from 4103.

edreamleo (approved):
I don't see anything that would generate "hundreds" of mypy complaints.

I'll handle any complaints that do appear after the merge.

edreamleo:
@tbpassin All tests pass after the merge. Leo's beautifier changed `qt_frame.py`, but that's no big deal.

Many thanks for this PR.  Onward!

tbpassin:
> I don't see anything that would generate "hundreds" of mypy complaints.

Maybe my copy of the mypy config file is different from yours for some reason.

PR #4120, opened 2024-10-22
PR: Fix PR #4111: don't enable VR3 automatically

edreamleo:
See PR #4111.

- [x] Add comment to top-level`init` function: `plugins/qt_layout.py` is not a true plugin.
- [x] Add a guard to the `<< initialize data structures >>` section of `LayoutCacheWidget.restoreFromLayout`.
- [x] Remove all `use-*` commands. The `layout-*` commands are better and suffice.
   There is no need to retain the `use-*` commands for compatibility.
- [x] Ensure that all tests pass.


edreamleo:
@tbpassin PR #4120 has been merged into "devel". The PR couldn't wait.

There should be no significant problem merging PR #4120 into any future work.

Thomas, please avoid indefinite pronouns such as "this" and "this one". The confusion is driving me crazy.

tbpassin:
OK, I agree with your correction at `if has_vr3` in qt_layout.

tbpassin:
Don't we want qt_layout to be loaded even if not enabled, just as VR is? If qt_layout isn't enabled, we seem to get the vertical-thirds layout with VR.  That's fine with me although I'm not quite sure how it happens.

edreamleo:

Point 1: I'm glad you agree that the `if has_vr3` guard is correct.

Point 2: The question about enabling and importing `leo.core.qt_layout` is a good one. The answer is a bit tricky:

- `qt_gui.py` contains `from leo.plugins import qt_frame`, so Leo always *imports* `qt_frame.py`.
- `qt_frame.py` is **not** a real plugin, and users never need to enable it via `@enabled-plugins`.
- The top-level `init` function in `qt_frame.py` always returns `True`, just in case somebody does enable the plugin.
  However, the real reason for the `init` function is to keep one of Leo's unit tests happy.

edreamleo:
@tbpassin I forgot to mention that the `layout-*` commands work even if neither VR nor VR3 is enabled.

Heh, I still see (in "devel") both the `layout-legacy` and `layout-quadrant` commands.

tbpassin:
> The question about enabling and importing `leo.core.qt_layout` is a good one. The answer is a bit tricky

If the plugin is enabled, the plugins menu will have a submenu of the layout commands. If it's not enabled, the submenu for qt_layout won't be there. I think there is some value in that.

tbpassin:
> Heh, I still see (in "devel") both the `layout-legacy` and `layout-quadrant` commands.

That was intentional, in case someone is using that value in their settings. Or if they don't have the setting, they will get the layout that, IIRC, is currently documented.  But as you probably noticed, they are aliased so there is only one underlying layout.

edreamleo:
Hmm. I think the `layout-quadrant` command just adds confusion. I think it should go.

edreamleo:
@tbpassin I agree that `qt_layout.py` should create a sub-menu listing layout commands.

But perhaps that sub-menu belongs in the `Settings` menu. I'll take a look.

tbpassin:
My upcoming PR has a new command  in qt_layout: 'layout-show-layouts'.  It shows all the layout docstrings in a tab in the Log frame. I think that's exactly what should be dispatched by a "Settings/Show Settings" menu item.

edreamleo:
@tbpassin I agree: `layout-show-layouts` command belongs in the `Settings/Show Settings` menu.

PR #4121, opened 2024-10-23
Finalize New Layouts;  Add help-for-layouts and layout-show-layouts Commands

tbpassin:
- [x] Removed "legacy" and "fallback" layouts;
- [x] Added mapping of command names and docstrings;
- [x] Improved logic for choosing layout if layout setting is missing or invalid;
- [x] Created command 'layout-show-layouts'.
- [x] Created menu item for above in menu Settings/Show Settings
- [x] Created "help-for-layouts" command;
- [x] Created menu item in Help/Open help topics menu for the "help-for-layouts" command;
- [x] Editorial clarifications for layout docstrings;
- [x] Added some type annotations;
- [x] (qt_layout) pylint: no messages;
- [x] (qt_layout) mypy: no messages;
- [x] pytest: `932 passed, 2 skipped, 2 warnings in 26.10s`; [Note: qt_layout has no unit tests]

PR #4126, opened 2024-10-25
PR: Urgent changes to Leo's new layout code

edreamleo:
PR #4121 needs work.

Leo *must* use the name `legacy` instead of `quadrant` to avoid breaking user settings.

- [x] Change `quadrant` to `legacy` everywhere.
- [x] Change the default for `@string qt-layout-name` to `legacy`.
   This default preserves Leo's legacy look and feel.
- [x] Retire the duplicate `layout-restore-default` command.
- [x] Simplify the `restoreDefaultLayout` function and correct its docstring.
- [x] Fix long lines in `commands/helpCommands.py`.
- [x] Simplify the `show-layouts` and `help-for-layouts` commands.

### DIFF
--- a/leo/commands/helpCommands.py
+++ b/leo/commands/helpCommands.py
@@ -785,49 +785,44 @@ class HelpCommandsClass(BaseEditCommandsClass):
     #@+node:ekr.20240822071015.1: *3* helpForLayouts
     @cmd('help-for-layouts')
     def helpForLayouts(self, event: LeoKeyEvent = None) -> None:
-        """Print a messages telling you how to get started with Leo."""
-        # A bug in Leo: triple quotes puts indentation before each line except
-        # the first.
+        """Print a message telling you how to use Leo's layouts."""
         c = self.c
-        #@+<< create listing >>
-        #@+node:tom.20241022174807.1: *4* << create listing >>
+        #@+<< create listing list>>
+        #@+node:tom.20241022174807.1: *4* << create listing list >> (help-for-layouts)
         dw = c.frame.top
         cache = dw.layout_cache
         layouts = cache.layout_registry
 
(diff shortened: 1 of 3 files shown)


tbpassin:
> Leo must use the name legacy instead of quadrant to avoid breaking user settings.

##@! I told you at the time I wanted to keep the legacy alias in case someone was using it, but you said no.  So I took it out.

You just ruined the help-for-layouts rendering.  I had the devil's own time getting that to work, what with the munging that the put() function imposes.  My code compensated for that but the format of the docstrings had be how I had it.  Now the layout diagrams are garbage.

If you had just told me that your default layout setting was "legacy" and you wanted to keep that name, I could have fixed the whole thing in 1 minute. There was no emergency.

Item 1 in my PR was:
 - [x] Removed "legacy" and "fallback" layouts;

edreamleo:
On Fri, Oct 25, 2024 at 5:05 PM Thomas Passin ***@***.***>
wrote:

> You just ruined the help-for-layouts rendering.

Oops. I didn't scroll far enough down in my tests. I'll fix it.

Otoh, show-layouts works with much cleaner code. It's all good.

Edward

edreamleo:
On Fri, Oct 25, 2024 at 6:04 PM Edward K. Ream ***@***.***> wrote:

> On Fri, Oct 25, 2024 at 5:05 PM Thomas Passin ***@***.***>
> wrote:
>
> > You just ruined the help-for-layouts rendering.
>
> Oops. I didn't scroll far enough down in my tests. I'll fix it.
>

The easy solution is to add capabilities to c.putHelpFor, via a kwarg or a
separate method.

The problem is that help text is assumed to be rendered as rST, but in this
case plain text would work better. That's what show-layouts does.

Edward

Issue #4123, opened 2024-10-24
log-pane command takes no effect after git pull to ef8105daa3


wangzhaohe:
None of the expand/contract-*-pane commands work with the new layout code.

<details>

Leo Log Window
Leo 6.8.2-devel, devel branch, build ef8105daa3
2024-10-22 11:02:35 -0500
Python 3.12.0, PyQt version 6.5.3
darwin

3 command takes no effect after git pull to ef8105daa3

    hide-log-pane
    expand-log-pane
    contract-log-pane

<details>

edreamleo:
@tbpassin @wangzhaohe Thanks for this report. None of these commands work for me on Windows.

tbpassin:
When I wrote that these commands worked for me, I was using the "vertical-thirds" layout. They may work differently with different layouts.

I think these commands assume that the log pane shares the secondary splitter with some other widget, like the outline frame. Some of the new layouts may not meet this assumption. Or the commands may act differently for different layouts.

For example, the "contract-log-pane" command will do nothing if the secondary splitter contains more or fewer than two child widgets. That's the case for the "render-focused-layout", for example, in which the secondary splitter contains three widgets.

These commands should be revised to allow for the new layout possibilities and how they nest their widgets.

For layouts defined by a layout structure - almost all of them - it's easy to tell what's in the secondary and main splitters. A layout structure includes an ordered dictionary defined by tuples (widget, splitter), like this:

```
    'SPLITTERS': OrderedDict(
            (('outlineFrame', 'secondary_splitter'),
            ('bodyFrame', 'secondary_splitter'),
            ('logFrame', 'secondary_splitter'),
            (VRX_PLACEHOLDER_NAME, 'body-vr-splitter'),
            ('secondary_splitter', 'main_splitter'),
            ('body-vr-splitter', 'main_splitter'))
        ),
```

For this layout, there are three widgets that share the secondary splitter, in the order listed, and two widgets in the main splitter.

edreamleo:
On Thu, Oct 24, 2024 at 10:36 AM Thomas Passin ***@***.***>
wrote:

> I think these commands assume that the log pane shares the secondary
> splitter with some other widget, like the outline frame. Some of the new
> layouts may not meet this assumption. Or the commands may act differently
> for different layouts.
>
...

> These commands should be revised to allow for the new layout possibilities
> and how they nest their widgets.
>
For layouts defined by a layout structure - almost all of them - it's easy
[...]

PR #4136, opened 2024-10-27
PR for #4123: Fix all expand/contract-*-pane commands

edreamleo:
See #4123.

**qt_layout.py**

- [x] Simplify mypy annotations to conform to Leo's conventions.
   class `LayoutCacheWidget` is now a subclass *only* of `QWidget`.
- [x] Rewrite `LCW.find_splitter_by_name` so that it can *only* return a `QSplitter`.
- [x] The `LCW.expand/contract_*_pane` methods call `LCW.expand/contract_pane`
   with *objects* (official ivars) instead of *object names*. There is no need to know object names!
- [x] `LCW.expand/contract_pane` call `LCW.resize_pane`, completing the infrastructure.
- [x] `LCW.resize_pane` does the work.

**Other files**

- [x] `qt_frame.py`: Define all commands by delegating all work to the `LCW` class.
- [x] `qt_gui.py`: Add `LeoQtGui.find_parent_splitter`.
- [x] `viewrendered3.py`: Add disambiguating headlines. Otherwise, this file is completely unchanged.

**Retain old splitter methods**

The `divideAnySplitter` and `divideSplitter1/2` methods methods should remain in the `LeoQtFrame`,
despite their limited relevance.

### DIFF
--- a/leo/plugins/qt_frame.py
+++ b/leo/plugins/qt_frame.py
@@ -4285,50 +4285,48 @@ class TabbedFrameFactory:
             c.redraw()
     #@-others
 #@+node:ekr.20200303082457.1: ** top-level commands (qt_frame.py)
-#@+node:ekr.20200303082511.6: *3* 'contract-body-pane' & 'expand-outline-pane'
+#@+node:ekr.20200303082511.6: *3* 'contract-body-pane'
 @g.command('contract-body-pane')
-@g.command('expand-outline-pane')
 def contractBodyPane(event: LeoKeyEvent) -> None:
     """Contract the body pane. Expand the outline/log splitter."""
     c = event.get('c')
-    if not c:
-        return
-    r = min(1.0, c.frame.compute_ratio() + 0.1)
-    c.frame.divideLeoSplitter1(r)
-
-expandOutlinePane = contractBodyPane
+    if c:
+        c.frame.top.layout_cache.contract_body()
 #@+node:ekr.20200303084048.1: *3* 'contract-log-pane'
 @g.command('contract-log-pane')
 def contractLogPane(event: LeoKeyEvent) -> None:
     """Contract the log pane. Expand the outline pane."""
     c = event.get('c')
-    if not c:
-        return
-    f = c.frame
-    r = min(1.0, f.compute_secondary_ratio() + 0.1)
-    f.divideLeoSplitter2(r)
-#@+node:ekr.20200303084225.1: *3* 'contract-outline-pane' & 'expand-body-pane'
+    if c:
+        c.frame.top.layout_cache.contract_log()
+#@+node:ekr.20200303084225.1: *3* 'contract-outline-pane'
 @g.command('contract-outline-pane')
-@g.command('expand-body-pane')
 def contractOutlinePane(event: LeoKeyEvent) -> None:
     """Contract the outline pane. Expand the body pane."""
     c = event.get('c')
-    if not c:
-        return
-    r = max(0.0, c.frame.compute_ratio() - 0.1)
-    c.frame.divideLeoSplitter1(r)
-
-expandBodyPane = contractOutlinePane
+    if c:
+        c.frame.top.layout_cache.contract_outline()
+#@+node:ekr.20241027124820.1: *3* 'expand-body-pane
+@g.command('expand-body-pane')
+def expandBodyPane(event: LeoKeyEvent) -> None:
+    """Expand the log pane. Contract the outline pane."""
+    c = event.get('c')
+    if c:
+        c.frame.top.layout_cache.expand_body()
 #@+node:ekr.20200303084226.1: *3* 'expand-log-pane'
 @g.command('expand-log-pane')
 def expandLogPane(event: LeoKeyEvent) -> None:
     """Expand the log pane. Contract the outline pane."""
     c = event.get('c')
-    if not c:
-        return
-    f = c.frame
-    r = max(0.0, f.compute_secondary_ratio() - 0.1)
-    f.divideLeoSplitter2(r)
+    if c:
+        c.frame.top.layout_cache.expand_log()
+#@+node:ekr.20241027124753.1: *3* 'expand-outline-pane'
+@g.command('expand-outline-pane')
+def expandOutlinePane(event: LeoKeyEvent) -> None:
+    """Expand the log pane. Contract the outline pane."""
+    c = event.get('c')
+    if c:
+        c.frame.top.layout_cache.expand_outline()
 #@+node:ekr.20200303084610.1: *3* 'hide-body-pane'
 @g.command('hide-body-pane')
 def hideBodyPane(event: LeoKeyEvent) -> None:
--- a/leo/plugins/qt_frame.py
+++ b/leo/plugins/qt_frame.py
@@ -4285,49 +4285,80 @@ class TabbedFrameFactory:
             c.redraw()
     #@-others
 #@+node:ekr.20200303082457.1: ** top-level commands (qt_frame.py)
-#@+node:ekr.20200303082511.6: *3* 'contract-body-pane'
+#@+node:ekr.20241027140721.1: *3* contract-*-pane commands
+#@+node:ekr.20200303082511.6: *4* 'contract-body-pane'
 @g.command('contract-body-pane')
 def contractBodyPane(event: LeoKeyEvent) -> None:
     """Contract the body pane. Expand the outline/log splitter."""
     c = event.get('c')
     if c:
         c.frame.top.layout_cache.contract_body()
-#@+node:ekr.20200303084048.1: *3* 'contract-log-pane'
+#@+node:ekr.20200303084048.1: *4* 'contract-log-pane'
 @g.command('contract-log-pane')
 def contractLogPane(event: LeoKeyEvent) -> None:
     """Contract the log pane. Expand the outline pane."""
     c = event.get('c')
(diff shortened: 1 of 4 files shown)


tbpassin (commented):
I think that all the contract-* and expand-* can be done with one method of `LayoutCacheWidget`:

```
def change_pane_size(self, widget: str, expand: Bool) -> None:
    # No error checking or guards
    instance = self.get_widget(widget)
    splitter = instance.parent()
    index = splitter.indexOf(instance)
    factor = self.delta if expand else -self.delta
    self.rebalanceSplitter(splitter, index, factor)  # <=== All the work happens here
```

The various minibuffer commands would look like this:

```
@g.command('expand-outline-pane')
def expandBodyPane(event: LeoKeyEvent) -> None:
    """Contract the body pane. Expand the outline/log splitter."""
    c = event.get('c')
    if c is None:
        return
    cache = c.frame.top.layout_cache
    cache.change_pane_size('bodyFrame', True)
```

I'm inclined to think that there should be a separate but similar method of layout_cache for show/hide. We may need to special-case VR3 because it should execute set_freeze() when hidden but I'm not sure if it does - the hide() command is on the splitter and I don't know if it calls the client widget's `hide()` method or not.  A minor point.

tbpassin:
`Pass widgets, not their names, to expand/contract_pane`

Disagree. Someone has to know how to get the widget, and that is better done by the layout_widget.  Otherwise the minibuffer command has to try to find the right widget, and it will need to use its name to do that.  Just pass the name. It's simpler and easier to understand.

edreamleo:
@tbpassin It is the `LCW` class, not the minibuffer commands, that knows about widgets. Here is LCW.expand_body:
```python
def expand_body(self):
    """Expand the body pane"""
    self.expand_pane(self.c.frame.body.widget)
```
In other words, we agree about the infrastructure.

Thanks for your remarks about `change_pane_size`. Perhaps there is no need for `LCW.expand/contract_pane_by_name`.

tbpassin:
I bow to your inside knowledge... I didn't know there was such a thing as c.frame.body.widget,  Is there a c.frame.viewrendered3.widget? Where does c.frame get to know about viewrendered3.widget?

tbpassin:
> Replace QW by QWidget. LayoutClassWidget is (only!!) a QWidget

It's not a QWidget, it's a class derived from QWidget. I defined the QW bounded type after mypy complained about not recognizing the class. Up until then I thought that mypy would settle for QWidget for derived classes.

edreamleo:
@tbpassin mypy is happy with my changes. You annotations were unnecessarily complicated and confusing.

tbpassin:
> @tbpassin mypy is happy with my changes. You annotations were unnecessarily complicated and confusing.

I came up with those annotations and the QW and OD types because mypy wasn't accepting the type annotations I started out with, which included "QWidget" for classes derived from QWidget.

The typing error messages from mypy came when I ran mypy from the minibuffer.  If it gives different results when run externally, that's a problem with the command.  If it gives different results after updating mypy, that's not good either.  

It seems to me that when I run mypy from the minibuffer, mypy is more strict than when you use it.  The only mypy configuration I have is whatever is in the leo-editor git repo after a pull.  If you are using some different version, then we need to get sorted out how a developer can get the same configuration files as you use, and get the same results.

I don't want to be forced to run mypy externally on the whole repo just to check an external file I'm working on so I can work out the right typing annotations. They have been chewing up more than enough effort as it is.

edreamleo:
@tbpassin You're not responsible for mypy checks. I am. Don't worry about it.

Otoh, my `mypy-leo.cmd` takes only a few seconds. It seems inherently more reliable than checking a single file. I often run pylint from the command line but I never run mypy from the command line.

In short, the present annotations are good enough for me. Let's move on.

tbpassin:
> You're not responsible for mypy checks. I am. Don't worry about it.

Well, yes, OK.  But we have had some history about this.  When I've forgotten to run mypy and there are type annotation errors, you have complained that I should have run mypy.  This time I sweated blood to get mypy as run by me from the minibuffer to accept the annotations, and you say my solutions were too complicated. Now you say that mypy compliance is for you and for me not to worry about it.

So here's how I plan to proceed in the future unless you have a better suggestion.  I will put in whatever type annotations seem to make sense to me and I won't make them pass mypy as run by me.

The mypy command should still be fixed.  Also, now I'm wondering that if I run mypy from the minibuffer on an external file that is not part of the Leo directory tree, whether it will even be configured the way it should be.

tbpassin:
> Previously, I had assumed you could produce reasonable mypy annotations. Now, I don't. I don't know why you are having troubles.

I went by the mypy messages I got.  It's not like I'm looking for ways to complicate the annotations.  Since you didn't get the same mypy messages I got, you assume I'm doing annotations wrong, and I assume there is some reason for the discrepancies and it's not that I am incompetent or stupid.  Of course, I've been wrong before...

One of the things I've learned the hard way is that when I write

```
QWidget = QtWidgets.QWidget
widget: QWidget = get_widget('name')
```

on my system, when executed from the minibuffer, mypy will claim there's a type error.  On my system, when executed from the minibuffer, mypy will accept

`widget: QtWidgets.QWidget = get_widget('name')`

Or maybe not even then, depending on how `get_widget()` was annotated  This is perplexing to me because I thought a type system is supposed to be able to trace types through the assignments.

edreamleo:
@tbpassin You said you weren't looking for complications. That's laudable. Yet somehow you wrote this:
```python
class LayoutCacheWidget(Generic[QW], QtWidgets.QWidget):
```
That's utter nonsense. My guess this annotation sent both you and mypy down a ravine.

The only correct annotation is:
```python
class LayoutCacheWidget(QWidget):
```
Note that `QWidget`, the base class, needs no annotation!

The weird `_: str` annotations were a strong clue that something was seriously amiss.

But let's move on. Here are my recommendations:

- Never run Leo's `mypy` command from the minibuffer.
- Always run mypy using `leo/scripts/mypy_leo.py`
  I use `mypy-leo.cmd`. It contains: `python -m leo.scripts.mypy_leo`
- Start again if annotations get out of control.

HTH.

edreamleo:
@tbpassin One more recommendation. Study the annotations in Leo's gui files. They all follow roughly the same pattern. None need to go to heroic lengths to annotate any. Yes, there are times when `Any` annotations and `type:ignore` suppressions are necessary and appropriate. I usually avoid `Union` annotations: they are difficult to get right.

edreamleo:
@tbpassin Please approve this PR. It's time for everyone to start testing it.

tbpassin (changes requested):
qt_gui.py
------------
### Lines 1524 ff:

```
QSignalSpy = QtTest.QSignalSpy
    assert QSignalSpy
```
QSignalSpy is a new addition that doesn't seem to be used (yet) and has nothing to do with the rest of this PR.  Yet there is an assert.  If there is any possibility that the assert might fail, then this code will quit (I'm not sure if Leo will quit at that point or not). If there is an assert about this import, why not about other Qt imports.  also, what is this import doing in the middle of the module instead of its start?

I recommend removing the assert, and preferably the entire QSignalSpy parts, which should await a separate PR.

### Line 1536
`def find_parent_splitter(self, widget: QWidget) -> Optional[Tuple[QSplitter, Any]]:`

The type of the return should be 

`Optional[Tuple[QSplitter, QWidget]]`

### Line 1542
```
child is the *direct* child of the splitter that contains the widget.
          The child might or might not be the widget itself.
```

I find this terminology confusing. I also don't understand the reason for returning some widget that isn't the one we asked for. The "direct" child might not even be the parent of the widget we asked for. In principle, it could be e.g. the grandparent of our widget.  I don't know how to word it concisely, but the widget in question in the first descendant on the line of descent from the splitter to our widget. "Direct" didn't convey that to me.

qt_layout.py
----------------
### Line 430:
`a layout data structure such as the following:0`

The last character was probably meant to be a second ":"  for RsT literal block syntax.

### Line 478 ff
```
        if is_module_loaded(VR3_MODULE_NAME):
            from leo.plugins.viewrendered3 import controllers
            vr3 = controllers.get(c.hash())
            self.contract_pane(vr3)
```
It's not guaranteed that VR3 will be in a splitter.  When it is created it is made a child of the layout cache.  If a layout doesn't have a slot for it, it won't be in a splitter.  All the current layouts  do have a slot for it, but someone could create another one that doesn't.

I'm not sure if the same is true when the VR instance is created

edreamleo:
@tbpassin Thanks for your review!

- `assert QSignalSpy`: The code must stay as it is.
  The import and assert are inside `LeoQtGui.enableSignalDebugging`, a *debugging* method.
  The caller is responsible for any problems.
  Anyway, (except during startup) Leo will *never* exit as the result of a crash.

- `def find_parent_splitter(self, widget: QWidget) -> Optional[Tuple[QSplitter, Any]]:`
  The annotation is correct.

  The `LCW.find_parent_splitter` method returns a *child* (the so-called **direct child**) of the found splitter,
  *not* the `widget` arg, which might be a descendant.
  This little trick collapses the complexity of `LCW.resize_pane`!

- `a layout data structure such as the following:0`
  Good catch.

- It's not guaranteed that VR3 will be in a splitter.
  `LCW.contract_pane` does not require that guarantee.

I'll make one more change: The headline of `LCW.resize_pane` was `LCW.resize_widget`.

**Summary**

Thanks for your detailed reading of the code.
Rev 5042a9f changes one headline and one docstring.

tbpassin (approved):

Issue #4141, opened 2024-11-02
declutter rule WEIGHT line does not work for PyQt6


jknGH:
a declutter rule including a line like

`WEIGHT Bold`

has no effect in PyQt6. (a) the weight scaling seems to have changed (1--1000 instead of 1--99), and (b) the placement of Weight within QFont may have changed.

I will be submitting a Pull Request to fix this shortly

jknGH:
Hi @edreamleo - I tried to push a branch to address this and create a Pull Request, but I am getting a 403 error. I recently created a GitHub personal access token. Do I need to be granted permission to push a branch, or similar? Thanks, Jon N

![image](https://github.com/user-attachments/assets/a4acfb99-980d-4432-ae6b-6ee5a486cce5)

edreamleo:
@jknGH I've just sent an invitation for you to become a member.

jknGH:
Thanks, accepted ;-)

tbpassin:
The way I've been working is that I have my own GitHub account and it contains a clone/fork of the Leo repo.  I do all my work in a branch in my own repo.  When I push the branch from my computer to GitHub, GH offers to create a PR in the Leo repo.

edreamleo:
@jknGH You're welcome and welcome aboard.

edreamleo:
@tbpassin @jknGH Thomas's work flow has caused problems in the past and may do so in the future.

Instead, I strongly suggest you use the standard way of collaborating:
```console
git clone https://github.com/leo-editor/leo-editor.git
```

edreamleo:
@jknGH @tbpassin After cloning Leo's repo, you **must** base all your work on Leo's "devel" branch:
```console
cd <your cloned repo>
git checkout devel
git checkout -b <your branch name>
(do your work)
git add .
git commit -m "Your commit message"
git push origin <your branch name>
```
A good format for branch name is: `jknGH-<issue number>-<short description of issue>`

**Update** The first time you push to origin you will see a link in the console. Click on that link to create a PR.

HTH.

jknGH:
I'm not 100% sure I am seeing the difference between what @edreamleo and @tbpassin are saying here re. workflow.

I basically did this (like in my day job ;-)

`
# I already have devel locally
$ git switch -c jknGH-new-branch-name    # same effect as Edward's commands
# do my work
# ...
$ git push # git reminds me to use 'origin branch-name'
[...]

PR #4142, opened 2024-11-02
Jkn 4141 declutter

jknGH:
oops, probably should have written this earlier

This is to fix issue #4141, as discussed here: https://groups.google.com/g/leo-editor/c/K9FkwGXQltA

Two points:
* the default for the getattr has been changed - the font weight scaling is different in PyQt6
* I am unsure if the original `getattr(QtGui.Qfont, param...)` ever worked properly.

Tested by myself (under Linux) and @tbpassin 

worth squashing commits BTW ;-/

### DIFF
--- a/leo/plugins/qt_tree.py
+++ b/leo/plugins/qt_tree.py
@@ -306,18 +306,25 @@ class LeoQtTree(leoFrame.LeoTree):
 
             elif cmd == 'WEIGHT':
 
+                def weight_modifier_X(item: Item, param: str) -> None:
+                    arg = getattr(QtGui.QFont, param, None)
+                    if arg is None:
+                        arg = {"Thin": QtGui.QFont.Weight.Thin,
+                            "ExtraLight": QtGui.QFont.Weight.ExtraLight,
+                            "Light": QtGui.QFont.Weight.Light,
+                            "Normal": QtGui.QFont.Weight.Normal,
+                            "Medium": QtGui.QFont.Weight.Medium,
+                            "DemiBold": QtGui.QFont.Weight.DemiBold,
+                            "Bold": QtGui.QFont.Weight.Bold,
+                            "ExtraBold": QtGui.QFont.Weight.ExtraBold,
+                            "Black": QtGui.QFont.Weight.Black
+                            }.get(param, QtGui.QFont.Weight.Medium)
+                    font = item.font(0)
+                    font.setWeight(arg)
+                    item.setFont(0, font)
+
                 def weight_modifier(item: Item, param: str) -> None:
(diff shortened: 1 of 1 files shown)


tbpassin:
Just as a matter of interest, note that an integer (e.g., 500) can still be used instead of the enum expression (e.g., QtGui.QFont.Weight.Medium). Using the enum for the default is better because it directly communicates the intent.

jknGH:
> Just as a matter of interest, note that an integer (e.g., 500) can still be used instead of the enum expression (e.g., QtGui.QFont.Weight.Medium). Using the enum for the default is better because it directly communicates the intent.

Agreed, but I hardly see the point. In C, I am hardly going to write

`#define PI (3.14)`

and then do

`circumference = 2 * r * 3.14;`

BTW, I have confirmed that the getattr() never used to work properly. On a system with PyQt5, I changed my declutter pattern to

`WEIGHT Normal`

and the headline was still shown as bold (ie the '75' in effect).

Of course there is no guarantee that a given font has all of the options available. I didn't think it worth adding a note about this, eg. in LeoDocs.leo. You can probably expect Normal and Bold to be present, I guess.

tbpassin:
> Of course there is no guarantee that a given font has all of the options available

That's right. I think that Qt tries to come up with a font that's reasonably close if there isn't an exact match.  Some fonts can have non-existant weights created programmatically, I think, and some can't.

edreamleo (approved):
Thanks for this work.

edreamleo:
@jknGH Is everything ready for merging?

jknGH:
> @jknGH Is everything ready for merging?

@edreamleo I believe so, yes

jknGH:
@edreamleo, I was leaving the 'merge pull request action to you. Is that not the way you want it done?

edreamleo:
@jknGH Thanks for the clarification. I'll merge this PR now.

I usually like to do the merge myself. I then immediately run all of Leo's tests. If there are problems I fix them, create a PR and then merge the PR back into devel.

This workflow guarantees that everything works for me. The extra work is trivial. On a typical day I run all of Leo's tests dozens of time.

edreamleo:
@jknGH All my tests pass after the merge :-)

jknGH:
> @jknGH All my tests pass after the merge :-)

must be some new tests needed ;-)

(I think previously, a declutter line like eg.

```
WEIGHT light
```

would have given an actual bold font, if present. But I can understand there being no test for this

edreamleo:
@jknGH My main concern when running tests is that Leo's core remains stable.

The authors of PRs are responsible for new unit tests, but I'm a prime offender, so don't worry ;-)

Indeed, Leo's unit tests run in a null gui. Leo has no good framework for testing gui code.

Issue #4151, opened 2024-11-07
Freewin - Fix Cursor Behavior In Editing Mode


tbpassin:
Somewhere along the line an unpleasant behavior has crept into the Freewin editor view. After typing any text or character, the cursor jumps back to the start of the page. The same happens when a block of text is pasted.

I have a 1-line fix and will submit a PR after some more testing.

edreamleo:
@tbpassin Thanks for your testing, your report, and your upcoming PR.

edreamleo:
Fixed via PR #4157.

edreamleo:
@tbpassin pylint complained about a too-long line during my after-pull testing . Did you run pylint??

I fixed this complaint via rev ef52b6e.

Please pull from "devel" so you will have the latest code.

edreamleo:
@tbpassin In the future, run *all* tests as if your life depended on them.

tbpassin:
> Did you run pylint??

I ran full_test_leo.py. It runs ruff_leo in lieu of pylint:

```
for command in [
    fr'{python} -m "leo.scripts.beautify_all_leo',
    fr'{python} -m "leo.scripts.run_test_leo',
    fr'{python} -m "leo.scripts.mypy_leo',
    fr'{python} -m "leo.scripts.ruff_leo',
    # fr'{python} -m "leo.scripts.pylint_leo',
]
```

Since mypy gives different results when run from the minibuffer than when run in full_test.py, and you told me to use full_test and not the minibuffer for these tests, no I didn't run pylint.

tbpassin:
> In the future, run _all_ tests as if your life depended on them.

I did what you told me to. You blew off the mypy-from-minibuffer messages I keep getting. full_test.py doesn't even run pylint any more. So why would I even consider running pylint from the minibuffer any more?

edreamleo:
@tbpassin I don't suggest you run tests from the minibuffer. I run pylint from `pylint-leo.cmd`:

```console
@echo off

cd c:\Repos\leo-editor

echo pylint-leo.cmd
time /T
rem pylint checks that the rcfile exists.
rem call python -m leo.scripts.pylint_leo  --rcfile C:\Users\Dev\.leo\.pylintrc %*
call python -m leo.scripts.pylint_leo
time /T
```

Please do something similar.

tbpassin:
```
C:\Tom\git\leo-editor\leo\scripts>pylint_leo.py
pylint_leo.py
[...]

PR #4157, opened 2024-11-09
Fix freewin cursor bug

tbpassin:
- [x] Position cursor in FW editor to match host node's cursor position when host's text changes;
- [x] Scroll editor if needed when new cursor position is off-screen;
- [x] Update docstring with version, What's New;
- [x] No messages about freewin.py when full_test_leo.py is run.

Resolves Issue https://github.com/leo-editor/leo-editor/issues/4151


edreamleo (approved):

edreamleo:
@tbpassin Thanks for this PR.

Issue #4160, opened 2024-11-10
PylintCommand Computes Wrong rc-file Directory


tbpassin:
See PR #4163.

The `get_rc_file` method in the PylintCommand class of checkerCommands.py computes one of the fallback rc-file paths incorrectly:

```
    table = (
        # In the directory containing the outline.
        g.finalize_join(local_dir, base1),
        g.finalize_join(local_dir, base2),
        # In ~/.leo
        g.finalize_join(g.app.homeDir, '.leo', base1),
        g.finalize_join(g.app.homeDir, '.leo', base2),
        # In leo/test
        g.finalize_join(g.app.loadDir, '..', '..', 'leo', 'test', base2),  # <--- should be base1
        g.finalize_join(g.app.loadDir, '..', '..', 'leo', 'test', base2),
    )
```

edreamleo:
@tbpassin Many thanks for this report. I apologize for the confusion it may have caused you. I'll fix this bug immediately.

edreamleo:
Closed via PR #4163.

PR #4163, opened 2024-11-11
PR 2 for #4160: fix a path in pylint.get_rc_file

edreamleo:
See #4160.

### DIFF
--- a/leo/commands/checkerCommands.py
+++ b/leo/commands/checkerCommands.py
@@ -653,7 +653,7 @@ class PylintCommand:
             g.finalize_join(g.app.homeDir, '.leo', base1),
             g.finalize_join(g.app.homeDir, '.leo', base2),
             # In leo/test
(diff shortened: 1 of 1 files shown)


edreamleo:
@tbpassin This PR is ready for review. It passes my full suite of tests.

tbpassin (approved):

tbpassin:
This change fixes the issue, of course.  However, the leo/test directory does not contain an rc file with the name being searched for.  Instead, it contains an rc file with this name: pylint-leo-rc-ref.txt. I wonder if this discrepancy is intentional or not. The way things are, a search that got as far as the leo/test directory would only succeed if the user had made a copy of pylint-leo-rc-ref.txt with the name that is being searched for.

edreamleo:
@tbpassin The name `pylint-leo-rc-ref.txt` is intentional. Users should create their own rc file.

tbpassin:
OK, that's what I suspected.

Issue #4164, opened 2024-11-11
Disable the "no popup handlers" popup


edreamleo:
See PR #4165. This unnecessary intrusion is annoying.

edreamleo:
Closed via PR #4165.

PR #4165, opened 2024-11-11
PR for #4164: Disable the 'no popup' popup

edreamleo:
See #4164. This PR seems to enable useful popups rather than entirely suppressing them.


edreamleo:
@tbpassin How does this PR work for you?

tbpassin (approved):
Fine by me (meaning the code review).

as for
> How does this PR work for you?

I've never seen the "no popups" popup message but I like the idea of getting rid of it.

In some of my programs I deal with this kind of situation by having a message bar at the bottom of the program window.  A harmless message appears there and then slowly fades away after several seconds.  A warning or error message appears there with a distinctive background, flashes a few times, and then slowly fades away.  This kind of message is not obtrusive but can attract the user's attention if it's serious, and requires no user action to make go away.  If the user wants a second chance to read the message he can repeat the action that caused the message.

edreamleo:

Issue #4193, opened 2024-11-18
execute-script should require selected text unless @language python is in effect


edreamleo:
See PR #4197.

`Ctrl-B` should execute text in languages other than python only there is *explicitly selected* text.

For example, executing all markdown text will almost always throw a SyntaxError.

tbpassin:
The ability of the command to execute selected lines with Python is handy and I hope we won't lose it.

edreamleo:
Fixed via PR #4197.

PR #4197, opened 2024-11-19
PR for #4193: enhance execute-script

edreamleo:
See #4193.

- [x] `c.executeScript`: all args must be kwargs.

### DIFF
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -1183,8 +1183,8 @@ class Commands:
             del sys.path[:2]
     #@+node:ekr.20171123135625.4: *3* @cmd execute-script & public helpers
     @cmd('execute-script')
-    def executeScript(
-        self,
+    def executeScript(self,
+        *,
         event: LeoKeyEvent = None,
         args: Any = None,
         p: Position = None,


edreamleo:
@tbpassin This PR is ready for review. I think it fixes the hangnail while retaining the ability to execute selected text.

tbpassin (commented):
Lines 1220 ff
---------------
```
            valid = (
                not c.forceExecuteEntireBody
                and useSelectedText
                and has_selection
                and language in ('jupytext', 'md', 'pandoc', 'python', 'rest')
            )
```
What is supposed to happen when the selected text is one of the others that isn't "python"?  E.g., if the language in effect is "md", what will "execute" mean?

edreamleo:
@tbpassin The original "ready-for-review" was wrong. It would prevent execution of python scripts w/o a selection!

To answer your question, the new code allows execution of selected text, regardless of language. Imo, is simplest and best. Regardless of *stated* language (via `@language`), we can imagine *any* selected text might contain valid Python code.

Finally, I have simplified and clarified the calculation of the "valid" test.

edreamleo:
@tbpassin The original "ready-for-review" was wrong. It would prevent execution of python scripts w/o a selection!

To answer your question, the new code allows execution of selected text, regardless of language. Imo, is simplest and best. Regardless of *stated* language (via `@language`), we can imagine *any* selected text might contain valid Python code.

Finally, I have simplified and clarified the calculation of the "valid" test.

PR #4205, opened 2024-11-24
VR3 - Restore Scroll Position After Updates

tbpassin:
With this PR, VR3 restores the scroll position of the rendered view. This is a tricky operation because every time that setHtml() is called, the QWebEngineView widget (qwev) displays the new page scrolled to the page top. There is no method to directly control the scroll position, unlike with widgets such as the QTextBrowser.  Instead, VR3 connects to the page load event, uses javascript to obtain and store the current scroll position, and uses javascript in the Chrome browser component to scroll to the saved scroll position.

VR3 then declines to update until the actual scroll position matches the target scroll position.  For long renderings, such as entire trees that contain images and are slow to load, this technique reduces (but does not eliminate altogether) visual page jumps. After all changes have rendered, the view will have scrolled back to the target scroll position.  This solves a long-standing UX annoyance.  A more complete solution would involve double-buffering, which is more complicated.

This PR also removes some kinds of renderings including @graphics, @html, @networx, and others that probably have not worked in years. They could be reworked if anyone wants. The docstring has been updated and slightly restructured.

- [x] Scroll position in rendered view is now restored after updates.
- [x] Version and docstring updated.
- [x] Obsolete rendering kinds removed.
- [x] The Default Kind menu now is initially set to the same kind as the setting _vr3-default-kind_.
- [x] Some Items in the Default Kind menu have been re-ordered. 
- [x] full-test.py: 2 skipped tests, no other messages.
- [x] pylint: no messages.

### DIFF
--- a/leo/plugins/viewrendered3.py
+++ b/leo/plugins/viewrendered3.py
@@ -2299,8 +2299,7 @@ class ViewRenderedController3(QtWidgets.QWidget):
         self.last_update_was_node_change = False
         self.setObjectName('viewrendered3_pane')
         self.adapt_fgbg_colors = True
-        self.qwev.loadFinished.connect(self.restore_scroll_position)
-        self.scroll_position = 0
+        self.scroll_position: int = 0
 
         #@-<< initialize configuration ivars >>
         #@+<< asciidoc-specific >>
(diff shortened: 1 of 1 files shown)


edreamleo (approved):

edreamleo:
@tbpassin Rev 365f4ba beautifies VR3. No big deal :-)

tbpassin:
My run of full-test had also found something to beautify in VR3.

Issue #4208, opened 2024-11-25
Improve make-sphinx script


edreamleo:
Improve `make-sphinx` script in `LeoDocs.leo` as follows:

- Check that the python version and the sphinx-build versions match.
  At present, EKR's internal sphinx-build cmd script uses a hard-coded version of the path to `python/scripts/sphinx-build.exe`
  This is a recipe for great confusion.
- Automatically update the "Last updated" and "Created using" fields on Leo's home page: `index.html`.

edreamleo:
Closed via PR #4209.

PR #4209, opened 2024-11-25
PR for #4208: improve @button make-sphinx in LeoDocs.leo

edreamleo:
See #4208.

**Part 1: Ensure `sphinx-build` runs with the proper version of Python**

- [x] Change EKR's internal `sphinx-build.cmd` to `python -m sphinx %*`
   Many thanks to @tbpassin for this suggestion. It simplifies everything!
   See the `make_html` function in the `@button make-sphinx` script for details.

I had blind spot concerning `python -m`. It will "just work" for any pip installed package. No need to worry about paths!
- [x] Move `leo/scripts/sphinx_build.py` to the attic.

**Part 2: update index.html automatically**

- [x] Automatically update the "Last updated" and "Created using" fields on Leo's home page: `index.html`.
    See the `patch_home_page` home function in the `@button make-sphinx` script.


tbpassin (commented):
I believe that sphinx can be induced to automatically include the build date.  This would avoid the need to patch the home page and to manually remember to change the dates anywhere.

edreamleo:
@tbpassin sphinx already updates the build dates for all of Leo's website. But sphinx does not build `index.html`, nor do I see how it could easily be made to do so.

In other words, `index.html` is a bit of a hack, but getting rid of that hack seems like way more work than the `patch_home_page` function. In any case, having sphinx manage `index.html` would be another issue.

tbpassin:
Aw, shucks, just `python -m sphinx`...

Most of my Sphinx projects don't even use a config file nor make.  Here's a typical button command to generate the docs:

```
@language python
from sys import executable
import subprocess
import os

target = None
HEAD = "GF4 User's Guide"
p0 = c.p
    if p.h.startswith(HEAD):
        target = p
        c.selectPosition(target)
        break
if target:
    c.k.simulateCommand('rst3')

    # The rst3 command writes to correct (@path sphinx) directory
    # but Sphinx will look in current directory, which may
    # not be the same.  So -
    # if we start from a selected node outside the
    # @path sphinx tree, temporarily cd to the docs directory.
    cwd = os.getcwd()
    if not cwd.endswith('sphinx'):
        temp_cwd = os.path.join(cwd, 'sphinx')
        os.chdir(temp_cwd)

    # Other likely themes: 
    #'-D', 'html_theme=sphinx_book_theme',
    #'-D', 'html_theme=bizstyle',

    cmd = [executable, '-m', 'sphinx', '-C',  # -C means no config file
            '-b', 'html',
            '-D', "master_doc=GF4_Users_Guide",
            '-D', 'source_suffix=.rst',
            '-D', 'html_theme=pyramid',
            '-D', 'project=GF4',
            '-D', 'extensions=sphinx.ext.autosectionlabel',
            '-D', 'copyright=Thomas B. Passin 2022',
            '-D', "html_theme_options.sidebarwidth=20em",

            # sourcedir, outputdir:
            r'.', r'..\docs']

    subprocess.call(cmd)
else:
    g.es('Cannot find the @rst tree to process', color = 'red')
```

Issue #4224, opened 2024-11-30
Support .m files in @jupytext nodes


edreamleo:
This issue was a brain spike. `@clean x.m` works fine.

edreamleo:
Closed as invalid.

PR #4225, opened 2024-11-30
PR for #4224: Support .m and .ipynb files in @jupytext

edreamleo:
See #4224.

This PR and the related issue were brain spikes.

- [x] Treat `@jupytext x.y` as a kind of `@clean` node when `y` is `.m` or `.ipynb`.
   This special case is significantly (and surprisingly) better than using `@clean` directly.


edreamleo:
I am closing this PR. It was a thoroughly bad idea.

PR #4252, opened 2024-12-11
PR: add show-buttons-and-at-commands command

edreamleo:
This new command contains useful information.

- [x] Add `show-buttons-and-at-commands` command.
- [x] Remove prototype from the attic.


edreamleo:
@tbpassin This command is ready for review.

The command shows how to search open commanders looking for `@command` and `@button` nodes.

tbpassin (commented):
leoKeys.py
-----------------
Lines 2464 ff -

I don't favor changing the historic name of a command without a good reason and I don't see a compelling reason to do so with this command.   Let's keep the command _show-buttons_  and not rename it to _show-buttons-legacy_. Also, how about keeping the old command _show-@buttons-and-@commands'_ as an alias of the new name _show-buttons-and-at-commands_? 


lines 2667: `def parse_setting()` -

This method contains two asserts. Usually one tries to avoid asserts in production code.  Why can't the method return None or (None, None) instead of failing?

edreamleo:

I agree that the `show-button` command should remain. It may appear in `@data history-list` nodes.
I'll also restore the function names&mdash; user scripts might call those functions.

A recent PR added the `show-@buttons-and-@commands` alias. Imo, there is no reason to retain it.

edreamleo:
@tbpassin I've made most of the changes you requested. Please approve this PR.

Leo's `git-diff-pr` command shows only the following:

- Add the new command: `show-buttons-and-at-commands`.
- Change headlines so they show the command names.
- Remove the *recent* `show-@buttons-and-@commands` alias for the old command.
  This alias would be confusing: it's too close to the new command name.

tbpassin (approved):
> Remove the recent show-@buttons-and-@commands alias for the old command.
I agree with this.  It was a hard name to work with anyway.

edreamleo:
@tbpassin Thanks, Thomas, for your help.

Issue #4284, opened 2025-01-20
Find command does not allow escaping /n


edreamleo:

edreamleo:
Closed via #4287.

PR #4287, opened 2025-01-24
PR for #4284: Allow back slashes in find strings

edreamleo:
See #4284.

- [x] Rewrite `find.replace_back_slashes`.
- [x] Revise `TestFind.test_replace_back_slashes`.
- [x] Test with plain and regex finds.

### DIFF
--- a/leo/core/leoFind.py
+++ b/leo/core/leoFind.py
@@ -2602,6 +2602,8 @@ class LeoFind:
                 result.append('\n')
             elif ch == 't':
                 result.append('\t')
+            elif ch == 'f':
+                result.append('\f')
             elif ch == '\\':  # 4284
                 result.append(ch)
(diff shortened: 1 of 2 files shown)


tbpassin (approved):
leoFind.py
-------------
Line 2596ff: `if ch != '\\' or i >= len(s):`
This line is logically correct but it contains two completely different kinds of test.  I would prefer that the two tests be separated.  For example:
```
        while i < len(s):
            progress = i
            ch = s[i]
            if i >= len(s):
                result.append(ch)
                break
            i += 1
            if ch != '\\':
                result.append(ch)
                continue
            ch = s[i]
```

edreamleo:

Rev a27db64 adds support for representing `\f`. That capability would been useful recently.

Issue #3236, opened 2023-04-02
Add a command to navigate to the source of any command


tbpassin:
See PR #4296.

Buttons added by the "Script Button" button have a context menu item to navigate to their defining script.  This is very useful.  I suggest that it would be equally useful to have a command that would navigate to the source of any minibuffer command.  It would work much like the F11 command, but instead of displaying the command's docstring it would navigate to the actual script.

edreamleo:
@tbpassin Great idea.

tbpassin:
I see it as working like the F11 command - you get tab completion to zero in on the right command.

I think the @button buttons can actually open the source outline, too (not 100% sure, but if they use the full UNL it would work that way).

tbpassin:
We would need to compute a gnx for a command given the command's file and line number, which we can get as follows:

```
import inspect

d = c.commandsDict
k = 'vr3'
f = d[k]

filename = inspect.getfile(f)
line_number = inspect.getsourcelines(f)[1]

g.es(filename, line_number)
```

Too bad this won't work for commands decorated by `@cmd`, like _listen-to-log_: the function returned is `new_cmd_wrapper`. I don't know if there is a practical way to get from the wrapper to the actual function.

edreamleo:
@tbpassin I just search for `'<command-name>'`. The convention is that Leo's decorator's all use single quotes to define commands. So there is no great need for this command, but your proposed command can probably do the search.

edreamleo:
Completed via PR #4296.

PR #4296, opened 2025-02-06
PR for #3236: find-source-for-command

edreamleo:
See #3236.

- [x] `c_help.openLeoPy` and `c_help.openLeoPyRef` return an optional `Commands` instance.
- [x] `find.find-source-for-command` looks in various files for patterns.
- [x] Verify that the command can find commands in all the expected places.
- [x] All tests pass.


edreamleo:
@tbpassin This PR searches various files for decorators and nodes that define commands.

This approach opens files temporarily, but I think this approach is best. What do you think?

tbpassin:
> This approach opens files temporarily, but I think this approach is best. What do you think?

I think these outlines could be opened with a null GUI for greater speed.  The a pattern is matched, then the outline would be re-opened with the usual GUI.  This approach is already used for reading settings when an outline is opened so it ought to be easy. My review comments include this suggestion.

edreamleo:
@tbpassin Thanks for your comments.  I'll do both of your suggestion.
 
However, today Rebecca and I will be leaving on our trip, so it may take awhile.

edreamleo:
@tbpassin Thanks for your comments.  I'll do both of your suggestion.
 
However, today Rebecca and I will be leaving on our trip, so it may take awhile.

tbpassin:
> Rebecca and I will be leaving on our trip

Happy travels!

PR #4298, opened 2025-02-19
Fix spurious error from createImporterData if '~/.leo/plugins' exists

spikethehobbitmage:
If '~/.leo/plugins' exists then createImporterData incorrectly attempts to load 'leo.plugins.importer.plug'. This patch applies the same test createWritersData uses.


edreamleo (approved):
Thanks for this fix.

PR #4303, opened 2025-02-28
Add minimal colorizer for Typst documents

gatesphere:
Add a minimal colorizer mode for Typst documents.

This is a hand-written colorizer, as I could not find a jEdit definition for Typst.  I release it into the public domain.

All tests passed on my machine.


edreamleo (approved):
Looks good. Thanks for this work.

PR #4314, opened 2025-03-13
Scripts, @commands and @buttons, can now have a return value.

### LINKED ISSUE #4313: If possible, allow for scripts, @commands and @buttons, etc to have a return value.
Maybe I dreamt this feature and woke up still thinking it was real,  but I implemented this in LeoJS, thinking it was already like that in Leo. Heh!

Maybe it's not possible in python, but I'll try. 

Please take a look at the draft pull request I made for this and critique / comment.

Thanks! 😄

boltex:
Allows for CTRL+b in a node which contains:

```
g.es('Hi! running a command!')
a = c.doCommandByName('fil-test-cmd') + 1
g.es(a)
```

Where the node @command fil-test-cmd contains:

```
g.es('hi from cmd')
x = 2
y = 40

def mySum(a, b):
    return a+b

value = mySum(x, y)
g.es(value)
return value
```

Fixes #4313

### DIFF
--- a/leo/config/leoSettings.leo
+++ b/leo/config/leoSettings.leo
@@ -520,6 +520,7 @@
 <v t="ekr.20210506073820.1"><vh>@bool allow-text-zoom = True</vh></v>
 <v t="ekr.20241012050620.1"><vh>@bool diff-leo-files = True</vh></v>
 <v t="ekr.20131112150804.18737"><vh>@bool force-execute-entire-body = False</vh></v>
+<v t="felix.20250313171245.1"><vh>@bool allow-script-return-result = False</vh></v>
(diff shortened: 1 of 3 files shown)


edreamleo:
@boltex This PR seems like a good idea. Please let me know when it is ready for review.

boltex:
I've finalized this branch by having it be 100% backward compatible. It defaults to trying to return only a 'result' global variable, and otherwise requires a config setting 'allow-script-return-result' to be true to actually modify the script before running it within a 'main' function.

edreamleo:
@boltex Thanks for this PR.

edreamleo:
@boltex rev 6aea811 fixes several valid pylint/beautifier complaints.

Issue #4306, opened 2025-03-01
write-zip-archive command creates folders from root up to containing folder.


boltex:
See PR #4318.

OP: write-zip-archive command ~~weirdly~~ creates folders from root up to containing folder.

EKR: This is expected. The behavior described prevents name clashes. The `arcname` kwarg to [`zipfile.ZipFile.write`](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.write) can be used to suppress the enclosing folders, but it seems unwise to use this kwarg for all files.  See the PR.

<details>

_I detected that by testing leojs, turns out I had simply copied Leo's behavior!_  :) 

Can be easily reproduced by doing the command from a leo file that is in a folder which is not directly on the root. 

e.g.:

c:/my-folder-1/my-folder-2/my-folder-3/my-file.leo

instead of 

c:/my-folder/my-file.leo

This will leave you with a zip archive that contains all those sub folders before actually getting to the last one which has the zipped content. 

(this command was probably tested in a folder directly on the root, so the issue was not apparent) 

</details>

PR #4318, opened 2025-03-17
PR for #4306: quirp involving write-zip-archive

edreamleo:
See #4306.

This PR will remain a draft while we discuss the subtleties.

Leo's existing `write-zip-archive` command is arguably good enough.  [`zipfile.ZipFile.write`](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.write) puts files in folders to avoid problems with name clashes.

- [x] Simplify the logic and error handling.
- [x] Use the `arcname` kwarg to suppress the folder for the `.leo` file itself.

### DIFF
--- a/leo/core/leoFileCommands.py
+++ b/leo/core/leoFileCommands.py
@@ -660,7 +660,7 @@ class FileCommands:
         c = self.c
         leo_file = c.fileName()
         if not leo_file:
-            print('Please save this outline first')
+            g.es_print('Please save this outline first')
             return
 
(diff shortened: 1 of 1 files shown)


edreamleo:
@boltex This PR is ready for review. I'm not sure the new code is an improvement.

boltex:
Thanks, I'll try it out this weekend

edreamleo:
Per discussions with boltex, I am going to close this PR. No change is necessary.

Issue #4323, opened 2025-04-01
Fix bugs involving file delimiters and syntax coloring


edreamleo:
Two (usually rare) bugs have recently become apparent. Each is an unusual edge case in complex code.

**Bug 1**: Depending on clones, nodes without explicit `@language` directives may be colored with the wrong language.

Scenario:  Add and `@language rest` directive at the start of an organizer *without* adding an `@language python` directive at the end of the node. When coloring a cloned node *outside* of the organizer node, Leo may colorize the clone using rest, not python.

Instead, Leo's colorizer should use the language implied by the nearest ancestor `@<file>` node. Only if there are no such ancestor `@<file>` nodes should the colorizer use the existing algorithm to determine the language in effect.

**Bug 2**: Leo may not properly handle `@<file>` nodes that do *not* contain an `@language` directive.

Imo, all `@<file>` nodes *should* contain an `@language` directive.
The question is, what should happen if they do not?

**Discussion**

The first bug is highly confusing when it happens.
The second bug is probably worse. PR #4324 fixes the first bug.

The logic involved is complex. Careful study and testing will be required. I would rather do nothing than create more problems!

Any fixes are likely to affect LeoJS.

edreamleo:
I am closing this issue. It appears to be invalid.

edreamleo:
Fixed via PR #4326.

edreamleo:
Fixed via PR #4326.

PR #4324, opened 2025-04-01
PR for #4323: Replace at/c.scanAllDirectives

edreamleo:
This PR contains a thorough cleanup of Leo's directive-related code while maintaining *full* compatibility with legacy code.

Any discrepancies will be considered bugs to be fixed in later (much smaller) PRs.

**Summary of changes** 

- Adds the following **new getters**:
  `c.getDelims`, `c.getEncoding`, `c.getLanguage`, `c.pageWidth`, `c.getPath`, `c.getPageWidth`, and `c.getTabWidth`.
  Leo's core and all official plugins use these methods instead of deprecated functions & methods (see below).
- Adds the `p.findDirective` helper.
- Deprecates `c.scanAllDirectives` and many directives-related functions in `leoGlobals.py`.
- All newly-deprecated functions and methods call the new `g.deprecated` function.
- Document the new getters and declare that `c.scanAllDirectives` is deprecated.

<details><summary><b>Part 1: Revise the read/write code in leoAtFile.py</b></summary>
<br>

- [x] Add `AtFile.__slots__`.  This enforces some sanity.
- [x] Remove unused ivars.
- [x] Init and annotate all ivars in `at.__init__`.
- [x] Create `at.init_all_ivars`.
- [x] `at.scanAllDirectives`: Re-write the logic that computes `at.language`.
   The issues involved are subtle:
   `at.scanAllDirectives` should *never* examine ancestor `@<file>` nodes.
   - **Reading**
     - `p.b` is *empty*. The legacy code worked by *accident*.
     - `@language` directives do *not* exist yet.
     - `FastAtRead.scan_lines` (*not* `at.scanAllDirectives`) sets comment delims
  - **Writing**
     - `@comment` directives in `@<file>` nodes *do* guarantee that language delims are set properly.
- [x] `at.scanAllDirectives`: Use `python` as the language for `@button` and `@command` nodes.
- [x] Add `at.delimsFromAtFileNodeBody`, `at.languageFromAtFileNodeBody`, and `at.languageFromAtFileNodeHeadline`.
    These new helpers ensure that the revised `at.scanAllDirectives` does not change any other existing code.
- [x] Add `g.language_from_headline`.
- [x] Simplify the `<< handle @first and @last >>` case in `FastAtRead.scan_lines`.
- [x] Beautify six plugins, changing `#coding` to `# coding` in three files.
   These changes could have been avoided, but all unit tests still pass.
- [x] Change unit tests involving non-python files: don't honor `@language` directives in `.txt` files.

</details>

<details><summary><b>Part 2: deprecate c.scanAllDirectives and all helpers using directives lists</b></summary>
<br>

- [x] Define `g.deprecate` and `g._context`.
- [x] Add calls to `g.deprecate` to all deprecated functions/methods.
- [x] Add new helpers to the `Commands` class.
- [x] Don't call `at.scanAllDirectives`. Use new helpers instead.
- [x] Don't call `c.scanAllDirectives`. Use new helpers instead.

</details>

**About compatibility**

Neither the diffs nor Leo's unit tests ensure that this PR is equivalent to Leo's legacy code.
I had to compare the new getters with the corresponding code in `g.get_directives_dict`:

<details><summary><b>Folding code from g.get_directives_dict into the new g.getX methods</b></summary>
<br>

I "distributed" the following lines from `g.get_directives_dict` to the new getters:
```python
# The headline has higher precedence because it is more visible.
for p2 in p.self_and_parents():
    for s in (p2.h, p2.b):
        for m in c.at_<whatever>_pattern.finditer(s):
```
These lines ensure that Leo searches for directives first in headlines, then in the body pane.

In addition, I copied one warning from `get_directives_dict` to `c.getPathFromNode`.

</details>


tbpassin (commented):
This is a huge PR, and I didn't review any plugins except for VR3.

It seems to me that in this PR, searches for the root of an atfile start from the top of the outline rather from the current node.  Is this right?  If so, isn't this a change from past behavior? If this is correct, it makes me a little nervous in case an at-file accidentally gets nested inside another one.

gotoCommands.py
-----------------------
Lines 376ff:
```
        try:
            c.target_language = c.getLanguage(root)
            delims = c.getDelims(root)
        finally:
            c.target_language = old_target_language
        delims1, delims2, delims3 = delims
```
1. Suggest moving the first line after the `finally` block to be the last line of the try block:

```
        try:
            delims = c.getDelims(root)
            delims1, delims2, delims3 = delims
        finally:
            c.target_language = old_target_language
```
2. Can `c.getDelims()` fail?  If yes, what should happen?  if no, why do we have a `try` block?
3. Why do we return the target language to be the old target language?  Won't the delims be wrong for the old target language?

leoAtFile.py
-------------
Lines 184ff - Why is `initReadIvars()` kept instead of being replaced by `initAllIvars()`? All it now does is to call the latter.  The docstring is also misleading because the method does more than init just the read IVars.

Line 276 - `not g.unitTesting and at.language`
Suggest adding parens to make clear what is being negated.  Otherwise the meaning is not clear to the reader who doesn't remember the operator precedence. IOW, do we intend `(not g.unitTesting) and at.language` or `not (g.unitTesting and at.language)`?
 
Line 374 - ` # Open the file.` This comment is obvious from the command and should be omitted.

Lines 580, 584, 686 - more unnecessary comments.

leoGlobals.py
-------------------

line 1590 - the deprecation message would be more useful if it said what replacement method to use.

boltex:
@edreamleo   Thanks! I'll review and add this to leojs!

edreamleo:
@tbpassin Many thanks for your detailed review.

About `gotoCommands.py`:

1. The structure code is exactly the same as before. Only the specific calls are different.
    This code will fail if and only if `c.getLanguage(root)` fails, and that can't happen.
    The change you suggest would be wrong: it would use `c.target_language` instead of `c.getLanguage(root)`.
2. Using a `try/except` block is recommended Python style to emphasize that `c.target_language` is restored no matter what.
3. The code restores `c.target_language` because it is a cached value of a setting.
    It is absolutely essential that the code in question never modifies `c.target_language`!

edreamleo:
@tbpassin About `leoAtFile.py`:

The name `at.initReadIvars` is correct even though it is equivalent to `at.initAllIvars`.
This equivalence is accidental and conceivably could change in the future.
The *functions* of these two methods are different.

`if not g.unitTesting and at.language:`  This code has been around for decades. There is no need to change it.

Various comments: These comment should stay. They help structure lengthy code.

edreamleo:
@tbpassin About `g.deprecated`:

I agree that suggesting an alternative would be useful. However, the suggestion might depend on context.
In particular, the alternative for `c.scanAllDirectives` might be any one of the new `c.getX` getters.

`g.deprecated` was useful during refactoring, and as a marker for deprecated functions and methods.
In any case, I'm not going to delay this PR to enhance `g.deprecated`.

tbpassin:
OK. I posed my comments as questions or suggestions because I wasn't sure how the code was supposed to work. Do you want me to approve the PR? Or will @boltex's approval be enough?

edreamleo:
@tbpassin I would appreciate your approval. However, I'll wait for @boltex to approve as well.

edreamleo:
@tbpassin ""It seems to me that in this PR, searches for the root of an atfile start from the top of the outline rather from the current node. Is this right?".

I'm still not sure what you meant, but a review of the diffs will show:

- There are no changes to the arguments to the `c.getX` getters.
- All the `c.getX` getters search the direct ancestors of `p`.
  This is because the PR distributes the code from `g.get_directives_dict` to each `c.getX` getter.
  In other words, everything should work *exactly* as before.  Unit tests still work, though that is hardly conclusive.

edreamleo:
@tbpassin ""It seems to me that in this PR, searches for the root of an atfile start from the top of the outline rather from the current node. Is this right?".

I'm still not sure what you meant, but a review of the diffs will show:

- There are no changes to the arguments to the `c.getX` getters.
- All the `c.getX` getters search the direct ancestors of `p`.
  This is because the PR distributes the code from `g.get_directives_dict` to each `c.getX` getter.
  In other words, everything should work *exactly* as before.  Unit tests still work, though that is hardly conclusive.

tbpassin (approved):

edreamleo:
@tbpassin Thank you Thomas. I just remembered that I want to add a few more unit tests for the new getters. I'll add them before merging.

boltex:
@edreamleo  Maybe not directly related to this pull request, but In editCommands.py, what is the 'path_for_p' method for? (It's using 'c' as a parameter instead of 'self.c', and feels misplaced or orphaned)

edreamleo:
@boltex I have no knowledge of `ec.path_for_p`.

The node sentinel starts with `#@+node:tbrown.20151118134307.1:`, so Terry wrote it.

leo/core/leoAtFile.py
```diff
@@ -46,34 +77,56 @@ class AtFile:
...
+        self.indent = 0  # write indentation, in blanks.
+        self.outputFile: io.StringIO = None
+        self.outputList: list[str] = []
+        self.sentinels = False
+        self.section_delim1 = '<<'
+        self.section_delim2 = '>>'
+        self.targetFileName: str = ''
+        self.unchangedFiles = 0
+        # User settings.
+        self.at_auto_encoding = 'utf-f'
```
boltex, line 116: Is 'utf-f' a thing? PErhaps it was meant to be utf-8.

leo/core/leoAtFile.py
```diff
@@ -2540,27 +2596,25 @@ def precheck(self, fileName: str, root: Position) -> bool:  # pragma: no cover
...
+            return True
         # #1450: First, check that the directory exists.
         theDir = g.os_path_dirname(fileName)
         if theDir and not g.os_path_exists(theDir):
             at.error(f"Directory not found:\n{theDir}")
             return False
-        #
         # Now check the file.
         if not at.shouldPromptForDangerousWrite(fileName, root):
             # Fix bug 889175: Remember the full fileName.
```
boltex, line 2608: Should this 'bug fix' be replaced with a single '#' as was done in other instances of this commit for this file?

leo/core/leoAtFile.py
```diff
@@ -2540,27 +2596,25 @@ def precheck(self, fileName: str, root: Position) -> bool:  # pragma: no cover
...
             at.rememberReadPath(fileName, root)
             return True
-        #
         # Prompt if the write would overwrite the existing file.
         ok = self.promptForDangerousWrite(fileName)
         if ok:
             # Fix bug 889175: Remember the full fileName.
```
boltex, line 2614: same as 5 lines above. "bug fix" to "#".

leo/core/leoAtFile.py
edreamleo: My intended standard is `# #<issue-number>`.

leo/core/leoAtFile.py
edreamleo: Rev b741add converts two old bug references to the `# #<number>` convetion.

boltex: You changed those at lines 3075 and 3119, good catch I didnt see those. But those I had marked were at lines 2608 and 2614 (Fix bug 889175: Remember the full fileName.)

edreamleo: Let's not worry about the "old-style" comments :-) Either way is clear enough.

boltex (approved):
I've reviewed it carefully and translated it into LeoJS.

edreamleo:
@boltex Many thanks for your review and your update of LeoJS.

I'll merge this PR after merging PR #4281.

edreamleo:
@boltex Many thanks for your review and your update of LeoJS.

I'll merge this PR after merging PR #4281.

edreamleo:
@boltex FYI: recent revs have added several new unit tests.

boltex:
@edreamleo Thanks for the heads up, I will integrate the new changes in LeoJS during the weekend :)

boltex:
(minor nitpicking hehe..)  TestCommands.test_c_tabWidth headline should be named TestCommands.test_c_getTabWidth

boltex (commented):
@edreamleo I've updated LeoJS with the new unit tests. Thanks again for those! :)

boltex (approved):

boltex:
@edreamleo (in scripting-miscellany.html)  In the documentation, there is a small paragraph that includes '_scanAllDirectives_' called **Working with directives and paths**
. It should be shortened to the 'path' part only.

edreamleo:
@boltex I've fixed the headline and revised the documentation. This latter improvement is important.

Many thanks again for your invaluable review of this code.

PR #4330, opened 2025-04-14
Update afterChangeBody messages for convertBlanks, convertTabs, and deleteComments commands.

### LINKED ISSUE #4329: Some commands do not set the proper undo/redo string when calling afterChangeBody.
Some commands, for example in commanderEditCommands 'convert-tabs' and 'convert-blanks' do not set the proper undo/redo string when calling afterChangeBody.

boltex:
Fixes #4329


edreamleo (approved):
Good catch!

PR #4319, opened 2025-03-21
Back out of PR #4314

edreamleo:
**Status**

Closed: PR #4333 replaces this PR.

This PR will remain a draft until @boltex completes his researches.

- [x] Back out of PR #4314.
- [x] All tests pass, including pylint, pyflakes, mypy, and Leo's beautifier.

**About PR #4314**

The original PR works because, in effect, the PR executes what *appears* to be top-level code inside a wrapper function.  So the following script is valid:
```python
return 42
```

**How important is PR #4314?**

Imo, PR #4314 is unnecessary:
- Scripts may use Python's import commands to gain access to most of Leo's code.
- The PR makes debugging more difficult.
- The PR unnecessarily complicates the code, making it more difficult to maintain.

**History**

Approving PR #4314 was a mistake on my part:
- I did not read the code carefully.
- I did not request any unit test illustrating the utility of the PR.

I'll learn from these mistakes.

PR #4333, opened 2025-04-16
Fix for PR #4314

boltex:
This PR aims at replacing completely @edreamleo 's PR #4319 (which reverted my improper PR #4314 ) By removing the problematic code, and its additional user setting in leoSettings completely. 

As @edreamleo justly noted, that problematic PR contained an error-prone 'hack' which tried to fake the javascript behavior of allowing the 'return' keyword at the top level of an evaluated piece of code (with the 'exec' python function). 

Lessons learned: I should never do non-pythonic stuff like that in Leo codebase... (I was just too excited of figuring out a hack to return values with the 'return' keyword at the top level _and I removed the 'draft' flag of this PR way too early,_ having it merged into devel before I could realize my mistakes!) This was beyond silly and I hope to clean this all up myself with this PR.

This PR also instead adresses properly my original intent of simply fixing 'doCommandByName', which did not return the value of the command called. (e.g. some commands like 'convert-blanks' and many others return something) This was done by simply adding a 'return' on the last line of _commander_command_wrapper_ in the CommanderCommand class.

A single thing remain from my original idea: The (safe and non-hacky) pythonic way to grab something programmatically from code that ran with the 'exec' function which is by capturing a 'result' named variable on the global scope after 'exec' call. 

So if this PR is accepted, the problematic code and its associated setting will be removed completely, the only two things that will be in devel would be:

1. The fix to doCommandByName to capture the returned value of existing **official commands** from Leo's codebase.
2. Having **user commands** (script in @command nodes, @button nodes, ect..) to be able to return something by setting a global named 'result'.

I've also added unit tests for those two concepts.

I will soon add a paragraph about this in the documentation, and will consult the google-group forum with a demo about those two concepts before moving on...

...and will keep this PR as draft until it's ready ! 😅

### DIFF
--- a/leo/config/leoSettings.leo
+++ b/leo/config/leoSettings.leo
@@ -520,7 +520,6 @@
 <v t="ekr.20210506073820.1"><vh>@bool allow-text-zoom = True</vh></v>
 <v t="ekr.20241012050620.1"><vh>@bool diff-leo-files = True</vh></v>
 <v t="ekr.20131112150804.18737"><vh>@bool force-execute-entire-body = False</vh></v>
-<v t="felix.20250313171245.1"><vh>@bool allow-script-return-result = False</vh></v>
 <v t="ekr.20111115083813.12518"><vh>@bool indent-added-comments = True</vh></v>
 <v t="ekr.20150227102835.1"><vh>@bool make-node-conflicts-node = True</vh></v>
 <v t="ekr.20220624062948.1"><vh>@bool redirect-execute-script-output-to-log_pane = False</vh></v>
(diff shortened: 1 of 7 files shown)


edreamleo:
@boltex I too am responsible for the defects of the previous PR. Not to worry, it looks like all will soon be well.

boltex:
@edreamleo I've cleaned up a bit and I've also added to the documentation and removed obsolete stuff related to the return values of commands ran 'by name'.  

(I've only modified the Leo files, not generated the website files which can be done later in the devel branch before releasing the next version.)

Issue #4335, opened 2025-04-20
Python 3.9 requirement is not enough when importing 'Self' from 'typing'!


boltex:
When testing under different os and machines, I could not start Leo on my ubuntu machine which has python 3.9 (the required minimal version for running Leo) since a few weeks back, when more typing annotations were introduced, **specifically the ones that import 'Self' from the 'typing' library**. 

This is because 'Self' was introduced in the 'typing' library in Python 3.11. See https://docs.python.org/3/library/typing.html#typing.Self 

So either those specific annotations using 'Self' should be rolled-back / removed, or the minimal Python version for Leo should now be stepped up to 3.11. 

@edreamleo  Not sure what the best way to go about this would be!

edreamleo:
@boltex I'm on it.

edreamleo:
Fixed via PR #4337.

PR #4337, opened 2025-04-20
PR for #4335: Leo does not start with Python 3.9.

edreamleo:
See #4335.

- [x] Conditionally import 'Self'.
   All Tests pass with 3.9 and 3.13.
   There are minor mypy complaints with Python 3.9. These complaints should be ignored.

### DIFF
--- a/leo/commands/killBufferCommands.py
+++ b/leo/commands/killBufferCommands.py
@@ -5,11 +5,16 @@
 #@+node:ekr.20150514050411.1: ** << killBufferCommands imports & annotations >>
 from __future__ import annotations
 from collections.abc import Callable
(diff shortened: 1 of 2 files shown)


edreamleo:
@boltex Thanks for finding this issue. The fixes are straightforward. This PR is ready for review.

Imo, Leo should continue to support Python 3.9.

However, devs may safely require Python 3.13 when running all tests.

boltex (approved):
Looks good! :)

boltex:
@edreamleo Looks like this PR is still marked as draft btw.

edreamleo:
@boltex I marked it as a draft just so you would have a chance to review it.

Issue #4325, opened 2025-04-03
expand/contract-body/outline-pane commands do not work for legacy layout


edreamleo:
The `expand/contract-log` pane commands *do* work for the legacy layout.

tbpassin:
The method involved here is `LCW.resize_pane()`.

The problem is that the secondary splitter, which contains the body pane, has a horizontal orientation.  So any expansion or contraction would be in the horizontal direction.  And if one makes VR or VR3 visible (it will appear to the right of the body pane), the commands will work as expected - the body width will change and the VRx pane will change in the opposite direction.

But when a VRx pane is not visible, the body cannot change width because it already fills its splitter horizontally. If the body pane were expected to expand/contract vertically, then:

1.  The main splitter and secondary splitters would have to change sizes.
2. When the VRx pane is visible, the command would change the body pane size in the horizontal direction but when the VRx pane is not visible, the body pane would change size in the vertical direction.  This would be confusing.

One way to get the desired behavior of the body pane is to change the design of the legacy layout so that the outline and body share one splitter. This will require defining at least one if not two other splitters but ought to be doable.

However, this approach would make a command to change just the body width unfeasible - one would need to change the width of the splitter that contains both the outline and body panes, and the `resize_pane()` method won't work for this.

I can look into re-designing the legacy layout if there is some interest but we will need to agree on the intended behavior.

edreamleo:
@tbpassin Thanks for your detailed comments.

I'll examine the layouts and code before commenting further, but it's time to finish the big PR first.

edreamleo:
Completed via PR #4338.

PR #4338, opened 2025-04-22
PR for #4325: expand-body

edreamleo:
See #4325.

- [x] Replace `g.scanAllAtPathDirectives` by `c.getPath` in several places.
- [x] Improve `LCW.resize_pane`: It resizes the nearest splitter's parent if it can't resize the nearest splitter.
- [x] Add two new commands: `expand-main-splitter` and `contract-main-splitter`.
   These new commands work regardless of what splitter contains the widget mentioned in Leo's `expand/contract` commands.
- [x] Improve the documentation of all the commands affected by this PR.
- [x] Add `expand-main-splitter` and `contract-main-splitter` to list of "bad" leoserver commands.

### DIFF
--- a/leo/core/leoserver.py
+++ b/leo/core/leoserver.py
@@ -3739,6 +3739,7 @@ class LeoServer:
 
             'contract-body-pane',
             'contract-log-pane',
+            'contract-main-splitter',
             'contract-outline-pane',
 
             'edit-pane-csv',
@@ -3746,6 +3747,7 @@ class LeoServer:
             'equal-sized-panes',
             'expand-log-pane',
             'expand-body-pane',
+            'expand-main-splitter',
             'expand-outline-pane',
 
             'free-layout-context-menu',
--- a/leo/plugins/qt_frame.py
+++ b/leo/plugins/qt_frame.py
@@ -2006,6 +2006,16 @@ class LeoQtFrame(leoFrame.LeoFrame):
     def equalSizedPanes(self, event: LeoKeyEvent = None) -> None:
         """Make the outline and body panes have the same size."""
         self.resizePanesToRatio(0.5, self.compute_secondary_ratio())
+    #@+node:ekr.20250422154709.1: *5* LeoQtFrame.contract/expandMainSplitter
+    @frame_cmd('contract-main-splitter')
+    def contractMainSplitter(self, event: LeoKeyEvent = None) -> None:
+        """Decrease the size of the main splitter."""
+        self.divideLeoSplitter1(self.compute_ratio() - 0.1)
+
+    @frame_cmd('expand-main-splitter')
+    def expandMainSplitter(self, event: LeoKeyEvent = None) -> None:
+        """Increase the size of the main splitter."""
+        self.divideLeoSplitter1(self.compute_ratio() + 0.1)
     #@+node:ekr.20110605121601.18305: *5* LeoQtFrame.hideLogWindow
     def hideLogWindow(self, event: LeoKeyEvent = None) -> None:
         """Hide the log pane."""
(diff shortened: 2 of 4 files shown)


tbpassin (commented):
I'm completely confused.  These changes have nothing to do with the `expand-body` command. They belong in some other PR.

edreamleo:
@tbpassin You're not confused. Please allow me to make this little shortcut.

tbpassin:
Well, OK, but this PR doesn't yet address the issue. We haven't even decided on what approach we want to take.

edreamleo:
@tbpassin I think I have found an intuitive approach:

If the VRx pane is visible, `expand/contract-body-pane` change the relative sizes of the body and VRx panes.
Otherwise, these commands change the relative sizes of the body and outline panes (in the main splitter).

So if the VRx pane *is* visible, there is *no way* to change the size of the main splitter.
But that's ok. Hiding the VRx pane will allow the user to change the main splitter.

I've only tested the new code with the legacy layout, but `resize_pane` knows nothing about layouts, so all should be well.

**Summary**: I believe the new code is good enough.

@tbpassin What do you think?

tbpassin:
I think that I'm not sure what the command is expected to do. The expand command in some of the other layouts expands the body pane horizontally, and that works whether or not a VRx pane is visible.

For the legacy layout and no VRx, the body could expand vertically.  That would seen to make some sense.  But if a VRx pane is visible, should the command expand the body vertically, and if so should the VRx pane also expand vertically?  Personally since the body shares the field with VRx, I might expect the body to expand horizontally.  But this is opposite to the non-VRx direction. That would be unexpected.

So maybe with or without VRx showing, expand-body should expand upwards. This would make sense because the main reason for expanding the body would be to be able to see more lines of text.

Another thing to consider is that some other non-VRx pane might share the space with the body pane.  One example would be RPCalc. The command should contiune to work as expected.

Anyway, I think we have to have more clarity about the intended behavior. My preference would be for the splitter containing the body pane to expand/contract vertically for the legacy layout, and any other panes that share the same splitter would go along with the ride.  That is, they would continue to have the same height as the body pane.

It would be easy enough to write a little command to do so.  If we like it, we can combine it into the expand/contract commands.  Or if we can't get it to work right, we can change the design of the legacy layout so we can get the intended behavior.

edreamleo:
@tbpassin Thanks for your comments. I think they are reasonable.

Two (easy!) new commands should address your concerns: `expand-main-splitter` and `contract-main-splitter`.

Rev 406596c adds these commands. What do you think?

tbpassin:
The new commands should be helpful to those who know what they mean and where the main splitter is.  But those will be a small minority of Leo users.

edreamleo:
@tbpassin A note about special cases.

Imo, the new commands are required. It would be intolerable to add complex logic to `LCW.resize_pane`. Then we would have a real documentation mess on our hands.

Notice: `LCW.resize_pane` contains no significant `if` statements. All the `if` statements are guards.

However, there *is* a new path through the code. `resize_pane` resizes the parent splitter (the main splitter) if it can't resize the splitter containing the `widget` arg. This behaviour should be reasonable and intuitive in all layouts.

In short, the code is good as it stands. I'll improve the documentation before merging.

edreamleo:
@tbpassin This PR is ready for review:

- The revised `expand/contract-main-splitter` commands work with most, but not all, layouts.
- The docstrings for each layout now indicate the location of the main splitter.

boltex (approved):
looks good to me

boltex:
I've experimented with the commands, they seem to work in both the horizontal and vertical window layout! Great! 😄

edreamleo:
@boltex Thanks for your testing!

Issue #4339, opened 2025-04-23
@path can not recognize string with `-`


wangzhaohe:
See PR #4340.

Leo Log Window
Leo 6.8.4-devel, devel branch, build f372ff7022
2025-04-22 05:45:35 -0500
Python 3.12.0, PyQt version 6.5.3
darwin

@path can not recognize string with `-`, for example `learn-c`. It only get `learn`.

```
@path learn-c
    @file test.txt

The error is:
Directory not found:
/Users/swot/Downloads/learn
```

Then add `-` to the  at_path_pattern in @file leoCommands.py will recognize `-`.

```
at_path_pattern = re.compile(r'^@path\s+([\w_:/\\]+)', re.MULTILINE)
    to
at_path_pattern = re.compile(r'^@path\s+([\w_:/\\-]+)', re.MULTILINE)
```

edreamleo:
@wangzhaohe Thanks for the report. I agree with your solution.

edreamleo:
@wangzhaohe See [this wikipedia page about filenames](https://en.wikipedia.org/wiki/Filename). 
Perhaps the regex should allow all printable characters??

wangzhaohe:
I agree with your thinking. It is a path not only filename, so some problematic characters should be included.

edreamleo:
@wangzhaohe I am going to reopen this issue until the Leo devs decide what to do. See [this post](https://groups.google.com/g/leo-editor/c/fMS2tDNdL2Y/m/Z16SeF8gAgAJ).

edreamleo:
Completed via PR #4340.

PR #4340, opened 2025-04-24
PR for #4339: @path can not recognize '-'

edreamleo:
See #4339.

- [x] Allow *all* characters in path names except trailing blanks.
   Note: Both the legacy and changed regexes make it impossible to specify path names with leading blanks.
- [x] `g.stripPathCruft` always strips blanks.

**Q & A**

**Thomas**: What I don't understand is how to know if a given space is a trailing space.  It used to be, didn't it, that a headline could have text after the path of an `@path` directive.  I never used it like that but I have been under the impression that it did.  

**EKR**: Imo, "truncating" an @path directive at a space (or any other character not matched by the legacy regex) was an unintentional Easter Egg. The [directives reference](https://leo-editor.github.io/leo-editor/directives.html#path-path) makes no mention of this truncation. Rather, the documentation clearly implies that the intended path is everything following `@path`.

**Thomas**: If so, text following a space that was intended to have been a trailing space would get included into the path.  The obvious cure is not to allow any text after the path.  Is this going to be a new restriction, or has it always been like that?

**EKR**: This question is the crux of the PR's dilemma. The PR does what Leo's legacy code intended, but yes, you could say that the PR creates a new restriction.

**Discussion**

It is impossible and unwise to define set of characters that are deemed to be valid in directory names. The new policy places all responsibility in the hands of the user, where it belongs. Specifically, the risk of allowing control characters or non-printing characters is minimal. Attempting to prohibit such characters adds nothing to Leo.

Hand tests show that `@path c:\Users\Edward Ream` works as intended.

**Summary**

This PR is unlikely to cause problems. If it does, we'll fix those problems in another PR.

This PR must be documented in the 6.8.4 Release notes and What's New notes.

### DIFF
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -1742,7 +1742,7 @@ class Commands:
     #@+node:ekr.20250404014820.1: *6* c.getPathFromNode
     # Use a regex to avoid allocating temp strings.
     # https://en.wikipedia.org/wiki/Filename
-    at_path_pattern = re.compile(r'^@path\s+([\w_:/\\-]+)', re.MULTILINE)
+    at_path_pattern = re.compile(r'^@path\s+(.*+)', re.MULTILINE)
 
     def getPathFromNode(self, p: Position) -> Optional[str]:
         """
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -1741,7 +1741,8 @@ class Commands:
         return path
     #@+node:ekr.20250404014820.1: *6* c.getPathFromNode
     # Use a regex to avoid allocating temp strings.
-    at_path_pattern = re.compile(r'^@path\s+([\w_:/\\]+)', re.MULTILINE)
+    # https://en.wikipedia.org/wiki/Filename
+    at_path_pattern = re.compile(r'^@path\s+([\w_:/\\-]+)', re.MULTILINE)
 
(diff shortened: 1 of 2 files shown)


edreamleo:
@tbpassin @boltex This PR is ready for review. I'd like to merge this PR soon so that everyone may test it.

tbpassin (commented):
For the `at_path_pattern`:

1. I don't see what the "+" provides in `(.*+)'.  Seems to me that it can be omitted.
2. The regex could remove any trailing spaces using this expression: `r'^@path\s+(.*?)(?=\s*$)'`
3. If the string contains only `@path ` without any following text, the match will return None, which has to be handled downstream as a special case. The regex can  be changed to return an empty path instead: `r'^@path\s*(.*?)(?=\s*$)'`

tbpassin:
> 1. Thanks. Imo, group one should be `(.+)` rather than `(.*)` because an empty path isn't useful.
>        Also, the regex should end in `$` for clarity and safety.

I completely agree with both and I meant to write` (.+)` instead of `(.*)`.

For #2 and 3, I just offered them for interest.  In practice, not that it would matter in this usage, they would be slower because they use a look-ahead to detect the trailing spaces.

tbpassin (approved):

edreamleo:
@tbpassin Thanks for your careful reading, helpful comments and approval.

boltex (commented):
I'll apply this to LeoJS.

PR #4345, opened 2025-05-02
PR: Remove new pyflakes and ruff complaints

edreamleo:
These new complaints are the result of upgrading pyflakes and ruff.

- [x] Comment out 'global' statements for which pyflakes complains. Imo, these complains are unfortunate.
- [x] Fix mal-formed `#noqa` comments.
- [x] Fix several valid flake8 complaints that are symptoms of bugs in Leo's beautifier.
- [x] Fix several flake8 other valid complaints.
- [x] `test_gotoCommands.py` and `test_leoAst.py`:
   Suppress flake8 complaints that conflict with pyflakes checks!
- [x] Suppress checks of `leoflexx.py` in `leo/scripts/pyflakes_leo.py`.
  Pyflakes doesn't know about the `undefined` and `windows` globals, so there is no way to suppress related complaints.
- [x] `setup.cfg`: Suppress all flake8 tests in `leo/unittests/py3_test_grammar.py`.
- [x] Improve the docstrings for all of Leo's test scripts.
- [x] Use `setup.cfg` as explicit config file for `flake8`.
- [x] `full_test_leo` always runs `flake8_leo.py`.
- [x] Add `flake8` to two requirements files.
- [ ] Retest with Python 3.9 and 3.13.

**Removing "unnecessary" `global` statements**

Neither pyflakes nor pylint nor flake8 seems to warn about assignments to a *local* var that shadows a *global* var.

Unless I am totally mistaken, this means that removing *supposedly* unnecessary `global` statements can create a time bomb.
Suppose (much later) I mistakenly add an assignment to what *should be* a global var? Trouble!
Now the formerly-correct code will reference the local var, not the global!

### DIFF
--- a/leo/core/leoAtFile.py
+++ b/leo/core/leoAtFile.py
@@ -671,7 +671,7 @@ class AtFile:
         """
         p must be an @jupytext node.
         - Convert the .ipynb file to a string s.
-        - Update p.b's tree using s. 
+        - Update p.b's tree using s.
         
         This code is adapted from at.readOneAtCleanNode.
         """
--- a/leo/core/leoColorizer.py
+++ b/leo/core/leoColorizer.py
@@ -1341,7 +1341,7 @@ class JEditColorizer(BaseColorizer):
         QSyntaxHighligher calls this method repeatedly and automatically.
         
         Don't even *think* about changing this method unless you
-        understand *every word* of the Theory of Operation:  
+        understand *every word* of the Theory of Operation:
         https://github.com/leo-editor/leo-editor/issues/4158
         """
         trace = (False or 'coloring' in g.app.debug) and not g.unitTesting
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -1621,35 +1621,35 @@ redirectStdOutObj = RedirectClass()
 # Redirect streams to the current log window.
 
 def redirectStderr() -> None:
-    global redirectStdErrObj
+    # global redirectStdErrObj
     redirectStdErrObj.redirect(stdout=False)
 
 def redirectStdout() -> None:
-    global redirectStdOutObj
+    # global redirectStdOutObj
     redirectStdOutObj.redirect()
 #@+node:ekr.20041012090942.1: *5* restoreStderr & restoreStdout
 # Restore standard streams.
 
 def restoreStderr() -> None:
-    global redirectStdErrObj
+    # global redirectStdErrObj
     redirectStdErrObj.undirect(stdout=False)
 
 def restoreStdout() -> None:
-    global redirectStdOutObj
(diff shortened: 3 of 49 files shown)


edreamleo:

I'm not thrilled about commenting-out `global` statements, but I seen no alternative.

I'll delay Leo 6.8.4b1 so you may comment.

boltex:
oh well, hehe... 
I was today's years old when I learned that the global keyword in Python is necessary within a function **only** when the intent is to modify a global variable.

boltex (approved):
lgtm

tbpassin:
How do we test with ruff? When  I ran ruff_leo.py, it returned instantly with no errors, which makes me suspect it's not testing anything.  Running full_test_leo.py also produced no errors. I've got ruff 0.4.2.

```
Leo 6.8.4-devel, devel branch, build ef6a0f3806
2025-05-02 09:01:24 -0500
Python 3.12.8
Windows 11 AMD64 (build 10.0.26100) SP0
```

tbpassin:
> oh well, hehe... I was today's years old when I learned that the global keyword in Python is necessary within a function **only** when the intent is to modify a global variable.

I must have known that at one time but forgot.  Or rather I acted like I knew it but it became implicit knowledge so I didn't realize the distinction between referring and modifying a global variable  consciously anymore.

edreamleo:

Yes, the `global` command is necessary only when a function *modifies* a global variable.

Alas, neither pylint nor pyflakes is smart enough to warn when a (say newly-inserted statement) modifies a local var that *shadows* a global.

Imo, the pyflakes warnings about "unnecessary" `global` statements are close to being just plain wrong.

I see no way around this situation except to entire suppress pyflakes check entirely, and I don't want to do that.

tbpassin:
> Alas, neither pylint nor pyflakes is smart enough to warn about a (say newly-inserted) assignment modifies a local var that _shadows_ a global. Imo, the pyflakes warnings about "unnecessary" `global` statements are close to being just plain wrong.
> 
> I see no way around this situation except to entire suppress pyflakes check entirely for various files, and I don't want to do that.

Maybe we should take it as a hint not to shadow (or shadow then modify) a global without a really good reason.

edreamleo:
@tbpassin ruff is very fast. Use `pip list` to see what's installed.

tbpassin (approved):

edreamleo:
@boltex @tbpassin I am going to close this PR in favor of two new PRs with the following branches:
- `ekr-3-2025-fix-invalid-new-complaints`: Convert `global` to `# global`.
- ekr-3-2025-fix-valid-new-complaints: Fix all valid flake8 and pyflakes and update scripts & requirements.

Keeping the new branches and PRs will clarify the issues for us all, including pyflakes devs.

Issue #838, opened 2018-03-25
Better .ipynb importer/exporter


edreamleo:
The output of the present importer is unpleasant to use.

- [x] Importer should put all json attributes except "source" uA's.
- [x] Writer should write json attributes from uA's.

edreamleo:
The new uA-based code is much simpler and produces much better output.  This project is a complete success.

Issue #837, opened 2018-03-24
LeoU: Running unit tests


edreamleo:
​All devs must run all unit tests before commit anything. Here's how:

- Open leo/test/unitTest.leo in a console.
- The "Active Unit Tests" node should already be selected. If not, select it.
- Do Alt-4. It is bound in the file to run-selected-unit-tests-locally.

**Important**: Some unit tests may fail on your environment.  Don't worry about them.  You need only ensure that your new code causes no *other* unit tests to fail.  If in doubt, ask on [leo-editor](https://groups.google.com/forum/#!forum/leo-editor).

PR #4347, opened 2025-05-04
PR: Fix valid new complaints and update related files

edreamleo:
This PR replaces PR #4345 (closed).

- [x] Update requirements files.
- [x] Update all testing scripts.
- [x] `setup.cfg`: Suppress flake8 warning F824: `global whatever is unused: name is never assigned in scope`.
- [x] Fix all valid flake8 complaints.

PR #4348, opened 2025-05-04
PR: Suppress invalid pyflakes complaints

edreamleo:
A refactoring of PR #4345 containing only those changes required to suppress what I consider to be *invalid* pyflakes complaints.

My hope is that the pyflakes dev will roll back [pyflakes PR #825](https://github.com/PyCQA/pyflakes/pull/825), thereby making this PR unnecessary.

Yikes! A pyflakes dev rejected and locked [pyflakes issue #837](https://github.com/PyCQA/pyflakes/issues/837) within minutes of its being opened.

There seem to be only two choices left to me:
- Close this PR. This means that I Leo's test scripts won't ever run pyflakes in batch mode.
- Merge this PR and accept the consequences.

I don't like either choice, but I am leaning towards merging this PR.

There was a third choice, namely continuing the conversation via [pyflakes issue #838](https://github.com/PyCQA/pyflakes/issues/838). However, the pyflakes devs summarily closed this new issue, accusing me of wasting their time. I'm disappointed and displeased.

**I am going to merge this PR**

I am not happy about these enforced changes, but I believe merging this PR is best:

- This PRs changes are undoubtedly safe *for now*.
- It is unlikely that missing `global` or `nonlocal` statements will cause problems in the future.
  If they do, we'll deal with the problems as always.
- Heroic plans such as monkey-patching pyflakes or filtering pyflakes messages would be unwise.
  Yes, such schemes might work *now*, but they inevitably will create maintenance problems in the future.

### DIFF
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -1621,35 +1621,35 @@ redirectStdOutObj = RedirectClass()
 # Redirect streams to the current log window.
 
 def redirectStderr() -> None:
-    global redirectStdErrObj
+    # global redirectStdErrObj
     redirectStdErrObj.redirect(stdout=False)
 
 def redirectStdout() -> None:
-    global redirectStdOutObj
+    # global redirectStdOutObj
     redirectStdOutObj.redirect()
 #@+node:ekr.20041012090942.1: *5* restoreStderr & restoreStdout
 # Restore standard streams.
 
 def restoreStderr() -> None:
-    global redirectStdErrObj
+    # global redirectStdErrObj
     redirectStdErrObj.undirect(stdout=False)
 
 def restoreStdout() -> None:
-    global redirectStdOutObj
+    # global redirectStdOutObj
     redirectStdOutObj.undirect()
 #@+node:ekr.20041012090942.2: *5* stdErrIsRedirected & stdOutIsRedirected
 def stdErrIsRedirected() -> bool:
-    global redirectStdErrObj
+    # global redirectStdErrObj
     return redirectStdErrObj.isRedirected()
 
 def stdOutIsRedirected() -> bool:
-    global redirectStdOutObj
+    # global redirectStdOutObj
     return redirectStdOutObj.isRedirected()
 #@+node:ekr.20041012090942.3: *5* rawPrint
 # Send output to original stdout.
 
 def rawPrint(s: str) -> None:
-    global redirectStdOutObj
+    # global redirectStdOutObj
     redirectStdOutObj.rawPrint(s)
 #@-others
 #@-<< define convenience methods for redirecting streams >>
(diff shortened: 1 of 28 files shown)


edreamleo:
@boltex @tbpassin Do you have any suggestions?

boltex:
@edreamleo  If I read correctly their new rule about 'global',  their new rule triggers a warning, not an error.

 I'm not familiar with python linting and / or type-checking but could you setup Leo's usage of pyflake to use a configuration that only warns, and not block Leo's "compilation" or unit tests for that rule?

Again I'm a total newbie with python linting and type-checking, and know almost nothing about pyflake, but in my own web-development projects, I setup the linter's and type-checker to only warn on some rules, and to completely 'block' compilation by giving 'errors' on some other specific rules.

(Other than merging this PR and removing those unnecessary 'globals' )

edreamleo:
@boltex Many thanks for you comments and questions!

>  ...their new rule triggers a warning, not an error.

The answer probably depends on what "they" means:

flake8 issues a F824 warning/error. For flake8 the distinction isn't very important because it's easy to suppress the messages whatever the category is. Leo's PR #4347 suppresses the error in `setup.cfg`.

But afaik, the situation is way different for pyflakes. To my knowledge, there is no way to suppress *any* pyflakes message except to change the offending code. In other words, pyflakes must be "very very careful" not to issue unwanted message.

tbpassin:
What is the relation between flake8 and pyflakes?

I suppose that the Leo code that handles the pyflakes output could just filter out the unwanted message lines. Not a preferred solution but it could work.

edreamleo:
@tbpassin Thanks for your suggestion. I've considered it, but it's not as simple as one might think.

The problem is that pyflakes writes a summary line. Leo would have to remove the summary line only if we want to remove *all* the errors for a particular file.  For example:
```console
:118:5: `global times_d` is unused: name is never assigned in scope
:120:5: `global trace` is unused: name is never assigned in scope
:121:5: `global verbose` is unused: name is never assigned in scope
3 errors in c:\Repos\leo-editor\leo\core\leoclient.py
```
I'll also consider monkey-patching the disputed pyflakes code.

edreamleo:
@tbpassin Thanks for your suggestion. I've considered it, but it's not as simple as one might think.

The problem is that pyflakes writes a summary line. Leo would have to remove the summary line only if we want to remove *all* the errors for a particular file.  For example:
```console
:118:5: `global times_d` is unused: name is never assigned in scope
:120:5: `global trace` is unused: name is never assigned in scope
:121:5: `global verbose` is unused: name is never assigned in scope
3 errors in c:\Repos\leo-editor\leo\core\leoclient.py
```
I'll also consider monkey-patching the disputed pyflakes code.

tbpassin:
If Leo's code removes lines it can also count those lines and substitute its own summary.

That's not a wonderful solution but monkey-patching pyflakes isn't ideal either. Any change to pyflakes might mess up the monkey-patch. We could try restricting the pyflakes version, but sooner or later there would be trouble. 

OTOH, we could restrict the pyflakes version to the last one that works as desired, and avoid monkey-patching. But some users are going to update their system without realizing not to update pyflakes.

edreamleo:
@tbpassin Thanks for your comment. I agree that neither monkey patching nor filtering is without problems, especially future problems.

Issue #4350, opened 2025-05-07
Allow Leo's beautifier to beautify scripts


edreamleo:
See PR #4351.

- The `execute-script` command beautifies `c.p` and its subtree if `@bool beautify-python-code-on-write = True`.
- Create new `beautify-tree` command.

PR #4351, opened 2025-05-07
PR: beautify scripts

edreamleo:
See #4350.

- [x] `leoTokens.py`: call `main` function only if `__name__ == '__main__'`.
- [x] Use `python -m leo.core.leoTokens` in beautifier scripts.  Required.
- [x] Add `tbo.beautify_script_tree` and its helper, `tbo.beautify_script_node`.
- [x] Add `beautify-tree` command.

**Aha: `tbo.beautify_script_node` is sound**

A hack: `tbo.beautify_script_node` translates section refs to `pass` statements.
The hacks exists because section references might have names that Python's tokenizer won't accept.

This hack is sound because section references must appear on a line of their own:
```python
if <<a test >>:   syntax error!
   whatever
if (  # Valid
    << a test >>
):
    whatever
```
As a result, the inserted `pass` statement should yield valid tokens. But  Leo's beautifier uses only tokens, so all should be well.

**Notes**

- Automatic beautification of scripts is undoable. So is the `beautify-tree` command.
- Unit tests of this code would be tricky. I'll add a unit test only if bugs are reported later.

### DIFF
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -1249,9 +1249,7 @@ class Commands:
         if beautify_flag and not script and not g.unitTesting:
             from leo.core.leoTokens import TokenBasedOrange
             tbo = TokenBasedOrange()
-            for p2 in script_p.self_and_subtree():
-                tbo.beautify_script_node(p2)
-            c.redraw(script_p)
+            tbo.beautify_script_tree(script_p)
 
         # Compute the script if necessary.
         if not script:
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -628,6 +628,14 @@ class Commands:
                     repr(self.fixedWindowPosition))
         else:
             c.windowPosition = 500, 700, 50, 50  # width,height,left,top.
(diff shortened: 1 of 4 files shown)


edreamleo:

As noted in the first comment, there is no reasonable alternative to the slightly hacky approach taken in `tbo.beautify_script_node`.

edreamleo:
@boltex I have added the `LeoJS` label, but you are free to ignore these changes.

edreamleo:
@tbpassin  @boltex I just realized that `tbo.beautify_script_node` is sound!

BTW, Python will all other Leo other directives as decorators because they all start in the first column.

tbpassin:
> I just realized that `tbo.beautify_script_node` is sound!

What does this mean?

tbpassin (commented):
By "script" I assume that we mean a subtree with Python code that is not an external at-file.

When I run the "beautify" command on such a subtree, I expect to end up with a beautified version of the original code no matter what my setting of _run-pyflakes-on-write_ is.

I do not expect that pyflakes would automatically be run on the subtree. I do expect that there could be a separate pyflakes command.

Are these expectations correct for this PR?  If not, what will happen instead?

Issue #4355, opened 2025-06-01
'goto-global-line' does not work with @auto nodes


boltex:
See PR #4359. This PR prints a warning (once) if the root of `c.p' isn't an `@file` or `@auto` node.

'goto-global-line' works nicely with `@clean` and `@files`.

But 'goto-global-line' does not seem to work with `@auto` nodes. I've tested with plain text files, and markdown, and in both cases the 'import-as-any' command seems to work nicely (even splits markdown files into proper sections) but the goto-global-line fails with messages about 'line not found' or 'only 5 lines' as per this screenshot (with the content of the markdown file selected in leo in the notepad above leo):

![Image](https://github.com/user-attachments/assets/3c4ba0d1-63cf-4129-a424-cfc10d18652c)

boltex:
@edreamleo 

In order to help finding out why this bug happens, I looked in gotoCommands.py, and ran Leo in a debugger to look at what happens specifically in find_file_line_helper for `@auto` nodes (_versus `@files`, `@clean`, etc. which work fine with the 'goto-global-line' command_)

I found that for `@auto` nodes, the `get_external_file_with_sentinels` returns only the opening and closing sentinels, missing the whole fine in between those sentinels. 

So I looked at the last time the problematic code was changed, and its 22 months ago ( in commit d973646c9e9d76c648e9537c523de0bf18e544eb ) where this was removed: 

![Image](https://github.com/user-attachments/assets/b6629122-94b6-4b45-8ec4-2cf6206cd87f)

also, I wonder why the 'finally' block did not bring that flag to false...?

Hope this helps pin-pointing this bug. I'll give it another go tomorrow  :)

edreamleo:
@boltex Thanks for your analysis. I'll fix this soon.

edreamleo:
@boltex I'll add a warning for `@auto` and `@asis` nodes as you suggest.

PR #4359, opened 2025-06-05
PR for #4355: Print warning unless root is an @file or @clean node

edreamleo:
See #4355.

An edge case: The code warns on the deprecated `@thin` spelling. Probably LeoJS doesn't support `@thin`, so this edge case is especially minor.


tbpassin (approved):
This seems fair enough.  I have never really understood what `@auto` is for or its foibles, so after importing a file I normally change it to `@file` or `@clean`.

boltex (dismissed):
~That's perfectly fine :)~
Edit: I found some flaws

PR #4360, opened 2025-06-06
PR 2 for goto-global-line

edreamleo:
See #4355 and PR #4359.

Last-minute testing of PR #4359 revealed a special case that should be valid for all kinds of `@<file>` nodes.

- [x] Add another special case to goto.find_file_line_helper.

### DIFF
--- a/leo/commands/gotoCommands.py
+++ b/leo/commands/gotoCommands.py
@@ -55,6 +55,9 @@ class GoToCommands:
             return self.find_script_line(n, p)
         # Step 0: goto-global-line works only for @file and @clean nodes.
         if not any((root.isAtCleanNode(), root.isAtCleanNode())):
+            # Support the special case used by g.findGnx.
+            if p and n == 0:
+                return p, 0
             # #4355: Print a warning once.
             g.es_print_unique_message('goto-global-line works only for @file and @clean')
             return None, -1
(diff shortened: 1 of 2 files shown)


tbpassin (commented):
Line 62: ` g.es_print_unique_message('goto-global-line works only for @file and @clean')`

This will not print the message if a second external file also tries to trigger it. Is this really the behavior we want? Or do we want the message to be emitted once for every qualifying at-file?

boltex:
@edreamleo @tbpassin I'm just tagging you here because I forgot to tag you in my message above, (and my message from yesterday) which I think is why it was not seen :smile:

tbpassin (approved):

boltex (changes requested):
(oh, i found how to request changes in the github interface 😄 )

In order to support the other types of file-nodes that are supported with this command, I think the condition at step 0 should be :

```python
if not (
    root.isAtCleanNode() or
    root.isAtFileNode() or
    root.isAtEditNode() or
    (root.isAtAsisFileNode() and len(root.v.children) == 0)
 ):
```

edreamleo:
@boltex Rev 2bc88bf5 makes the change you requested and tweaks the unit tests so it passes again.

edreamleo:
@boltex @tbpassin Oops. Rev 664a196 computes the condition correctly. There is no need to change the unit tests!

boltex (approved):
I also tested with leointeg using it via leoserver. Works great. :)

PR #4361, opened 2025-06-07
Boltex/issue4357

boltex:
Added `show_line_in_leo_outline` and `insert_file_node` commands to leoserver. 

Also added a `_capitalize_drive` utility used in `insert_file_node` because file names set by vscode on windows may have a lowercase drive letter. (unlike in python)


edreamleo (approved):
This PR looks good to me, but I don't understand the server issues. I'll trust that you know what you are doing.

Issue #4356, opened 2025-06-01
importing files should make the path relative if in same folder (or deeper) than the Leo file itself.


boltex:
_(Note: maybe there's a reason this is not already the case, but since I dont know of any, I filed this issue just in case )_ 😄

When importing a file (`import-any-file`) that is in the same folder as the Leo file itself (e.g. `C:\some-folder\myLeoFile.leo` ) , the resulting `@auto` headline should not be an absolute path from root like so:
`@auto C:\some-folder\myImportedFile.txt`

it should be just `@auto myImportedFile.txt`

This could be done by checking if the absolute imported file path starts by the same absolute path of the Leo document itself, and if so, just truncate it from the imported file path.

 And if **not** starting with it, **only then** use the absolute path.

 (For now, let's not have the third option that imported files would try to use 'folder up' , e.g.: `../blabla` or `../../blabla` )

tbpassin:
I would add that if the imported file is in a subdirectory of the outline, let the path also be relative.

edreamleo:
@boltex @tbpassin Thanks for your comments.

This is a funny blind spot. I typically remove the unnecessary path prefixes *by hand* when importing files. I'll fix this issue soon.

boltex:
@edreamleo 

In my example above, that `commonPath `variable should be named `specificPath`, because its the rest of the imported file path **after the common part** of the path with the commanderFilename  was removed.

tbpassin:
It's probably as simple as `PurePath.relative_to(other, walk_up=False)`, perhaps together with `PurePath.is_absolute()`.

edreamleo:
@tbpassin That was my idea too. With a unit test :-)

I'm not sure what to do if the `.leo` file isn't in any parent directory. Maybe an absolute path is best in that case.

boltex:
@edreamleo I would second the idea that absolute paths be used for anything not equal or deeper than the Leo file's directory. 

**After doing some experiments in LeoJS in the last few days with this concept,** I suggest that the process of truncating a file path to the common part of the .leo file should be centralized in a leoGlobal function, and used at a few specific and crucial places in the code.  

_(e.g. its important not to pass such a truncated path to 'looksLikeDerivedFile')_

I mention this small example as a reason to state that **it should be truncated later**, just about when the headline is set for the newly inserted node, **rather than sooner** at the user's choice of file in the dialog.

boltex:
[...]

PR #4366, opened 2025-06-16
PR for #4356: relative imports

edreamleo:
See #4356.

- [x] Add `g.relativeDirectory` and corresponding unit test.
- [x] Add `c.relativeDirectory`.
- [x] Call `c.relativeDirectory` from the import logic.
- [x] Test `c.relativeDirectory` by hand. Creating an actual unit test would be tricky.

### DIFF
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -2729,6 +2729,12 @@ class Commands:
         c = self
         path = c.getPath(p)
         return g.finalize_join(path, p.anyAtFileNodeName())
+    #@+node:ekr.20250616161500.1: *4* c.relativeDirectory
+    def relativeDirectory(self, path: str) -> str:
+        """Return the path relative this outline, or the full, absolute path."""
+        c = self
+        baseDir = os.path.dirname(c.fileName())
+        return g.relativeDirectory(baseDir, path)
     #@+node:ekr.20171123135625.39: *4* c.getTime
     def getTime(self, body: bool = True) -> str:
         c = self
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -2731,7 +2731,7 @@ class Commands:
         return g.finalize_join(path, p.anyAtFileNodeName())
     #@+node:ekr.20250616161500.1: *4* c.relativeDirectory
     def relativeDirectory(self, path: str) -> str:
-        """Return the path relative this outline, or the full, absolute path."""
+        """Return the path relative to this outline, or the full, absolute path."""
         c = self
         baseDir = os.path.dirname(c.fileName())
         return g.relativeDirectory(baseDir, path)
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -3310,7 +3310,12 @@ def readlineForceUnixNewline(f: IO, fileName: Optional[str] = None) -> str:
     return s
 #@+node:ekr.20250615134309.1: *3* g.relativeDirectory
 def relativeDirectory(baseDir: str, path: str) -> str:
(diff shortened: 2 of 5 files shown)


edreamleo:

Note that `c.relativeDirectory` does not exist in `leoJS`, but I think it's worth having.

tbpassin (commented):
leoGlobals.py
----------------
Lines 3321ff -
```
   except ValueError:
        pass  # Windows throws ValueError if the drives are different.
```
I think an error message should be issued, because if there is a drive mismatch, it's probably a mistake and the user might want to correct it.

Otherwise the PR looks all right to me.

boltex (commented):
I'd suggest that the specific case of an unsaved/untitled Leo document might be better to use the old 'absolute path' instead of a relative path.

leo/core/leoCommands.py
```diff
@@ -2729,6 +2729,12 @@ def fullPath(self, p: Position) -> str:
         c = self
         path = c.getPath(p)
         return g.finalize_join(path, p.anyAtFileNodeName())
+    #@+node:ekr.20250616161500.1: *4* c.relativeDirectory
+    def relativeDirectory(self, path: str) -> str:
+        """Return the path relative to this outline, or the full, absolute path."""
+        c = self
+        baseDir = os.path.dirname(c.fileName())
```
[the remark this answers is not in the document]
edreamleo: Rev f9dc5ca does as you suggest.

leo/core/leoGlobals.py
```diff
@@ -3308,6 +3308,26 @@ def readlineForceUnixNewline(f: IO, fileName: Optional[str] = None) -> str:
     if len(s) >= 2 and s[-2] == "\r" and s[-1] == "\n":
         s = s[0:-2] + "\n"
     return s
+#@+node:ekr.20250615134309.1: *3* g.relativeDirectory
+def relativeDirectory(baseDir: str, path: str) -> str:
+    """
+    Return the path relative to the base directory.
+    Return path if baseDir is not an ancestor of path."""
+    if not baseDir:
+        # Likely an unsaved outline.
```
boltex, line 3317: Just one last small thing, For users to be made aware of this detail on the spot. I think I would have printed this when _not baseDir:_

`g.es_print_unique_message('Imported using absolute path (unsaved outline). Save the outline to enable relative paths.')
`

edreamleo: Thanks for this. Please see b423981

boltex (changes requested):
I systematically tried various file types when using the import commands, I found two small things to fix. (I had the same two issues in LeoJS by the way! 😁 ) 

## import_txt_file

I had missed this function of commanderFileCommands.py : **import_txt_file**, which needs the line changed when setting the headline:
 `p.h = f"@edit {c.relativeDirectory(fn)}"`
 (it's needed to leave fn as is for later in the method)
 
## createOutlineFromWeb

 Another strangeness (_although not related to the 'relativeDirectory' feature_) that I found while testing various import-any-file formats, is the **createOutlineFromWeb** method of leoImports.py (used when importing cw, cweb, nw or noWeb files) which work fine, but hide the imported node inside whatever node was currently selected!

So instead of `parent.insertAsLastChild` it should be `parent.insertAfter`, just like the other importers. That way, the imported node would be apparent once imported. 

Many thanks for making this new feature of Leo a reality!

edreamleo:
@boltex Rev 62366a1 fixes the `import_txt_file` function.

edreamleo:
@boltex Rev a4e553f fixes ic.createOutlineFromWeb.

Afaik, this rev should complete this issue. Do you agree?

edreamleo:
@boltex Thanks for suggesting this change!

Issue #4363, opened 2025-06-12
Import-any-file should offer *.* as file types choices only, or list all valid importer types, like .md, etc.


boltex:
See PR #4369.

Leo has an importer that does a very good job with markdown ".md" files. This should perhaps be reflected in the choices presented to the user below "all files *.*"

![Image](https://github.com/user-attachments/assets/2553bc1e-f515-4b0e-946f-c9201db019a9)

tbpassin:
I don't know why there has to be a selection list at all. The menu Item is "Import Any File", and the file extension will determine the file to be imported.  The only reason I see is that the dialog will filter the file names presented to the user by their extension. 

With the menu name, I expect Leo to figure out what importer to use.  With a selection list, I expect Leo to use the importer I specify. To me, the current menu item should be named "Import By File Type", not "Import Any File".

boltex:
@tbpassin I think you've got a good idea in simply offering **\*.\*** since the command is literally called  `import any file` and it will use the correct importer regardless.

boltex:
@edreamleo @tbpassin I've edited the title to reflect the discussion.

tbpassin:
Thanks for being so thoughtful and thorough!  

There is a value in letting the dialog filter by file type.  It's just that the list of known types is so long it's unwieldy.  Also, if we only support `*.*`, no update has to be done each time a new language gets added. That's an advantage.

Does the file dialog in Linux filter the same way?  I don't recall.

boltex:
@edreamleo Thanks for this 😄 

Additionally, as I mentionned in the title of this issue,  for the 'import-as-any' dialog list of **"available types to filter by"**,  I think we should add the missing importers suported file extensions to the list, so that the user can truly choose to filter with the one he likes, if he wants to.  (a few are missing, like .md for  markdown as example.) **As long as the first, default choice is `*.*`**

tbpassin:
I thought you meant that `the never-used defaultextension` meant that the arg isn't used when the method is called.  But looking at the code for `runOpenFileDialog()` I see it's not even used **within the implementation**.  Now that's **really** unused.

edreamleo:
@boltex Please tell me the complete list of missing extensions. I'll be happy to add them.

boltex:
@edreamleo  no problem, will do :)

boltex:
@edreamleo The complete list to fill filetypes in importAnyFile would then become:
```python
filetypes = [
    ("All files", "*"),
[...]

PR #4369, opened 2025-06-19
PR for #4363: tweak args in file dialogs

edreamleo:
See #4363.

- [x] Remove all `defaultextension` args in calls.
- [x] Add comments in `x.runOpenFileDialog`, `x.runOpenFilesDialog`, and `x.runSaveFileDialog` that
    the `defaultextension` kwarg is unused.
- [x] Add more file types to the `import-any-file` command.

There is no need to deprecate the `defaultextension` kwarg, but it seems harmless to remove it from LeoJS.


edreamleo:
@boltex I have added the file types as you suggest. I've also consolidated groups of types.

boltex (approved):
Thanks for those changes!

PR #4374, opened 2025-06-22
PR: Fix failing unit test on Linux

edreamleo:
`os.path.relpath` and `g.relativeDirectory` give different results on Window and Linux. The unit test must take these differences into account. 

As a result, Leo's import commands will also give different results, but those differences won't (usually) cause problems when moving an outline from one platform to another.

- [x] Use different test cases on Windows/Linux.
- [x] Verify that all tests pass on Windows/Linux.


tbpassin (commented):
with _ekr_ in some of the paths, only ekr will be able to run this unit test.

Issue #4377, opened 2025-06-24
Binding legends may be inaccurate for @command and @button nodes


edreamleo:
See PR #4378.

The **legends** (The first letters in each line of the `show-settings` command) for `@button` or `@command` nodes in `myLeoSettings.leo` may be incorrect.

For example, EKR's `myLeoSettings.leo` contains:
```console
@settings
  @buttons
    @button button-test @key=F`
       @language python
       g.es_print('@button button-test!!!')
```
The `show-bindings` command shows:
```console
F  button: F3  button-test  # Oops: The 'F' legend should be 'M'.
```

edreamleo:
Closed via PR #4378.

PR #4378, opened 2025-06-24
PR for #4377: binding legends

edreamleo:
See #4377.

- [x] Add optional `fileName` kwarg to `k.registerCommand` and `k.registerCommandShortcut`.
  Also, require kwargs in both methods. These are minor **breaking changes**.
- [x] `sc.registerAllCommandsCallback` calls `k.registerCommand` with proper `fileName` kwarg.
- [x] Add `__slots__` to `g.BindingInfo` class.

**Discussion**

`k.registerCommandShortcut` uses the `fileName` data to more accurately assign legends.

### DIFF
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -1922,7 +1922,7 @@ class LoadManager:
             g.es(f"{kind:>10}:", os.path.normpath(theDir), color='blue')
     #@+node:ekr.20120215062153.10740: *3* LM.Settings
     #@+node:ekr.20120130101219.10182: *4* LM.computeBindingLetter
-    def computeBindingLetter(self, c: Cmdr, path: str, scope: str) -> str:
+    def computeBindingLetter(self, c: Cmdr, path: str) -> str:
         lm = self
         if not path:
             return 'D'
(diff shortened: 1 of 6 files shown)


edreamleo:
@boltex @tbpassin @jknGH This PR is ready for review. The fix was surprisingly simple and effective.

edreamleo:

- You can discover what `aso` means by selecting the node then doing `Alt-N`, `goto-next-clone`. I often use this pattern.
  aso is short for `ActiveSettingsOutline`, which imo was a failed experiment.
- The distinction between `stroke` and `key` is too difficult to describe, change or remember.
- Imo, the likelihood of anyone using the methods whose signatures have changed is virtually nil.
  However, I'll add them to the list of breaking changes.

tbpassin:
> * The distinction between `stroke` and `key` is too difficult to describe, change or remember.

But is the usage different between the two code snippets I gave that are close together in the code? If the usage isn't different, let's use the same name for both.

edreamleo:
@tbpassin The only significant change is at line 3164:

`k.bindKey(pane, stroke, func, commandName, tag=f"register-command:{fileName}")`

Using `fileName` rather than `c.shortFileName` makes things work. That's the content of this PR.

tbpassin (approved):

Issue #180, opened 2015-04-06
goto-prev-history-node bugged with bookmarks


JohnLunzer:
If you select a node using the mouse or keyboard and then navigate to a new node using a bookmark  and then use ```goto-prev-history-node``` it will successfully return to the previous node. However, if you navigate to another node using a bookmark without using the keyboard or mouse and try to use ```goto-prev-history-node``` it will navigate to the previous *bookmark* not the node that was previously selected in the outline tree.

It appears as though bookmarks.py is the culprit as it does nothing to update the beadList or beadList pointer in the NodeHistory instance.

I'm using Leo 5.1-b1 on windows with python 2.7 and pyqt 4.8.6.

tbnorth:
Haven't been able to reproduce this - can you give a click by click procedure? Bookmarks call g.handleUrl which calls recursiveUNLSearch c.selectPosition c.frame.tree.select ...
c.NodeHistory.update

Point being that the higher level commands bookmarks.py uses to select a new node should be handling the update to beads.

JohnLunzer:
Create four nodes (name them what you want) and one bookmarks node. Create your bookmarks pane. Make bookmarks for the third and fourth nodes.

Navigate to the first node, then to the second. Use ```goto-prev-history-node``` and the first node should now be selected (as expected). Now navigate back to the the second node again using the mouse or keyboard. Now click the bookmark for the third node, the third node should now be selected. Now use ```goto-prev-history-node``` and the second node should be selected (as expected). 

Here is where things break down. Now click the bookmark for the fourth node, the fourth node should be selected (as expected). Now use ```goto-prev-history-node```, rather than the second node being selected as you would expect (because it was in fact the previously selected node) the third node is now selected, which was the previously visited bookmark.

edreamleo:
What's the status of this issue?

tbnorth:
I suspect this is not really a bookmark specific problem.  Rather, the point in code execution where visited positions are recorded for ``goto-prev-history-node`` is to "high", so that a lot of things which change the selected node are missed by it.

What is the definition of visiting a node? It could be giving focus to the body editor, but that would be confusing, so it seems it should be having the tree drawn with a node selected counts as visiting the node.

No time right now :-/, but that's my guess.

edreamleo:
[...]

PR #4380, opened 2025-06-25
PR for LeoJS issue #180

edreamleo:
See [LeoJS issue #180](https://github.com/boltex/leojs/issues/180).

- [x] Write a more concise message, and write it only once.


edreamleo:
@boltex What do you think?

boltex (commented):
I've updated LeoJS with this same change.

Issue #4373, opened 2025-06-21
Crash in u.undoGroup


edreamleo:
See PR #4381.  This issue probably affects LeoJS.

```console
Traceback (most recent call last):
  File "C:\Repos\leo-editor\leo\core\leoGlobals.py", line 320, in new_cmd_wrapper
    func(self, event=event)
  File "C:\Repos\leo-editor\leo\core\leoUndo.py", line 1745, in undo
    u.undoHelper()
  File "C:\Repos\leo-editor\leo\core\leoUndo.py", line 1903, in undoGroup
    newP = u.newP.copy()  # Must exist now, but may not exist later.
AttributeError: 'NoneType' object has no attribute 'copy'
```

edreamleo:
Closed via PR #4381.

PR #4381, opened 2025-06-25
PR for #4373: Crash in u.undoGroup

edreamleo:
See #4373. This PR probably affects LeoJS.

- [x] Add defaults if args do not exist. The old code did not respect the comments!
- [x] Add a unit test.


edreamleo:
@boltex Ready for review. The unit test seems strong enough, but I don't have enough notes to reproduce the original crash.

PR #4389, opened 2025-07-09
Fix leoserver.py for support of websockets >= 14.

### LINKED ISSUE #4388: Fix leoserver.py for support of websockets >= 14.
Note: See pull request https://github.com/leo-editor/leo-editor/pull/4389 

Starting at version 14 from november 2024, websockets has changed its default api.

Fix leoserver to detect the version equal or above 14 and import "websockets.legacy" as "websockets" instead where the older compatible module now resides.

boltex:
Fixes #4388

### DIFF
--- a/leo/core/leoserver.py
+++ b/leo/core/leoserver.py
@@ -5779,6 +5779,7 @@ def main() -> None:  # pragma: no cover (tested in client)
                     answer = controller._do_message(d)
                 except TerminateServer as e:
                     await websocket.close(code=1000, reason=str(e))
+                    return
                 except ServerError as e:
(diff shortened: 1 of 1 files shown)


boltex:
@edreamleo Yes, if the version is 14 or above, then we want to use what is called the 'legacy' module. 

(because we are using a newer version, not the opposite, as a 'legacy' module did not exist in the older versions: they added a legacy module for those who want to use the older API in their new versions 14 and above.)

edreamleo (approved):

boltex:
@edreamleo Thank you! :)

edreamleo:
@boltex You're welcome. Would you like Leo 6.8.6 to go out the door in a few weeks?

Issue #4385, opened 2025-07-07
Improve @clean for large repos


edreamleo:
PR #4386 completes both phases.

The sources for the [rust compiler](https://github.com/rust-lang/rust/tree/master/compiler) contains almost 4000 files.

My present study outline for the rust compiler represents these files as `@@auto` nodes. This outline takes about 8 seconds to load. Importing these files takes **much** longer:

- Import to `@clean`: `imported 38912 nodes in 1980 files in 459.96 seconds`.
- Import to `@auto`: `imported 38937 nodes in 1980 files in 387.91 seconds`.

**Phase 1**: Don't change `@clean` trees if they are up-to-date with their corresponding external files:

This is a *spectacular* optimization.

**Phase 2**:  Optimize changed nodes within `@clean` trees:

Phase 2 rescans updated nodes with Leo's importers to handle split nodes.

New `x.postpass` helper functions clean up unpleasant artifacts of the update process.

edreamleo:
Closed via PR #4386.

PR #4395, opened 2025-07-15
PR for #4394: Fix @clean bugs

### LINKED ISSUE #4394: Fix recently- reported problems with @clean
PR #4385 made huge changes to code related to `@clean`.
PR #4395 contains the changes inspired by this issue.

- Thomas reported problems with `@clean` update [here](https://groups.google.com/g/leo-editor/c/BkXXsCWAgoA/m/ZDGfRQgMDgAJ).
- `c.refreshFromDisk` should call `at.readFileAtPosition` instead of duplicating its code!

edreamleo:
This PR continues PR #4385.

*Note*: PR #4397 will add support for `@leo` nodes and the  `open-at-leo-node` command.

**Visible improvements**

- [x] Support `@bool report-changed-at-clean-nodes` with default False.

**Changes and bug fixes**

- [x] Tweak `at.clone_all_changed_vnodes`:
    Fix several bugs. Callers are responsible for redraws and changing `c.p`.
- [x] Tweak `at.readOneAtCleanNode`:
    - *Always* clear the `mod_time` uA. Only `at.writeOneAtCleanNode` should set this uA!
    - Call `c.setChanged(force=True)` to ensure that the "changed" indicator in the tab is visible.
- [x] Move a strange error message, seldom (never?) given, to `at.readFileAtPosition` where it undoubtedly belongs.
- [x] Alphabetize cases in `at.readFileAtPosition`.
- [x] Improve and simplify `c.refreshFromDisk`:
    - Add `p` kwarg to `c.refreshFromDisk`. The new signature is compatible with the old.
      As a result, there is no need to call `c.selectPosition` before calling  `c.refreshFromDisk`.
   - Call `at.readFileAtPosition` instead of duplicating code.
   - Clear the `_mod_time` uA if any node has changed.
   - Call `c.selectPosition(p) appropriately.
   - Call`c.setChanged()`.
- [x] `efc.idle_check_commander` delegates redraw to `c.refreshFromDisk`.
- [x] `ServerExternalFilesController.idle_check_at_file_node` delegates redraw to `c.refreshFromDisk`.
- [x] `efc.toAtAuto` and `efc.toAtEdit` just call `c.refreshFromDisk`. They no longer save/restore the insert point.
    The old code has been moved to the attic.
- [x] `c.redraw`: default to `c.rootPosition()` and issue a warning if `not c.positionExists(p)`.
- [x] Add the `force` kwarg to `c.setChanged`.
- [x] `v.setHeadString` always clear the `_mod_time` uA.

**Cleanups**

- [x] `leoShadow.py`: Remove unused `node_pat` pattern ivar.
- [x] Simplify `v.setBodyString` and `v.setHeadString` by moving common code.

**Testing**

- [x] Add a new unit test that verifies that all command functions have an `event` kwarg.
- [x] Test (by hand) all forms of update with both values of  `@bool report-changed-at-clean-nodes`.
- [x] Pass a non-empty root to `at.replaceFile`, thereby allowing that method to be simplified.


edreamleo:
@boltex There's no setting yet. I don't foresee the setting being useful, but of course it would be easy to add.

boltex:
@edreamleo I've made a post on the user's group instead of here explaining my reasons for making the insertion of nodes an opt-in, so that more leonistas can participate : https://groups.google.com/g/leo-editor/c/BkXXsCWAgoA/m/P-PGfPIcAAAJ

edreamleo:
@boltex Rev fbe0c20 adds support for `@bool report-changed-at-clean-nodes` with default False.

edreamleo:
@boltex @tbpassin This PR is ready for review. I know of no problems that remain.

boltex:
Thank you @edreamleo  !!! :) 

I review it live right now as I port it to LeoJS, as I've found it's a great way to spot bugs & incompatible types! (found a few in the past this way)

edreamleo:
@boltex You're welcome. I won't merge this PR until you approve.

boltex:
@edreamleo no problem, I'll finish the testing during the weekend.

boltex:
@edreamleo 

Although this new system of detection for refreshing the '@'clean nodes who are considered 'changed' is an incredible improvement speed wise, and an invaluable change for anyone working with a large codebase that uses many '@'clean nodes, the fact remains that it is based on file-timestamps and a **_mod_time** user attribute applied on each '@'clean node.

While this is of incredible effectiveness in any scenario involving outlines with many '@'clean nodes, it it still a **fundamental change in Leo's theory of operation** for '@'clean nodes, because it is based on timestamps, instead of actual file-content changes.

Being conscious about not wanting to add yet-again new settings, I'm inclined to reflect on the obvious question: "Should this 'Hyper-speed-mode' be also an opt-in feature based on a setting?"  

### Scenarios about problematic situations with file timestamps-based comparisons

- A user moves the Leo file in another folder containing a 'mirror' of his files, either to compare, or he used that 'copy' folder for backup after messing things up (or for some other reason)

- in a collaborative context, and the leo file itself (only the leo file, not the external files) was sent by another person and expects the other's files in to be kind-of the same project folder setup/configuration. 

- again in a collaborative context where the user expects the other's files in to be kind-of the same project folder setup/configuration: should the _mod_time be kept in the db like for marked nodes? so that a user sending his leo file to a collaborator has no pre-set _mod_time U.A. on any node?

- A user has an outline that uses `@path` in his tree, which are parents of his `@clean` node(s), if he changes the `@path` nodes to point somewhere else for some reasons, will those timestamps and _mod_time ua still be making sense?

- a user copy/pastes an @clean node to rapidly create a new one pointing on another file by altering its headline, or, he simply changes the headline without making a copy so it points on another filename: Would it keep its _mod_time U.A?

- a user has altered the mod time in some way via the O.S, or has performed some operation that messed up the file-timestamps or the path pointed to by the `@clean` node, and so, the user uses the 'refreshFromDisk' command. Would that command first clear the _mod_time U.A. to make sure the @clean node is refreshed based on content changes and not based on timestamps when using that specific command. (currently it does not)

- Other situations where the _mod_time U.A. should be cleared: Any ancestor that contains an '@' in its headline (@path, @ignore, etc...) is altered or moved, Or if the node itself is moved, pasted, renamed, ect...

**Conclusion**: Having the '@'clean refresh only based on the timestamps and their node's inner _mod_time U.As could be problematic, as perhaps good-old file-contents comparison would be necessary in those instances.

Not sure what to suggest here, I hesitate between some (or all) of those ideas:

- Make this 'lightning-fast-change-detection-mode' an opt-in feature via a setting
- Having much more detection of triggers that would clear that '_mod-time' UA so that @clean nodes are refreshed based on actual contents instead of their _mod_time UA + file timestamps (e.g. moving / renaming nodes, using the 'refreshFromFile' command, etc.)
- Have dialogs pop-up when opening leo files and/or warn the user in the log pane about a new setting for this new  'lightning-fast-change-detection-mode' 
- Since this is such a big improvement speed-wise for `@clean` nodes: Have this new mode be switchable in the menus, instead of having to go manually into myLeoSettings.
- Instead of making another setting: Make a new directive: `@fast-detection`, or `@timebased`, (whatever) for which any descendants `@clean` (or other <@>file) nodes would be using that new change detection method.

Note: This last suggestion: (a new directive for a node for which its descendant nodes would use the new fast change-detection) is the one that is the most 'visible and discoverable' to the user without having to pop dialogs or messages in the log pane, nor does it add yet another setting to know about and setup in myLeoSettings. But i'm not sold on it personally I think i would still prefer the simple 'opt-in' setting.

Of course I dont want to force anything, or seem like i dont like that new detection method: i actually think its a fabulous idea and should have been available from the get-go decades ago 😄 

Thanks for reading all of my concerns about this great improvement. I'll continue testing it out throughout the week-end in both Leo and LeoJS to see if I cant find any other quirps and I'll post them here along with any other suggestions or ideas that come up to mind.

Félix

boltex:
@edreamleo

 I just retried automatically refreshing @auto nodes by modifying external files **on the master branch**, and upon refreshing I see that no outline nodes are selected, nor the body refreshed. (But, after manually selecting the node in the outline, I see that the changes to the body was correctly picked up.)  So I guess there was a gui '**redraw**' call missing **(or redraw_later is bugged)** after updating '@'clean or '@'auto nodes since a while ago before those changes to '@'clean nodes
 
 Example:  After externally changing the sentence 'Make sure the LeoJS extension is installed' to 'Make sure the XXXXX extension is installed' , I get this popup:
 
image


**Now after accepting, I see that no nodes are selected in the outline and that the body has not changed:**


image


**But then if I select the node conteining what was supossed to change in the body: I see that in fact it refreshes with the proper changes:**


image

edreamleo:
> I notice that the node is updated upon external file change, but that the currently **selected node is not set** to the '@'clean node once it finished updating.

@boltex The code in question, `at.clone_all_changed_vnodes`, may update many `@clean` nodes. The `@clean` update update algorithm is unchanged and it has not (and should not) touch `c.p.

If `@bool report-changed-at-clean-nodes` is `True`, Leo should select the last top-level node. Otherwise, Leo should retain `c.p` before the update.

edreamleo:
@boltex 

< In leoShadow.py, a member node_pat is [unused].

Good catch. Fix at rev 93b6128.

edreamleo:
> The docstring for **readOneAtCleanNode** seems truncated.

@boltex Fixed at rev 4d708c1.

edreamleo:
@boltex Many thanks for your lengthy discussion of `_mod_time` uAs. I don't have a quarrel with any of the scenarios you list, but I don't claim to understand all their implications.

I think opting out of the new scheme isn't a great solution. In a cooperative environment the old way is likely to be unacceptably slow.

Unless I am mistaken, the essence of all the potential problems you mention is that the `_mod_time` uA might become misleading or invalid. I don't have an opinion about how common that situation might be.

Executing Leo's `refresh-from-disk` command will correct all `@<file>` nodes in `c.p` and its subtree. In at least some situations, this "by hand" approach should suffice.

If `@bool report-changed-at-clean-nodes` is True, the user will get a list of changed *nodes*, but  `refresh-from-disk` *applied to the top-level organizer node* won't necessarily change all the changed *files*. Maybe `refresh-from-disk` could be enhanced to find all the `@clean` nodes corresponding to the changed nodes. That's not the case now.

The most general solution would be to call `refresh-from-disk` on *all* the nodes of the outline. Yes, this will take a long time, but thereafter the uAs will be correct.

Do you have any other ideas?

edreamleo:
@boltex I have tested all the possibilities by hand. Imo the present code is solid. Please let me know if you disagree!

boltex:
@edreamleo 

Thanks for this! 

I'll finish porting it to leojs and study it in the next 1-2 days! (I think you totally got the desired results in a way i didn't think of, hah!) - I'll finish reviewing it and approve it very soon.

boltex:
After testing a bit, I've noticed that when opening a file that had some at-clean nodes update because they were changed since last time that leo file was opened, that the leo file is considered dirty: **e.g. Leo will ask about saving the file if I press the X to close the window.** 

(even with the option **changed-at-clean-nodes** set to false as to not generate changed nodes report outline)
image


But the tab does not have the asterisk to signify that the Leo document state is dirty. (like that second tab in that screenshot below)
image

So, I guess a 'redraw' of the tab part of the UI should be done when finishing opening a Leo file.

---
It's late so I'll continue testing tomorrow :) 
Félix

edreamleo:
@boltex Glad you approve. I'll fix the missing * in the tab and also add a new `open-at-leo-file` command.

edreamleo:
@boltex Rev 2e49290 fixes problems with the "changed" marker.

It also corrects a crasher when `@bool report-changed-at-clean-nodes` is False.

In short, I think this PR is good to go.

edreamleo:

Rev 34939f0 fixes the problem with headlines. The new rule is that reading a changed file *always* clears the `mod_time` uA.
No fancy (global!) tests are required.

Rev 7857892 fixes the problem with the missing "changed" indicator when updating an `@clean` file on entry.
The fix required a little hack to `c.setChanged()`.

boltex:
@edreamleo I made more rigorous testing, and I noticed that I forgot an important point when removing that `mod_time `Ua from a node: 

is should be removed with **del** instead of setting to **None**, in a manner similar to this:
```python
if '_mod_time' in p.v.u:
    del p.v.u['_mod_time']
```

Since the 'u' member of Ua's is initialized with `{}` an empty dict, to cleanly remove a Ua the **del** keyword should be used, otherwise a Ua set to **None** remains.

Otherwise it leaves data in the ua in the .leo and .leojs files which only accumulates as nodes are copied/renamed.

Secondary to this, when I ran manual tests a minute ago, I ran into this weird behavior/bug: 

I started by adding those 3 print statements in readOneAtCleanNode (note the green marks on the left)
image

Then, in this simplest possible leo document:
image

I re-open this leo file after modifying that test1.txt external file, to trigger the readOneAtCleanNode logic, I get this in the terminal as expected:
image

But then, If I run the 'show-all-uas' command, I still see that UA: 
image

...And just for curiosity, I save the file and inspect the .lo file content, and the UA is still there:
image

So something is wrong regarding this, or I completely misunderstand something fundamental... hehe :) 

I hope this helps - we can also zoom about it so I can share my live-screen if you'd like - no problem at all 

Félix

edreamleo:
@boltex Thanks for your latest comments.

Rev 38129ca3 clears the `_mod_time` uA as you suggest. It also clears the uA unless the root node is an `@clean` node.

edreamleo:
@boltex Many thanks for your thorough testing!

As of rev b264de5, `at.replaceFile` *always* writes `@clean` files, even if their contents haven't changed!

My local tests show that the `clear-all-uas` command does, in fact, cause Leo to write the new  `_mod_time` uAs when Leo saves the outline, even if the only change to the outline were then changed uAs.

`at.replaceFile` is the reason why `_mod_time` uAs did not change after clearing all uAs and saving the outline. Previously `at.replaceFile` (quite naturally) wrote only *unchanged* files, so the following line in `at.writeOneAtCleanNode` didn't change the uA!
```python
root.v.u['_mod_time'] = g.os_path_getmtime(fileName)
```

boltex:
@edreamleo Thanks! This worked for the Ua's upon opening a file and reading the changed @clean nodes!

**...And again, sorry for my impetuous and insistent comments and change requests! :)**

You last modifications fixed the `mod_time` Ua that persisted when reading an `@clean` that had changes since its last `mod_time`. Also, the new technique of removing the Ua (via 'del' instead of setting to None) makes the 'ghosts' of those Uas disappear properly! 😄 

Now, **The only remaining issue** that should be easy to fix (I hope) is that changing a headline should clear the `mod_time` Ua for the reasons I explained 2 or 3 messages above (pasted here below): 

Here is the problem: a lingering `mod_time` ua on a newly renamed `@clean` could prevent its refreshing from disk when the user clears its body/outline content, and **tries to have it get its content filed up with the refresh-from-disk command**. 

That is why renaming a node should simply have the side-effect of removing `mod_time` Ua if present on that node. (renaming a node invalidates its mod_time because its pointing somewhere else, or simply not an @clean anymore, doesn't matter, either way it should be removed.) 

I hope it is more clear now why changing a headline should clear its `mod_time` Ua if present.

Many thanks again for your patience and indulgence with me! This is the greatest speed improvement of Leo with big projects involving` @clean` nodes, and it should thus be 'polished'.

- hoping this last request makes sense for you too and that you agree with it.
Félix

edreamleo:
@boltex 

> Here is the problem: a lingering `_mod_time` uA on a newly renamed `@clean` could prevent its refreshing from disk when the user clears its body/outline content, and tries to have it get its content filed up with the `refresh-from-disk` command.

I haven't paid attention to the details because I don't want to scatter special cases throughout the code.

If all you care about is making sure that `refresh-from-disk` works for `@clean` nodes, a *general* solution is to clear the uA in `refresh-from-disk` *before* calling `at.readFileAtPosition`. This disables the following early return in `at.readFileAtPosition`:
```python
 if old_mod_time and old_mod_time >= new_mod_time:
        return True
```
In the case of changed headlines, the later return *can't happen* because *private* lines do contain sentinels, and thus the `@+node` sentinels lines will be different:
```python
if new_private_lines == old_private_lines:
    return True
```

**Summary**

Clearing the `_mod_time` (in `c.refreshFromDisk`) *before* calling calling `at.readFileAtPosition` should be a general solution to all possible special cases.

What do you think?

edreamleo:
Oops: The first version of the comment above said that the returns are in `at.readFileAtPosition`.

I've correct the comment to say that both returns in `at.readOneAtCleanNode`. My apologies for this confusing typo.

edreamleo:
@boltex Thanks for your comments, but I don't understand why clearing the uA in `refresh-from-disk` doesn't suffice:

- Changing a headline (of *any* node) p will guarantee that Leo prompts the user to change p's `@clean` parent.
- Likewise, pasting a node will guarantee that Leo will write the pasted node's `@clean` parent.

Alright, maybe a few nodes might have invalid (useless) `_mod_time` uAs, but so what?

I can live with the new code in `setHeadString` but I don't understand it. Let's zoom if you think I'm missing something.

boltex:
@edreamleo Here is an even better **setHeadString**

```python
    def setHeadString(self, s: object) -> None:
        # pylint: disable=no-else-return
        # Fix bug: https://bugs.launchpad.net/leo-editor/+bug/1245535
        # API allows headlines to contain newlines.
        v = self
        old_head = v._headString
        if isinstance(s, str):
            v._headString = s.replace('\n', '')
            v.updateIcon()
        else:  # pragma: no cover
            s = g.toUnicode(s, reportErrors=True)
            v._headString = s.replace('\n', '')  # type:ignore
            self.contentModified()  # #1413.
            v.updateIcon()
        if v._headString != old_head and '_mod_time' in v.u:
            del v.u['_mod_time']  # Renaming a vnode clears the mod_time.
``` 

I'll send you an invite by email to zoom about it 😄

File: leo/core/leoAtFile.py [hunk 1]
```diff
@@ -2991,24 +3010,25 @@ def replaceFile(self,
...
-            if not g.unitTesting and c.config.getBool(
-                'report-unchanged-files', default=True):
-                g.es(f"{timestamp}unchanged: {sfn}")  # pragma: no cover
-            # Check unchanged files.
-            at.checkPythonCode(contents, fileName, root)
-            return False  # No change to original file.
-        #
+        if not root.isAtCleanNode():  # #4394: Always write @clean nodes!
+            # Compare the old and new contents.
+            old_contents = g.readFileIntoUnicodeString(fileName,
```
boltex, line 3015: oh oh: **old_contents**  has to be set BEFORE this` if not root.isAtCleanNode(): ` line above because it is also used after that 'if' below, at line 3035. (would be completely unset otherwise)

boltex (changes requested):
Only that 'old_contents' line that should be above the 'if'  remains for approval :)

boltex (approved):
Looks good! :)

edreamleo:
@boltex Thanks for another great collaboration.

Issue #4404, opened 2025-08-02
@clean can corrupt data!


boltex:
See PR #4405.

**EKR**: Leo 6.8.6 used importers to post-process `@clean` nodes using Leo's importers.

That idea appears fatally flawed. The PR removes all the buggy code and moves it to the attic.
I have no plans to resurrect this code.

Félix (the OP) and I discussed this issue at length on zoom. We agree that `at.do_changed_vnode` and related code are the culprits.

<details><summary><b>Original report</b></summary>
<br>

Using this node in my outline: `@clean index.html` I noticed the read algorithm is not the right one being used on external file change detection.

When I modify an external html file (not an unknown .xyz file type, but a file for which an auto importer exists) then the read algorithm completely messes up my outline because it does not use the usual `@clean` read algorithm .

My outline starts like this:

<img width="303" height="148" alt="Image" src="https://github.com/user-attachments/assets/4f55dbe8-ba47-49f0-b89c-c4fe482dbb9d" />

I then change a single word in my html page within an external editor, expecting Leo to update the `@clean` outline, while keeping the existing structure. (as it normally did before)

But this happens instead: 

<img width="528" height="350" alt="Image" src="https://github.com/user-attachments/assets/f295d789-c4fb-42ed-be43-aecfd2e19567" />

Neither the `@clean` nor `@auto` read-algorithms changed in the last update, so I guess it's just a 'if/then' condition that uses one instead of the other that is at fault here.

I hope it's not too hard nor long to fix as this is quite a serious mis-behavior of the external-file-change detection and update feature!

Félix

</details>

edreamleo:
@boltex Thanks for this report. The relevant calling sequence seems to be:

- `efc.idle_check_commander` calls `c.refreshFromDisk(p)`.
- `c.refreshFromDisk` calls `at.readFileAtPosition(p)` which then calls `at.readOneAtCleanNode(p)`.

I see no significant `if` statements in this chain.

But the *server* logic is different. I may have botched it.

- `sefc.idle_check_commander` calls `sefc.idle_check_at_file_node`.
- `sefc.idle_check_at_file_node` calls `c.refreshFromDisk(self.lastPNode)`
  This call looks fishy to me!

In short, the main logic looks sound. Are you reporting a problem with the server??

boltex:
It also does it in leo itself outside of leointeg.

PR #4405, opened 2025-08-03
PR for #4404: Fix data loss in @clean updates

edreamleo:
See #4404. This PR will form the entire content of Leo 6.8.6.1.

- [x] Remove all code that post-processes `@clean trees`. The code is now in the attic.

I could try to make the importers work, but that seems too risky. I have no plans to resurrect the code.

### DIFF
--- a/leo/core/leoAtFile.py
+++ b/leo/core/leoAtFile.py
@@ -756,79 +756,8 @@ class AtFile:
             c.setChanged(force=True)
             root.v.setDirty()
             at.changed_roots.append(root.copy())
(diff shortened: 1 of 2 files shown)


edreamleo:
@boltex Many thanks for all your help. I'll release Leo 6.8.6.1 tomorrow.

tbpassin (commented):
leoAtFile.py
---------------
Line 759 - `return True  # No errors`
I'm unable to see if the method might return False somewhere because I didn't see any other exit in the loop. If it can't return anything else, why do we need any return value?

PR #4407, opened 2025-08-04
fix TypeError when attempting to run show-fonts command

jknGH:
without this fix, the command 'show-fonts' (from Menu or minibuffer) gives an error when using Qt:

```
  File "/home/jkn/leo-editor/leo/core/leoKeys.py", line 2475, in callAltXFunction
    func(event)
TypeError: showFonts() missing 1 required positional argument: 'event'
```


jknGH:
(To fix issue #4408 - I probably added this after the PR, ie in the wrong order ;-o )

edreamleo (approved):
Thanks for this change!

PR #4409, opened 2025-08-05
PR: Investigate re-importing changed @clean nodes

edreamleo:
This PR tries to understand why `at.do_changed_vnode` failed so spectacularly for html files.

- [x] Restore dangerous code. Highly experimental

A few experiments with html files show that `at.do_changed_vnode` definitely *should not* ever be called for html files. The problem is that html importer will likely destroy the existing outline structure.

That statement *might* also be true for other file types.

In short, I think it's best never to "improve" changed nodes! So I am closing this PR!


edreamleo:
I am closing this PR. Re-importing changed nodes is a doomed idea.

PR #4410, opened 2025-08-05
Boltex/issue4401

boltex:
@edreamleo I Started from branch ekr-4401-python314 

This leoserver.py works with 3.14 and older versions.  ...Wasn't too hard 😄 

Still needs a bit of cleanup and testing/experimenting just to make sure its totally fine. I'll do that in the following days.


edreamleo:
@boltex Thanks for this work.

Please base your PR on devel. I think only `leoserver.py` needs to change.

boltex:
@edreamleo no problem

Issue #4411, opened 2025-08-05
Rewrite parse-body


edreamleo:
See PR #4412 and [this post](https://groups.google.com/g/leo-editor/c/HvtFuI3hk0I/m/W7gWAP8wBAAJ).

The PR's version of `parse-body` uses the importers as helpers. This is a huge simplification of the code.

boltex:
@edreamleo  What about nodes with the pattern: "**MyMethod and helpers**" common throughout Leo's codebase for instance. (a node which contains multiple methods/functions intentionally keps in a single node.)

If one of those nodes has a modification, the proposed new @clean algo will forcefully split them? 

(That is what I meant by nodes being "exploded into many sub nodes" where it was not desired by the user who crafted a specific outline structure.)

boltex:
@edreamleo _Addendum to the comment i just posted above_: This is why I thought this should be a command, (split-node or whatever) instead of an automatic behavior of a @clean (or other) read algorithm.

edreamleo:
@boltex Great question. I am about to write an ENB resolving all such questions.

The short answer: automatic update has no chance of being what we want, regardless of settings or directives. Instead, we want a suite of new or improved commands.

edreamleo:
Completed via PR #4412

PR #4412, opened 2025-08-05
PR for #4411: rewrite parse-body

edreamleo:
See #4411.

- [x] Add `g.app.importerClassesDict` and `g.app.importerModulesDict`.
- [x] Rewrite the `parse-body` command. It is now defined for all importers (languages) except `xml` and `html`.
- [x] Rewrite corresponding unit tests.
- [x] Test all supported importers by hand, especially undo/redo.

**Related changes**

- [x] `at.readOneAtCleanNode` returns `None`. This will be the *only* change to `leoAtFile.py`!
- [x] Add `mark-changed-nodes` command as an alias of the existing `mark-changed-items` command.
- [x] Delete `u.beforeParseBody` and `u.afterParseBody`. Replace them with existing undoer helpers.
- [x] Improve "Position mismatch" messages in `u.beforeUndoGroup` and `u.afterUndoGroup`.
- [x] Ensure that all importers define `x.block_patterns` as a tuple.

### DIFF
--- a/leo/commands/commanderOutlineCommands.py
+++ b/leo/commands/commanderOutlineCommands.py
@@ -1365,6 +1365,7 @@ def createMoveMarkedNode(c: Cmdr) -> Position:
     return p
 #@+node:ekr.20031218072017.2923: *3* c_oc.markChangedHeadlines
 @g.commander_command('mark-changed-items')
+@g.commander_command('mark-changed-nodes')
 def markChangedHeadlines(self: Cmdr, event: LeoKeyEvent = None) -> None:
     """Mark all nodes that have been changed."""
     c, current, u = self, self.p, self.undoer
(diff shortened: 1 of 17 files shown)


tbpassin (commented):
This PR doesn't seem ready for review yet. The Python  node splitter is still instrumented with printouts, and some methods don't do anything yet (e.g., `clean_headline()`.  The Rust node splitter doesn't seem to have any code to do anything yet.

edreamleo:
@tbpassin You are correct. This PR will soon undergo major changes.

In general, any time you see traces or breakpoints with `###` comments you know those statements are temporary.

edreamleo:

tbpassin (commented):
leoImport.py
----------------

- Lines 1123 - 1124:
```
       if not p:
           return
```
The user will not know of the failure nor the reason for it.  Can we provide some info here?
A similar comment applies at 1142-1143.

edreamleo:
@tbpassin Thanks for your comment. Rather than retaining this guard I think it's reasonable to delete it.

boltex (approved):
Looks good

edreamleo:
@boltex Thanks for your review :-)

PR #4418, opened 2025-08-07
PR fixes #4416: The _mod_time uA should not pollute descendentVnodeUnknownAttributes

### LINKED ISSUE #4416: Don't add _mod_time to v.descendentVnodeUnknownAttributes
See PR #4418.

Leo 6.8.6.1 pollutes outlines with  `_mod_time` and `descendentVnodeUnknownAttributes` xml attributes.

The following script shows that `_mod_time` is the source of the pollution:

```python
    if hasattr(p.v, "unknownAttributes"):
        g.printObj(p.v.unknownAttributes, tag=p.h)
```
Here is example output:
```console
@clean ../../MANIFEST.in: {
key: _mod_time:
1753876229.4462912
}
@clean ../../setup.cfg: {
key: _mod_time:
1753876229.4462912
}
@clean ../external/leoftsindex.py: {
key: _mod_time:
1753876229.4621859
}
@clean ../external/make_stub_files.cfg: {
key: _mod_time:
1753876229.4969122
}
@clean ../external/py2cs_theory.md: {
key: _mod_time:
1753876229.4969122
}
@clean ../plugins/leo_babel/__init__.py: {
key: _mod_time:
1753876229.509613
}
@clean ../plugins/leo_babel/tests/__init__.py: {
key: _mod_time:
1753876229.509613
}
```

edreamleo:
See #4416.

- [x] Add '_mod_time' to `FastRead.nativeVnodeAttributes`.
- [x] `fc.createUaList`: Remove `_mod_time` from uAs.
- [x] Oops: Make changes to `qt_commands.py` from another PR.
- [x] Verify that `FastRead.scanTnodes` gets the `_mod_time` attribute from `<t>` elements.
- [x] Ensure that Leo writes a `_mod_time` attribute when writing an *unchanged* `@clean` node.

**Related changes**

- [x] Improve reporting in `at.replaceFile`.

**Notes**

- For compatibility, `FastRead.resolveUa` must remain as it is.

### DIFF
--- a/leo/core/leoAtFile.py
+++ b/leo/core/leoAtFile.py
@@ -697,7 +697,7 @@ class AtFile:
         # #4385: Do nothing if the file has not changed.
         try:
             old_mod_time = root.v.u['_mod_time']  # #4385
-            # g.trace(f"{old_mod_time:20} {root.h}")
+            g.trace(f"{old_mod_time:20} {root.h}")
         except Exception:
             old_mod_time = None
         new_mod_time = g.os_path_getmtime(fileName)
@@ -1672,14 +1672,12 @@ class AtFile:
                 at.addToOrphanList(root)
             else:
                 contents = ''.join(at.outputList)
-
-                # #4385: at.replaceFile always writes @clean roots,
-                #        even if the file hasn't changed.
-                #        This forces the `_mod_time` uA to change.
                 at.replaceFile(contents, at.encoding, fileName, root)
 
                 # #4385: This is the *only* place that sets the `_mod_time` uA.
                 root.v.u['_mod_time'] = g.os_path_getmtime(fileName)
+                g.trace(f"{root.v.u.get('_mod_time')} {root.h}")
+
         except Exception:
             at.writeException(fileName, root)
     #@+node:ekr.20090225080846.5: *6* at.writeOneAtEditNode
@@ -2964,20 +2962,20 @@ class AtFile:
         if not old_contents:
             old_contents = ''
 
-        if not root.isAtCleanNode():  # #4394: Always write @clean nodes!
-            # Compare the old and new contents.
-            unchanged = (
-                contents == old_contents
-                or (not at.explicitLineEnding and at.compareIgnoringLineEndings(old_contents, contents))
-                or ignoreBlankLines and at.compareIgnoringBlankLines(old_contents, contents))
-            if unchanged:
-                at.unchangedFiles += 1
-                if not g.unitTesting and c.config.getBool(
-                    'report-unchanged-files', default=True):
-                    g.es(f"{timestamp}unchanged: {sfn}")  # pragma: no cover
-                # Check unchanged files.
-                at.checkPythonCode(contents, fileName, root)
-                return False  # No change to original file.
+        unchanged = (
+            contents == old_contents
+            or (not at.explicitLineEnding and at.compareIgnoringLineEndings(old_contents, contents))
+            or ignoreBlankLines and at.compareIgnoringBlankLines(old_contents, contents))
+
+        if unchanged:
+            at.unchangedFiles += 1
+            if not g.unitTesting and c.config.getBool(
+                'report-unchanged-files', default=True):
+                g.es(f"{timestamp}unchanged: {sfn}")  # pragma: no cover
+
+            # Check *unchanged* files.
+            at.checkPythonCode(contents, fileName, root)
+            return False  # No change to original file.
 
         # Warn if we are only adjusting the line endings.
         if at.explicitLineEnding:  # pragma: no cover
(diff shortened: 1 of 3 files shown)


edreamleo:

Examination of my local test file shows that the unwanted `descendentVnodeUnknownAttributes` attributes on `<v>` elements have been eliminated. The `todo.py` plugin (for example) still works as expected.

boltex:
@edreamleo This does not work.

The mod_time only appears on `@clean` first creation. What I mean by that is that if leo is closed, and then the external file modified, then, upon re-opening in leo, that `@clean` is indeed refreshed because the timestamp of the external file is newer, but then upon re-saving the Leo file, there is no mod_time UA at all in the .leo file at all. 

I was expecting a new mod_time that reflects the new external file timestamp.

edreamleo:
@boltex ~~Good catch!~~ Leo writes the `_mod_time` attribute on `<t>` elements, ~~but the read code doesn't yet know that~~.

edreamleo:
@boltex Tests show that `_mod_time` *is* handled correctly. 

Why? Because `FastRead.scanTnodes` handles the `_mod_time` attribute on `<t>` elements.  This would be hard to see because the PR *doesn't change* `scanTnodes`!

Here is the script that I used to verify that all works as expected:
```python
g.cls()
for v in c.all_unique_nodes():
    if v.u:
        g.printObj(v.u, tag=v.h)
```

I verified that `v.u.get('_mod_time')` changes when any `@clean` tree changes.

edreamleo:
@boltex A more convincing test is inserting a trace in `at.readOneAtCleanNode`:
```python
try:
    old_mod_time = root.v.u['_mod_time']  # #4385
    g.trace(f"{old_mod_time:20} {root.h}")
except Exception:
    old_mod_time = None
```
Now we can clearly see that changing an `@clean` node *effectively* updates `_mod_time` in the read logic.
For example: here are the traces from my test file:
```console
readOneAtCleanNode    1754819058.800121 @clean c:\test\leo\at_clean_test.py
readOneAtCleanNode   1754578507.0247476 @clean c:\test\leo\at_clean_test.rs
readOneAtCleanNode    1754530773.705714 @clean c:\test\leo\README.txt
readOneAtCleanNode   1754530773.7140796 @clean c:\test\test.html
```
Now I change `at_clean_test.py`, save the outline, and reload. The traces are:
```console
readOneAtCleanNode     1754819428.73223 @clean c:\test\leo\at_clean_test.py
readOneAtCleanNode   1754578507.0247476 @clean c:\test\leo\at_clean_test.rs
readOneAtCleanNode    1754530773.705714 @clean c:\test\leo\README.txt
readOneAtCleanNode   1754530773.7140796 @clean c:\test\test.html
```
Only the `_mod_time` for `at_clean_test.py` has changed.  I rest my case.

boltex:
@edreamleo I re-did my experiments carefully to make sure i did not have the wrong branch active, or some other rookie mistake. 

The bug still stands. 

One point I  should have stressed in order for you to reproduce the problem is that, at the very last part of my experiment, i did **not** change the file before saving it again. 

If I type something in the body of a node of the `@clean` and save the leo file, then yes, it writes the mod_time ua. So the last sentence of my comment above where I described the problem should have read:

> ... but then upon re-saving the Leo file _without changing anything_, there is no mod_time UA at all in the .leo file at all.

Emphasis on the added part in italic 😄 

I can share my screen to show you if you'd like.

So to summarize: 
- the mod_time is written when created, 
- the mod_time is used to compare with the file date/time to make sure not to refresh if not changed. 
- If actually changed, it refreshes the @clean with the changes upon opening the Leo file.
- But then if saved right there, without additional changes in Leo, the resulting .leo file has no mod_time ua. 

Hope this clears it up :)

tbpassin (approved):

edreamleo:
@boltex Thanks for the clarification. While working on another matter, I found a problem with `at.replaceFile` that is related to `@clean` nodes. I'll fix both problems (if they are separate) in this PR.

edreamleo:
@boltex Rev 0ed8a63 simplifies `at.replaceFile`, thereby fixing an unrelated problem with reporting unchanged files.

This rev also enables two traces that demonstrates that the `_mod_time` uA is read and written correctly. Examination of my test `.leo` file shows all is well.

In short, I think you are confused by the behind-the-scenes operation of uAs.

edreamleo:
@boltex One more thing. Rev 0ed8a63 removes a false and misleading comment in `at.writeOneAtCleanNode`.

The erroneous comment stated that `at.replaceFile` *always* wrote `@clean` files. That was *never* true! The simpler version of `at.replaceFile` makes clear what happens for unchanged files. Leo writes an optional "changed" message and checks the file. That's all.

The point is that `at.writeOneAtCleanNode` updates the `_mod_time` uA using the *file system's* modification time. That has got to be correct regardless of whether `at.replaceFile` wrote the file!

edreamleo:
@boltex We can zoom tomorrow (or any time) if you like.

PR #4428, opened 2025-08-20
PR fixes #4427: Always mark changed @clean roots as dirty

### LINKED ISSUE #4427: When a Leo file is saved, mod_time UA is missing from @clean vnode if read from a 'Leo startup' and left unmodified.
This subtle (important!) bug is the culprit behind recent problems with the `_mod_time` uA (PR #4418).

The seemingly simple fixes are the result of several hours of discussion between Félix and myself.

- `at.readOneAtCleanNode` must *always* call `root.v.setDirty` if the file has changed.
- `c.refreshFromDisk` must call `p.v.setDirty`.

edreamleo:
See #4427 for a full discussion.

I am closing this PR. Félix will create a new PR.

### DIFF
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -403,6 +403,7 @@ def refreshFromDisk(self: Self,
     # Create the 'Recovered Nodes' tree.
     c.fileCommands.handleNodeConflicts()
     c.setChanged()
+    p.v.setDirty()  # #4427
     c.redraw()
     c.undoer.clearAndWarn('refresh-from-disk')
 #@+node:ekr.20210610083257.1: *3* c_file.pwd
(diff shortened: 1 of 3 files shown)


edreamleo:

boltex:
I've slept on it and have come to realize this is the wrong solution

The node should not be marked as dirty.
Only the commander should be marked dirty, as it was before.
 
The readoneaycleannode method should set the ua upon reading, ( to the timestamp it is when read) so that when the leo file is saved, the new leo file contents is saved with this new ua, (without rewriting the external file 'again' for nothing except for generating the missing ua. )

Of course writeoneatcleannode still needs to generate the ua when writing as it currently is. It's just that reading an atclean node should set a ua on the node  in case it is not further modified until the user saves the leo file itself.

I will make a new pull request tonight replacing this one.

Sorry I didn't realize this last night! ( it was getting late and I couldn't think straight)

edreamleo:
Closing this PR in favor of an upcoming PR by Félix.

PR #4430, opened 2025-08-20
PR fixes #4398: coloring docstrings fails after cutting/pasting text

### LINKED ISSUE #4398: Cutting or pasting tesxt in Python docstrings confuses the colorizer
See PR #4430.

There are several ways this can happen:

- Cutting or pasting text.
- Adding a `*` character.

As a first step, it would be good to do the following:

- Cache the trace that would be produced if `--trace=coloring` were in effect.
- Add a command to print the cached trace.

This scheme would allow me to see what has *already* happened when the colorizer gets confused.

edreamleo:
See #4398.

- [x] Call `c.recolor` after cutting/pasting text.
     It's hard to reproduce the bug, but this seems like a good start.

### DIFF
--- a/leo/core/leoFrame.py
+++ b/leo/core/leoFrame.py
@@ -503,6 +503,7 @@ class LeoFrame:
             p.v.b = w.getAllText()
             u.afterChangeBody(p, 'Cut', bunch)
         # If it's the headline, the headline has not officially changed yet.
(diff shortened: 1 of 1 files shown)


tbpassin (approved):

edreamleo:
@tbpassin Many thanks for your review!

PR #4433, opened 2025-08-21
Fixes #4427 : Set mod_time UA in @clean vnodes that were read from leo startup and left unmodified.

### LINKED ISSUE #4427: When a Leo file is saved, mod_time UA is missing from @clean vnode if read from a 'Leo startup' and left unmodified.

boltex:
vnode's UA is now set to new_mod_time for a proper default upon reading, until possible further modifications.

Fixes #4427

### DIFF
--- a/leo/core/leoAtFile.py
+++ b/leo/core/leoAtFile.py
@@ -696,13 +696,12 @@ class AtFile:
 
         # #4385: Do nothing if the file has not changed.
         try:
-            old_mod_time = root.v.u['_mod_time']  # #4385
-            g.trace(f"{old_mod_time:20} {root.h}")
+            old_mod_time = root.v.u['_mod_time']  # #4385 The file's *last-seen* mod time.
         except Exception:
             old_mod_time = None
-        new_mod_time = g.os_path_getmtime(fileName)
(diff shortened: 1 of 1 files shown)


boltex (commented):
@edreamleo
I've made sure to run the unit tests :)

I've also made sure to try it out extensively in all possible scenarios.

This single line fix solves the problem, and also preserves the "isDirty" nodes flags and "hasChanged" commander flags that were prevalent for many years. (no need to force setDirty on anything new compared to past Leo versions) :
*Thus making the mod_time UA system completely invisible/transparent to the user's experience compared to 6.8.5.*

boltex:
I've also cleaned up the two traces in leoAtFile.py.

edreamleo (approved):
Excellent work!

PR #4434, opened 2025-08-21
Fixes #4426: Set the external files controller's timestamp for external files when opening a leo file instead of leaving the efc's _time_d untouched.

### LINKED ISSUE #4426: Dialog pops up when a tab is *re-opened* where its external files contains changes while leaving other tabs opened.
See PR #4434 

If leo is already running, and a new tab is opened, closed, (while leaving other tabs) and re-opened, ...for a leo file which contains an at-clean node that had changes in its external file since last time... **a dialog pops up!**

That dialog is the same that pops up when a leo file is already opened and an external file change is detected from one of its at-file/at-clean node.

That dialog should not appear. Moreover, we see that the changes were already loaded and choosing yes/no changes nothing.

(This has to do with the interaction between opening a leo file, and the default/previous values of **_time_d** in the **efc** if the file was previously opened in this Leo session.)

boltex:
(same effect as leaving untouched if was not already opened because efc's "has_changed" defaults to set_time's own default if _time_d entry is unset.)
Fixes #4426

### DIFF
--- a/leo/core/leoAtFile.py
+++ b/leo/core/leoAtFile.py
@@ -432,8 +432,14 @@ class AtFile:
         t1 = time.time()
         c.init_error_dialogs()
         files = at.findFilesToRead(root, all=True)
+        efc = g.app.externalFilesController
         for p in files:
             at.readFileAtPosition(p)
+            # An @<file> was read: In case this was a tab opened earlier, for which efc had entries,
+            # we need to update the external files controller's timestamp for those external files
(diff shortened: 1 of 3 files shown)


boltex (commented):
@edreamleo
I've made sure to run the unit tests and try out the expected behavior in various circumstances. :)

edreamleo (approved):
Excellent code change and comments.

PR #4448, opened 2025-09-02
PR fixes #4443, part 2

### LINKED ISSUE #4443: Problems with undo when executing scripts
**Problem 1**: Fixed via PR #4444.

<details>
<br>

While undoing in a script node:
```console

  File "C:\Repos\leo-editor\leo\core\leoGlobals.py", line 320, in new_cmd_wrapper
    func(self, event=event)
    ~~~~^^^^^^^^^^^^^^^^^^^
  File "C:\Repos\leo-editor\leo\core\leoUndo.py", line 1727, in undo
    u.undoHelper()
    ~~~~~~~~~~~~^^
  File "C:\Repos\leo-editor\leo\core\leoUndo.py", line 1885, in undoGroup
    newP = u.newP.copy()  # u.newP must exist now.
           ^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'copy'
```

</details>

**Problem 2**: Undo does not work reliably after the `execute-script` command.

edreamleo:
See #4443.

**The constraints**

- Leo must never call `u.afterUndoGroup` unless it has first called `u.beforeChangeGroup`.
- Conversely, Leo must never call `u.beforeUndoGroup` unless will later call `u.afterChangeGroup`.

**What I did**

- [x] Fix the same bug in three different commands.
- [x] Review all calls to `u.afterChangeGroup` to ensure the above constraints are satisfied.

### DIFF
--- a/leo/commands/editCommands.py
+++ b/leo/commands/editCommands.py
@@ -3898,14 +3898,16 @@ class EditCommandsClass(BaseEditCommandsClass):
         c, u, undoType = self.c, self.c.undoer, 'clear-all-uas'
         # #1276.
         changed = False
-        u.beforeChangeGroup(c.p, undoType)
+        first_p = c.p
         for p in c.all_unique_positions():
             if p.v.u:
+                if not changed:  # #4443
+                    u.beforeChangeGroup(first_p, undoType)
+                    changed = True
                 bunch = u.beforeChangeUA(p)
                 p.v.u = {}
                 p.setDirty()
                 u.afterChangeUA(p, undoType, bunch)
-                changed = True
         if changed:
             c.setChanged()
             u.afterChangeGroup(c.p, undoType)
--- a/leo/core/leoImport.py
+++ b/leo/core/leoImport.py
@@ -1130,9 +1130,6 @@ class LeoImportCommands:
             g.es_print(f"No importer class for @language {language}")
             return
         importer = importer_class(c)
-
-        # Handle undo.
-        u.beforeChangeGroup(p, undoType)
         try:
             old_p = p.copy()
             c.selectPosition(p)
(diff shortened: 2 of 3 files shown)


edreamleo:

boltex (approved):
Why did I not spot those when converting to typescript!? haha!
(I've also ported those changes to LeoJS 😄 )

tbpassin (commented):
editCommands
-------------------

I cannot understand the code around lines 3901 - 3906:
```
        first_p = c.p
            if p.v.u:
                if not changed:  # #4443
                    u.beforeChangeGroup(first_p, undoType)
                    changed = True
                bunch = u.beforeChangeUA(p)
                p.v.u = {}
                p.setDirty()
                u.afterChangeUA(p, undoType, bunch)
        if changed:
            c.setChanged()
            u.afterChangeGroup(c.p, undoType)
```

It's probably because of other things the surrounding code does.
1. `u.beforeChangeGroup(first_p, undoType)` is not matched by an matching `afterChangeGroup()` for each changed position. There is only one call.
2. Why is p.v.u set to `{}` for every position, regardless of whether `changed == True`?

boltex:
@tbpassin The beforeChangeGroup is meant to be called only once. the 'changed' flag assures this is the case. 

That 'group' of actions are the changes in UA, each fenced with a before/after ChangeUA. 

When all is done for none/one/many changes in uas as needed with the loop for p in all unique positions, ...then if there was at least one action to be undone in that 'group' of actions, then the changed flag would be set, and therefore the ultimate, unique 'afterchangeGroup' is performed.

boltex:
@tbpassin note the `if p.v.u:` which would be false if anything else than `{ } `, therefore not doing all positions. Only changing the required ones, if any.

edreamleo:
@boltex Thanks for your approval and your (correct) explanations.

PR #4453, opened 2025-09-11
PR fixes #4450: unique_subtree

### LINKED ISSUE #4450: unique_subtree has no parameter 'copy' but refers to it.
See PR #4453.

In leoNodes.py, (in the p.generators section) **unique_subtree** has no parameter 'copy' but refers to it like **self_and_subtree**, **subtree**, etc...

edreamleo:
See #4450.

### DIFF
--- a/leo/core/leoNodes.py
+++ b/leo/core/leoNodes.py
@@ -661,18 +661,16 @@ class Position:
 
     unique_vnodes_iter = unique_nodes
     #@+node:ekr.20091002083910.6103: *4* p.unique_subtree
-    def unique_subtree(self) -> Generator:
+    def unique_subtree(self, copy: bool = True) -> Generator:
         """Yield p and all other unique positions in p's subtree."""
         p = self
         seen = set()
         for p in p.subtree():
             if p.v not in seen:
                 seen.add(p.v)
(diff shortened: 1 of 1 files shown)


edreamleo:
@boltex Ready for review!

leo/core/leoNodes.py
```diff
@@ -661,18 +661,16 @@ def unique_nodes(self) -> Generator:
 
     unique_vnodes_iter = unique_nodes
     #@+node:ekr.20091002083910.6103: *4* p.unique_subtree
-    def unique_subtree(self) -> Generator:
+    def unique_subtree(self, copy: bool = True) -> Generator:
         """Yield p and all other unique positions in p's subtree."""
         p = self
         seen = set()
         for p in p.subtree():
```
boltex: 'copy' parameter should be passed to p.subtree also. otherwise a 'false' will always be overturned by the 'true default of p.subtree.

boltex (changes requested):

leo/core/leoNodes.py
edreamleo: Thanks. Fixed in rev d339a51.

boltex (approved):
Looks good ! 😄

PR #4456, opened 2025-09-22
Added 'leo_to_html_outline_viewer.py' plugin.

boltex:
I Added it to default so its single entry (Export HTML Outline Viewer) to the Export Files menu is available out of the box.

The Export Files submenu contains already relatively less useful exporters for today's modern use of Leo, so I think it is worth having this new menu entry as a default _in the Export Files submenu_ because it offers a 'discoverable' way to export for relatively new (or older) users to use, and even more, as it is a way to browse outlines for non-leo users.  

The plugin also makes the command export-html-outline-viewer in the minibuffer.

I Also fixed menu bug in leo_to_rtf.py plugin which made the command invisible in the menus when enabled.

(I made sure to run unit tests successfully before submitting this) 😄

### DIFF
--- a/leo/config/leoSettings.leo
+++ b/leo/config/leoSettings.leo
@@ -3725,7 +3725,7 @@ nodetags.py
 quicksearch.py      # Nav pane.
 todo.py             # Task pane.
 viewrendered.py     # (Or viewrendered3.py) Plugins menu support.
-</t>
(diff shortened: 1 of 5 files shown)


edreamleo (approved):
My commit fixes several minor complaints.

Many thanks for this impressive work.

PR #4460, opened 2025-10-08
PR: Fix mypy and pylint complaints for Python 3.14

edreamleo:
Upgrading mypy and pylint generated a few minor complaints.

This PR fixes them.


edreamleo:
@boltex These changes probably do not affect either LeoJS or LeoInteg.

PR #4474, opened 2025-11-09
When importing, make sure adding @language and @tabwidth is not done at the end of a line of code, but instead on the start of a line.

### LINKED ISSUE #4473: Imports add the '@language xxx' at end of existing line instead of at start of newline if source ended without newline.
**See PR #4474** 

For any language being imported, if the source ended without a newline (i.e. its last line of text in the source does not end with a newline) the sequence added in base_importer.py by those lines will not start on a newline. (and thus be concatenated with ths source's last line as per the screenshot below)

```python
            # Add trailing lines.
            if self.root.isAnyAtFileNode():  # #4385.
                parent.b += f"@language {self.language}\n@tabwidth {self.tab_width}\n"
```

In the mentioned screenshot below, the source is on the left, and the problem is underlined with the red arrow.

Image

boltex:
Fixes #4473

### DIFF
--- a/leo/plugins/importers/base_importer.py
+++ b/leo/plugins/importers/base_importer.py
@@ -131,6 +131,9 @@ class Importer:
 
             # Add trailing lines.
             if self.root.isAnyAtFileNode():  # #4385.
+                # Make sure parent.b ends with a newline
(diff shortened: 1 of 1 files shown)


edreamleo (approved):
Thanks. Good catch.

PR #4477, opened 2025-11-18
PR fixes #4476: adjacent @doc parts

### LINKED ISSUE #4449: Colorizer does not color correctly after @/@c pair
**EKR: This issue is on hold. I have not been able to reproduce it.**

A single "@" at the start of a line causes Leo to ignore the following line until the next line that starts with "@c".  After the "@" the colorizer colors the lines as if they were comments.  But the colorizer does not seem to recognize the "@c" and does not resume colorizing the lines following the "@c" as they should be.

Here is an example of body text that illustrates this issue:

```import sys

g.es(sys.executable)

@
g.es('should not print')

@c
g.es("All this text should be colorized as code but it isn't")
```

The last line of code executes but is colorized as if it were a comment.

### LINKED ISSUE #4476: @ folowed by @ is not colored as a Leo keyword
See PR #4477.

Example
```python
@
Line 1
@
Line 2
```
The second `@` is colored as part of the @doc part, not as a Leo keyword.

I wonder whether this could be related to #4449.

edreamleo:
See #4476. This PR might fix #4449!

- [x] In `jedit.match_doc_part`: Test for `@` without using `g.match_word`. It doesn't work properly.
- [x] In `jedit.restartDocPart`: Add special case for `@`.
- [x] In both methods above, omit `i` where it is known to be zero and simplify the test for `@`.

### DIFF
--- a/leo/core/leoColorizer.py
+++ b/leo/core/leoColorizer.py
@@ -1668,7 +1668,7 @@ class JEditColorizer(BaseColorizer):
             return 0
         # Test for @ or @doc and set j.
         # Careful: g.match_word doesn't test for '@' correctly.
-        if s[0] == '@' and (len(s) < 2 or s[1] in (' ', '\t', '\n')):
+        if s[0] == '@' and (len(s) == 1 or s[1] in ' \t\n'):  # 4476
             j = 1
         elif g.match_word(s, i, '@doc'):
             j = 4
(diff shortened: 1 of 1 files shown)


tbpassin (approved):
This PR seems OK.  But wouldn't it have been better to fix `g.match_word()` so it can match a bare `@`?

edreamleo:
@tbpassin I rejected "improving" `g.match_word()` because changing that function might break existing code.

I considered defining a  new (base) colorizer method that would test for a single non-alphabetic character. This would reduce a bit of code duplication, but I think the PR's code is good enough.

PR #4482, opened 2025-11-23
PR fixes #4472: make_leo_stubs script

### LINKED ISSUE #4472: Improve pyflakes
My study of pylint (see #4468) shows that pylint does not trust Python annotations when checking code.

Annotations should cut the Gordian Knot of type inference. A checker that *did* trust annotations would be *much* faster than pylint or mypy. I'll improve pyflakes locally, via monkey patches. There should be no need for a pyflakes PR.

<details><summary><b>Pyflakes fails to find obvious errors</b></summary>
<br>

```python

class Test:  #@
    a: int = 666

test = Test()
print(test.b)  # Wrong.

def f(arg: Test) -> int:
    return arg.b  # Wrong.

val = f(test)
print(val)
```

Indeed, pyflakes does *nothing* to check attribute access. There is no `Checker.ATTRIBUTE` method, so pyflakes merely visits the children of `ast.ATTRIBUTE` nodes.

</details>

<details><summary><b>Monkey-patching pyflakes is easy!</b></summary>
<br>

Here is `@button patch-pyflakes`:
```python
g.cls()
if c.changed:
    c.save()

import ast
from pyflakes.api import check
from pyflakes.checker import Checker

<< define test >>
@others  # Define patched_ATTRIBUTE

g.funcToMethod(patched_ATTRIBUTE, Checker, 'ATTRIBUTE')
check(test, filename='pyflakes_test.py')
```
The new `Checker.ATTRIBUTE` visitor:
```python
def ATTRIBUTE(self, node) -> None:
    if isinstance(node.value, ast.Name):
        g.trace(f"{node.ctx.__class__.__name__} {node.value.id}.{node.attr}")
    self.handleChildren(node)
```
And here is the output:
```console
ATTRIBUTE:  Load test.b
ATTRIBUTE:  Load arg.b
```

</details>

<details><summary><b>Notes</b></summary>
<br>

**stubgen experiments failed**

pylint's [--prefer-stubs](https://pylint.pycqa.org/en/latest/user_guide/configuration/all-options.html#prefer-stubs) command-line would be another way of helping pylint. Alas,  [stubgen](https://mypy.readthedocs.io/en/stable/stubgen.html) seems to be slow and buggy.

**Original design**

[This gemini 3 thread](https://gemini.google.com/share/343f1a7fb5d8) was a starting point.

- Use annotations when checking function calls and attributes.
- Create `leo/scripts/quick_check.py`: It could use the [astroid](https://pypi.org/project/astroid/) package if necessary.
- A local script, `quick-check.cmd`, would run after mypy. This script would check only changed files.
- Another local script would run after mypy and check *all* of Leo's source files.

**The program under test** is the program being checked. Eventually, I'll test the script on Leo itself.

**Data**

- The **files list** is a *static* list of program files in the program under test.
  Command-line arguments might affect this list.
- The **file description cache** will have one entry for each of file in the files list.
- The **class cache** is a dict whose keys are (full) paths to files and whose values will be inner **per-file class description dicts**.

**Passes**

- Pass 0: Init and compute the files list.
- Pass 1: Update the entries in the class cache.
- Pass 2:
  - Resolve all class annotations to the proper entry in the class cache.
  - Check each file using the resolved classes.

</details>

edreamleo:
See #4472.

- [ ] Create stubs for Leo's core files using `leo.scripts.make_leo_stubs`, a thin wrapper around mypy's [stubgen](https://mypy.readthedocs.io/en/stable/stubgen.html) utility.
- [ ] Test these stubs with pylint's `--prefer-stubs` option.

### DIFF
--- a/leo/scripts/make_leo_stubs.py
+++ b/leo/scripts/make_leo_stubs.py
@@ -1,0 +1,28 @@
+#@+leo-ver=5-thin
+#@+node:ekr.20251123134502.1: * @file ../scripts/make_leo_stubs.py
+"""
+A script to create stub (.pyi) files for Leo's most important files.
+"""
+
+import glob
+import os
+from leo.core import leoGlobals as g
+from leo.plugins.importers.python import Python_Importer
+
+g.cls()
+
+#@+<< calculate test files >>
+#@+node:ekr.20251123141418.1: ** << calculate test files >>
+core_dir = g.app.loadDir
+core_files = glob.glob(f"{g.app.loadDir}{os.sep}*.py")
+if 1:
+    include = ['leoApp', 'leoAtFile', 'leoCommands']
+    test_files = [os.path.normpath(z) for z in core_files if any(z2 in z for z2 in include)]
+else:
+    exclude = ['__', 'runLeo', 'leoclient', 'leoHistory', 'leoJupytext',
+        'leoPymacs', 'leoQt', 'leoTips', 'leoVersion',
+    ]
+    test_files = [os.path.normpath(z) for z in core_files if not any(z2 in z for z2 in exclude)]
+g.printObj(test_files, tag='Test files...')
+#@-<< calculate test files >>
+#@-leo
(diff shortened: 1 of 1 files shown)


edreamleo:
I am going to close this idea. Stub files seem like the wrong way to improve pylint. Improving pyflakes seems like a more promising path.

PR #4521, opened 2026-01-26
Fix two more quirps

edreamleo:
Fix several quirps:

- [x] Improve 'line-number' docstring.
- [x] `LM.openWorkBook`: Adjust `g.app.numberOfUntitledWindows`.

### DIFF
--- a/leo/commands/editCommands.py
+++ b/leo/commands/editCommands.py
@@ -1602,7 +1602,7 @@ class EditCommandsClass(BaseEditCommandsClass):
     # @+node:ekr.20150514063305.247: *4* ec.lineNumber
     @cmd('line-number')
     def lineNumber(self, event: LeoKeyEvent) -> None:
-        """Print the line and column number and percentage of insert point."""
+        """Print the character, line number, column number and total number of columns."""
         k = self.c.k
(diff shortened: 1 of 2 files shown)


edreamleo:
@boltex Ready for your review.

leo/core/leoApp.py
boltex, line 2499: this part of the branch also creates an empty file. (requiring incrementing also the numberOfUntitledWindow)

leo/commands/editCommands.py
```diff
@@ -1602,7 +1602,7 @@ def copyGnx(self, event: LeoKeyEvent) -> None:
     # @+node:ekr.20150514063305.247: *4* ec.lineNumber
     @cmd('line-number')
     def lineNumber(self, event: LeoKeyEvent) -> None:
-        """Print the line and column number and percentage of insert point."""
+        """Print the character, line number, column number and total number of columns."""
```
boltex: The last part is not the number of columns, but instead the total lenght, perhaps it should be "**Print the character at the insert point, its row and column, and the total number of characters.**"

boltex (changes requested):
I've added comments for small fixes.

PR #4523, opened 2026-01-27
PR Fixes #4522: contract-parent

### LINKED ISSUE #4522: Contract-parent leaves outline without a selection.
See PR #4523.

Spotted by JKN https://groups.google.com/g/leo-editor/c/QGr4Kzrmf8o

edreamleo:
See #4522.

**Discussion**

*Warning*: All changes to drawing code have the potential to corrupt data. This probably happened once to me while working on this PR. Careful testing and analysis are required.

Only the changes to `leo/commands/commanderOutlineCommands.py` are likely to affect LeoJS directly. 

**Straightforward changes to Leo's core and Qt gui code**

- [x] Call `c.redraw` instead of `c.redraw_after_contract`.
- [x] Retire `c.redraw_after_contract`, and `c.redraw_after_expand`, making them synonyms for `c.redraw`.
- [x] Retire `qtree.redraw_after_contract`, `qtree.redraw_after_expand` and `qtree.update_expansion`.
   No script should ever calls these methods.
- [x] `c.redraw`: Add comment that `redraw_after_select` and `redraw_after_head_changed` must remain as they are.
   Making them synonyms for `c.redraw` would break the Qt gui!
- [x] `c.expandToLevel` always calls `c.redraw`. This might be an ancient bug.
- [x] Don't execute one slow unit test by default. This takes about 10 seconds on my machine!

**Potentially dangerous changes to `leo/plugins/qt_tree.py`**

- [x] `qtree.onItemCollapsed` and `qtree.onItemExpanded`: always call `c.redraw` instead of `c.outerUpdate`.
  The danger: too many redraws. Tests with `--trace=drawing` show that all is well.

**Straightforward changes to the `CTree` class in `leo/plugins/cursesGui2.py`**

I have tested these changes by hand using `--gui=console`.

- [x] Retire `CTree.redraw_after_contract` and `CTree.redraw_after_expand`.
- [x] Replace `CTree.busy()` by a single test on `CTree.redrawing`, removing three unnecessary lockouts.
- [x] Simplify `CTree.redraw_after_select`.

### DIFF
--- a/leo/commands/commanderOutlineCommands.py
+++ b/leo/commands/commanderOutlineCommands.py
@@ -556,7 +556,7 @@ def expandAllHeadlines(self: Cmdr, event: LeoKeyEvent = None) -> None:
     while p:
         c.expandSubtree(p)
         p.moveToNext()
-    c.redraw_after_expand(p0)  # Keep focus on original position
+    c.redraw(p0)  # Keep focus on original position
     c.expansionLevel = 0  # Reset expansion level.
 
 
@@ -653,7 +653,7 @@ def expandNode(self: Cmdr, event: LeoKeyEvent = None) -> None:
     p = c.p
     c.endEditing()
     p.expand()
-    c.redraw_after_expand(p)
+    c.redraw(p)
     c.selectPosition(p)
 
 
@@ -683,7 +683,7 @@ def expandNodeOrGoToFirstChild(self: Cmdr, event: LeoKeyEvent = None) -> None:
     c.endEditing()
     if p.hasChildren():
         if p.isExpanded():
-            c.redraw_after_expand(p.firstChild())
+            c.redraw(p.firstChild())
         else:
             c.expandNode()
 
--- a/leo/commands/commanderOutlineCommands.py
+++ b/leo/commands/commanderOutlineCommands.py
@@ -497,9 +497,9 @@ def contractNode(self: Cmdr, event: LeoKeyEvent = None) -> None:
     c = self
     p = c.p
     c.endEditing()
-    p.contract()
-    c.redraw_after_contract(p)
     c.selectPosition(p)
+    p.contract()
+    c.redraw()
 
 
 # @+node:ekr.20040930064232: *3* c_oc.contractNodeOrGoToParent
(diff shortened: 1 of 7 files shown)


edreamleo:
@boltex This PR is ready for review, but more testing would be wise.

tbpassin (approved):

boltex:
@edreamleo I'll switch to this branch and test it out :)

edreamleo:

edreamleo:
@boltex @tbpassin This PR is ready for review. I plan no further changes.

@boltex I'll wait for your review before merging.

tbpassin (approved):
I tested various expand/contract commands on LeoPyRef.  I used this outline because it was an easy way to get a large number of expanded nodes. They seemed fast enough to me. I didn't check for any data corruption, which seems unlikely to me.

boltex:
@edreamleo I've tested contract-parent also with leopyref because its large and has many @files etc...

### Contract parent

I've tried expanding many levels to have a large number of nodes opened and used the command contract-parent multiple times, in many nodes - and @file / @clean nodes. 

Saving leopyref (thus forcing writing changed @<files> nodes) and comparing the diff of the whole repo led me to see there were no changes (no data loss in external files) provoked by doing all this.

So contract-parent looks fixed...

### New bug found

On the other hand, I found a (weird) bug with expand-to-level-X : **redoing the same expand-to-level consecutively does nothing** after the first time. Doing **another differenf one** of those  expand-to-level-X commands then works! 

E.g.: if you just did expand-to-level-2 with the minibuffer (alt-x), on a node, let's say 'code' in the screenshot below, then contract it back manually by clicking on the caret (or select a different contracted node), and redo (alt-x) expand-to-level-2 to try to also expand it up to level 2: it will not do anything - as many tries as you want, but doing a different  expand-to-level-X, let's say  expand-to-level-3, will then work! 

image

Easy to spot and fix in the code fix: In leoCommands.py's **expandToLevel** method, the check if c.expansionLevel != old_expansion_level to redraw is problematic: the node could have been contracted manually by clicking , and/or another node could have been selected.

So in my opinion, this should always redraw.

I'll approve the merge after that small fix no problem 😄

edreamleo:
@boltex @tbpassin Rev a044f7d appears to fix the expansion bug.

This rev makes no changes to the logic  involving `c.expansionLevel` and `c.expansionNode`, as I verified with a cff command.

I am going to merge this PR now so that we may all test the logic.

Issue #4525, opened 2026-01-28
Duplicate code for forward/back history commands.


boltex:
See PR #4526.

@edreamleo This is very straightforward: In **CommanderEditCommands.py**, the node **c_ec.goToNext/PrevHistory** contains

```python
@g.commander_command('goto-next-history-node')
def goToNextHistory(self: Self, event: LeoKeyEvent = None) -> None:
    """Go to the next node in the history list."""
    c = self
    c.nodeHistory.goNext()

@g.commander_command('goto-prev-history-node')
def goToPrevHistory(self: Self, event: LeoKeyEvent = None) -> None:
    """Go to the previous node in the history list."""
    c = self
    c.nodeHistory.goPrev()
```

While in **commanderOutlineCommands.py**, the node **c_oc.Goto commands** contains the exact same code

```python
@g.commander_command('go-forward')
def goNextVisitedNode(self: Cmdr, event: LeoKeyEvent = None) -> None:
    """Select the next visited node."""
    c = self
    c.nodeHistory.goNext()

@g.commander_command('go-back')
def goPrevVisitedNode(self: Cmdr, event: LeoKeyEvent = None) -> None:
    """Select the previously visited node."""
    c = self
    c.nodeHistory.goPrev()
```

Only one set of those commands could be kept, and still support both names aliases by stacking decorators like this _(let's say we keep the ones in commanderOutlineCommands because it navigates us in the outline, without performing any changes nor edits)_
```python
@g.commander_command('goto-next-history-node')
@g.commander_command('go-forward')
def goNextVisitedNode(self: Cmdr, event: LeoKeyEvent = None) -> None:
    """Select the next visited node."""
    c = self
    c.nodeHistory.goNext()

@g.commander_command('goto-prev-history-node')
@g.commander_command('go-back')
def goPrevVisitedNode(self: Cmdr, event: LeoKeyEvent = None) -> None:
    """Select the previously visited node."""
    c = self
    c.nodeHistory.goPrev()
```

And thus the ones in **CommanderEditCommands.py** could be deleted without much trouble, as they are used only in one place directly : **in the nav_qt plugin. on lines 96 and 97:**
```python
  act_l.triggered.connect(lambda checked: c.goToPrevHistory())
  act_r.triggered.connect(lambda checked: c.goToNextHistory())
```

so those line should become 
```python
  act_l.triggered.connect(lambda checked: c.goPrevVisitedNode())
[...]

PR #4526, opened 2026-01-29
PR: Remove duplicate goto commands from commanderEditCommands.py

edreamleo:
See #4525.

- [x]  Remove duplicate goto commands from `commanderEditCommands.py`, retaining the spelling of all commands.

### DIFF
--- a/leo/commands/commanderEditCommands.py
+++ b/leo/commands/commanderEditCommands.py
@@ -646,21 +646,6 @@ def findMatchingBracket(self: Self, event: LeoKeyEvent = None) -> None:
         g.MatchBrackets(c, p, language).run()
 
 
-# @+node:ekr.20110402084740.14490: ** c_ec.goToNext/PrevHistory
-@g.commander_command('goto-next-history-node')
-def goToNextHistory(self: Self, event: LeoKeyEvent = None) -> None:
-    """Go to the next node in the history list."""
-    c = self
-    c.nodeHistory.goNext()
-
-
-@g.commander_command('goto-prev-history-node')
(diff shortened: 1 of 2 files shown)


edreamleo:
@boltex @tbpassin The `test_leo.py` found a quirp with this PR after I merged it: `c.goToNextHistory` and `c.goToPrevHistory` no longer exist.

This defect can't be remedied with assignments such as:
```python
goToNextHistory = goNextVisitedNode
gotoPrevHistory = goPrevVisitedNode
```
Why? Because the `@g.commander_command` decorator injects the symbol into the commands class.

The only way to fix this compatibility problem would be to define two new methods, which would be pretty much the status quo ante.

**Summary**

I don't want to re-duplicate the code, but I suppose we should document the removal of the two old methods.

edreamleo:
@boltex @tbpassin Rev bcc9088 fixes the bug reported by `test_leo.py`. I'll document the compatibility problem next.

edreamleo:
@boltex @tbpassin Here is the new language in the "What's new" section of `LeoDocs.leo":
```
**Minor breaking API change**

- Retire two Commands methods: ``c.goToNextHistory`` and ``c.goToPrevHistory``.
  Use ``c.goNextVisitedNode`` and ``c.goPrevVisitedNode`` instead.
```
In short, this wasn't the most productive PR in Leo's history.

tbpassin:
Will this change affect the behavior of the forward/back navigation button?

edreamleo:
@tbpassin Good question. The answer should be no.  The only problem that I see is that two Commander methods no longer exist. It's unlikely that any scripts would use those methods, but the possibility exists and should be documented.  Alternatively, we could revert this PR.

PR #4528, opened 2026-02-13
Escape HTML entities BEFORE converting URLs to clickable links!

boltex:
Fixed HTML exporter.

Now displays code like this properly (Doh!): 

`<div>hello with visible div tags greater than/ lesser than , etc. </div>`


edreamleo (approved):
This is your code. I'll leave you to it :-)

PR #4533, opened 2026-02-17
PR: Simplify the classes in leoColorizer.py

edreamleo:
Recent pylint warnings have alerted me to infelicities in the class structure in `leoColorizer.py`.

- [x] Make `PygmentsColorizer` a subclass of `JEditColorizer`.
- [x] Move all Qt-specific code out of the `BaseColorizer` class.
- [x] Move common ivars into the `BaseColorizer` class.
- [x] Rename `mainLoop` to `pygmentsMainLoop`.

Use Leo's `git-diff-pr` command to see the diffs clearly.

### DIFF
--- a/leo/core/leoColorizer.py
+++ b/leo/core/leoColorizer.py
@@ -346,75 +346,6 @@ class BaseColorizer:
             s = s[1:]
         return s.replace(' ', '').replace('-', '').replace('_', '').lower().strip()
 
-    # @+node:ekr.20260216112007.1: *3* BaseColorizer.setTag
-    def setTag(self, tag: str, s: str, i: int, j: int) -> None:
-        pass
-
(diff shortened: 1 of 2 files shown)


edreamleo:

The flat diffs are complex. Leo's `git-diff-pr` shows that many diffs are the result of moving methods from the `BaseColorizer` class to the `JEditColorizer` class.

@boltex This PR likely does not affect LeoJS, but I always value your comments.

tbpassin (commented):
I can't tell why there are so many blocks of default color definitions.  Shouldn't they be set in LeoSettings.leo?

edreamleo:
@tbpassin The default color definitions are fall-backs in case they are not set in LeoSettings.leo.

PR #4548, opened 2026-02-25
Add metadata files for desktop GNU/Linux packaging

bkmgit:
See:
- https://docs.fedoraproject.org/en-US/packaging-guidelines/AppData/
- https://docs.fedoraproject.org/en-US/packaging-guidelines/#_desktop_files


edreamleo (approved):
Thanks!

Issue #4549, opened 2026-02-25
Potential bug in JEditColorizer.match_eol_span: Missing character match check


wangzhaohe:
Hi @edreamleo  and the Leo team,

I hope you are doing well. I believe I’ve found a small logic regression in `leo/core/leoColorizer.py` that affects syntax highlighting for several languages (specifically those using `//` for comments, like AsciiDoc, JavaScript, C, and TypeScript).

**The Issue:** In certain modes, a single forward slash `/` causes the rest of the line to be highlighted as a comment, even when it is not followed by a second slash. For example, in an AsciiDoc file, a string like `"Hello /World!"` results in `/World!"` being colored as a comment.

**The Cause:** Upon investigating `leoColorizer.py`, I noticed that in the `match_eol_span` method, the actual sequence matching logic seems to have been commented out. Because of this, the method treats any trigger character as a successful match for the entire rest of the line without verifying the full sequence (e.g., checking if `/` is actually followed by another `/` to form `//`).

**Suggested Fix:** In `leo/core/leoColorizer.py`, within the `match_eol_span` method, I restored the missing check:

Python

```
# Around line 2028 in leoColorizer.py
if not g.match(s, i, seq):
    return 0
```

I have tested this locally with AsciiDoc, JavaScript, and TypeScript, and it correctly resolves the highlighting issue without breaking existing functionality.

I would be happy to submit a Pull Request for this if you agree with the fix. Thank you for your continued work on Leo!

Best regards,
wangzhaohe

wangzhaohe:
I decided to go ahead and submit a Pull Request (#4550) while the details are still fresh in my mind. Re-analyzing the colorizer logic later would be quite time-consuming, so I wanted to get the fix documented and tested now.

I've synced it with the latest devel branch and verified it locally. Hope this helps!

edreamleo:
Fixed via PR #4550.

PR #4550, opened 2026-02-25
fix: Restore missing sequence check in JEditColorizer.match_eol_span

wangzhaohe:
This PR addresses the syntax highlighting issue discussed in Issue #4549. I have synced this branch with the latest devel branch and verified the fix locally with AsciiDoc and JS.


edreamleo (approved):
Many thanks for this fix.

PR #4552, opened 2026-03-06
Ensure blank lines between imported markdown sections

edreamleo:
- [x] Add do-nothing md_i.postprocess.
- [x] Add  corresponding unit test.

### DIFF
--- a/leo/plugins/importers/markdown.py
+++ b/leo/plugins/importers/markdown.py
@@ -135,6 +135,12 @@ class Markdown_Importer(Importer):
         assert 0 <= level < len(self.stack), (level, len(self.stack))
         return self.stack[level]
 
+    # @+node:ekr.20260305084156.1: *3* md_i.postprocess
+    def postprocess(self, parent: Position) -> None:
+        """
+        MarkDown_Importer.postprocess. Don't do anything!
+        """
+
     # @-others
 
 
--- a/leo/plugins/writers/markdown.py
+++ b/leo/plugins/writers/markdown.py
@@ -19,11 +19,9 @@ class MarkdownWriter(basewriter.BaseWriter):
         """Write all the *descendants* of an @auto-markdown node."""
         self.root = root
         self.write_root(root)
-        total = sum(1 for _ in root.subtree())
         count = 0
         for p in root.subtree():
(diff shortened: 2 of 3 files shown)


boltex (changes requested):
This works **but only in one direction**: Reading the .md file into the @auto adds the extra newlines at the end of body text when the markdown sections have extra newlines before the heading of another following section. 

But in the other direction. **when writing a markdown @auto node to file**, nodes that end with extra newlines do not see those extra newlines written to file: a single newline is present between then end of a section and the heading of the following section.

boltex:
@edreamleo  I've made some experiments and looked at the writer: it was enforcing 2 newlines for some reason instead of letting the user make sections ending with arbitrary newlines. I think this new simplified version could work:

```python
    def write(self, root: Position) -> None:
        """Write all the *descendants* of an @auto-markdown node."""
        self.root = root
        self.write_root(root)
        count = 0
        for p in root.subtree():
            count += 1
            if g.app.force_at_auto_sentinels:  # pragma: no cover
                self.put_node_sentinel(p, '<!--', delim2='-->')
            if self.placeholder_regex.match(p.h):
                # skip this 'placeholder level X' node
                pass
            else:
                self.write_headline(p)
                lines = p.b.splitlines(False)
                for s in lines:
                    if not g.isDirective(s):
                        self.put(s)
        root.setVisited()

```

edreamleo:
@boltex Thanks for your suggestion. Rev 46b9648 make the change you requested.

edreamleo:
@boltex The PR is ready for your review.

boltex:
@edreamleo Thanks, I'll review this and test it out carefully before the weekend.

boltex (approved):
I've played with it to test out importing, and writing markdown from @auto nodes and it seems to work nicely now. 

I've also compared with the old (devel) branch to see that the difference is really only allowing nodes to end with any number of trailing newlines. Thus not enforcing any sort of arbitrary standards like before.

Again, thanks for providing even better support for .md (markdown) files because it has become the standard for 'text based' documentation! :)

edreamleo:
@boltex Many thanks for your help and review.

PR #4553, opened 2026-03-12
Fix quicksearch.py body matches for Python 3.14

### LINKED ISSUE #4554: quicksearch.py no longer produces body text matches on Python 3.14
Nav pane on python 3.14 no longer produces body text matches.

Fix and details in PR #4553

gatesphere:
The quicksearch.py (Nav pane) plugin stopped producing matches for node body text after upgrade to python 3.14.

This is a result of a few behavior changes:

Python 3.14 now recognizes `/z` as a synonym for `/Z` in regexes (https://docs.python.org/3/whatsnew/3.14.html#re)

The `fnmatch.translate` function in Python 3.14 now returns strings that use the new `/z` (https://docs.python.org/3.14/library/fnmatch.html#fnmatch.translate).  The code in quicksearch.py's `doSearch` didn't account for the lowercase variant.

Simple fix, just replace both variants with empty strings.


edreamleo (approved):
Thanks for this fix.

PR #4563, opened 2026-03-22
Remove unneeded dependency

### LINKED ISSUE #4546: TK package on PyPI does not install Python-Tkinter
Tk installs TensorKit a deep learning package https://pypi.org/project/tk/

bkmgit:
tk is a machine learning library not the tkinter
framework which is part of base python.

closes #4546


edreamleo (approved):
Ah. At last I understand.
Many thanks for removing the unnecessary dependency.

PR #4564, opened 2026-03-22
Made the set_name_and_title helper method also set tooltip on frame's tab. (setTabName now accepts optional tooltip)

### LINKED ISSUE #4558: tooltip on window tab is not updated upon save.
Make a new untitled file, see the tooltip, save it, the tooltip remains untitled. 

Also for existing files (see below as I saved-as a different name)

Image

boltex:
Fixes #4558
I ran py -m unittest just in case: all good :)


edreamleo (changes requested):
With Leo's default settings for mypy, there is no need to indicate that kwargs can be None.

Please change `tooltip: str | None = None` to `tooltip: str = None`

edreamleo (approved):

PR #4566, opened 2026-03-22
PR fixes #4565: update @clean files

### LINKED ISSUE #4565: Updating an external @clean file containing a shared clone does not mark all @clean nodes dirty
See PR #4566.

**OP**

I'd like to use Leo to maintain 2 projects, that share some code. My idea was to have C-files with a node for code concerning the one file (e.g. file_1.c XOR file_2.c) and a clone node for common  code.

**EKR**

- I created two `@clean` test files containing a shared clone
- Updating either clone in Leo marks both `@clean` nodes dirty.
- Updating either external file updates the shared clone node and marks the (single) shared clone node dirty in Leo.
  The icons for the shared clone are dirty in both `@clean` files.

But (bug!) only one of the `@clean nodes` gets marked dirty.

Note: This bug likely does not depend on the value of `@bool report-changed-at-clean-nodes`

edreamleo:
See #4565.

- [x] Update dirty bits in `efc.idle_check_commander`.
- [x] `efc.idle_check_commander` calls `v.setAllAncestorAtFileNodesDirty` once with the computed `to_do_set` kwarg.
- [x] Rewrite `v.setAllAncestorAtFileNodesDirty`, adding the `to_do_set` kwarg.

Note: The new version of `v.setAllAncestorAtFileNodesDirty` scans for nested `@<file>` nodes to maintain compatibility with legacy code. This is a dubious choice, but it seems best.

### DIFF
--- a/leo/core/leoNodes.py
+++ b/leo/core/leoNodes.py
@@ -2735,19 +2735,27 @@ class VNode:
         Modified by EKR.
         """
         v = self
-        ### g.trace(v)  ###
+        to_do: list[VNode] = [v]
+        for v2 in v.parents:
+            to_do.append(v2)
         result: set[VNode] = set()
+        seen: set[VNode] = set()
 
-        def find_all_ancestor_at_file_nodes(v: VNode) -> None:
-            ### g.trace(v.h)
-            if v.isAnyAtFileNode():
-                result.add(v)
-            for parent_v in v.parents:
-                find_all_ancestor_at_file_nodes(parent_v)
+        def find_all_ancestor_at_file_nodes() -> None:
+            while to_do:
+                v2 = to_do.pop()
+                seen.add(v2)
+                if v2.isAnyAtFileNode():
+                    result.add(v2)
+                for parent_v in v2.parents:
+                    if parent_v not in seen:
+                        to_do.append(parent_v)
 
-        find_all_ancestor_at_file_nodes(v)
-        for v in list(result):
-            v.setDirty()
+        find_all_ancestor_at_file_nodes()
+        if not g.unitTesting:  ###
+            g.printObj([z.h for z in list(result)], tag=v.h)
+        for v2 in list(result):
+            v2.setDirty()
 
     # @+node:ekr.20040315032144: *4* v.setBodyString & v.setHeadString
     def setBodyString(self, s: object) -> None:
--- a/leo/core/leoNodes.py
+++ b/leo/core/leoNodes.py
@@ -2741,9 +2741,9 @@ class VNode:
         # #4565: Rewrite using a loop.
         while to_do:
             v2 = to_do.pop()
+            seen.add(v2)
             if v2.isAnyAtFileNode():
                 result.add(v2)
-                seen.add(v2)
             else:
                 for parent_v in v2.parents:
                     if parent_v not in seen:
--- a/leo/core/leoNodes.py
+++ b/leo/core/leoNodes.py
@@ -2732,19 +2732,16 @@ class VNode:
         """
         Original idea by Виталије Милошевић (Vitalije Milosevic).
 
-        rewritten by EKR.
+        #4565: Add the optional initial to_do set.
         """
         v = self
         result: set[VNode] = set()
         seen: set[VNode] = set([v.context.hiddenRootNode])
-        # to_do: list[VNode] = [v] + [z for z in v.parents]
         to_do_list: list[VNode] = list(to_do) if to_do else [v] + v.parents
         if to_do:
             for v2 in to_do:
                 to_do_list.extend(v2.parents)
         to_do_list = list(set(to_do_list))
-
-        # #4565: Rewrite using a loop.
         while to_do_list:
             v2 = to_do_list.pop()
             seen.add(v2)
--- a/leo/core/leoNodes.py
+++ b/leo/core/leoNodes.py
@@ -2735,24 +2735,24 @@ class VNode:
         #4565: Rewritten by EKR to use the to_do_set kwarg.
         """
         v = self
-        result: set[VNode] = set()
+        # Init seen and to_do_list.
         seen: set[VNode] = set([v.context.hiddenRootNode])
         to_do_list: list[VNode] = list(to_do_set) if to_do_set else [v] + v.parents
         if to_do_set:
             for v2 in to_do_set:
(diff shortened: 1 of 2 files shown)


edreamleo:
@boltex This PR is ready for review. This PR probably affects LeoJS.

boltex (approved):
Looks good!
- Also now part of LeoJS and leo-web 😄

edreamleo:
@boltex Thanks for your review. Please note the changes since your review.

leo/core/leoNodes.py
```diff
@@ -2728,28 +2728,32 @@ def saveCursorAndScroll(self) -> None:  # pragma: no cover
...
         """
         v = self
-        result: set[VNode] = set()
+        # Init seen and to_do_list.
         seen: set[VNode] = set([v.context.hiddenRootNode])
-        to_do: list[VNode] = [v] + [z for z in v.parents]
-        # #4565: Rewrite using a loop.
-        while to_do:
-            v2 = to_do.pop()
+        to_do_list: list[VNode] = list(to_do_set) if to_do_set else [v] + v.parents
```
boltex: that '+ parents' is redundant because below we also get the parents of this array's members.

boltex (approved):
See small comment below

boltex:
@edreamleo Recent changes now also part of leojs and leo-web :)

leo/core/leoNodes.py
edreamleo: @boltex Nice catch.

edreamleo: @boltex Fixed at Rev 75db230.

boltex (dismissed):
@edreamleo 

Altough this tries to fix the bug in a reasonable way, it does not solve it.

The other **cloned node** stays clean, so its parent \@\<file\> also stays clean. This only dirties the \@\<file\> node itself when external file changes are detected.

The code written would work (tryng to set all other \@\<file\> dirty if one of their descendant is dirty) if refreshing a node from file would dirty specific nodes inside the \@\<file\> node itself.

(... or I totally misunderstood the problem at the source of this issue)

boltex:
@edreamleo Oh ! my bad: It works for \@clean but not for \@file ! 

When changing the external file for an **\@clean node** in the text contained be a cloned node, that cloned node becomes dirty, and thus with the new algorithm, all other  \@\<file\> nodes  containing that clone become dirty.

(and thus are also saved with the changes when the user saves with CTRL+S)

But, when it's an **\@file node** being refreshed for external changes being detected,  that the top \@file node itself **only** gets dirty - instead of having specific descendant (cloned) node(s) it may contain get dirty.

boltex:
@edreamleo Could the \@file read algorithm check for previous body content when being about to fill up a body for a gnx that already exists in the gnx dict? 

(which is the only condition that makes a node a clone, ...that it is already in the outline with that same exact gnx)

Maybe easier said than done, but If it could simply make that check at that specific time, then that cloned node could be set dirty, at that time, thus acheiving the desired effect even for \@file nodes ! 😄

boltex:
Note: I just want to add that of course i understand \@clean nodes have a reference for comparing when reading from file, and thus can easily make any specific node dirty.

For \@file, there is no reference at all. So that is why changed \@file only dirtied their top \@file node when changes were read back into the node.

But as I suggested in my above post, maybe the mere existence of a node with the same gnx as one being filled up by the 'read' algorithm for \@file nodes in a commander could check for a difference with the already present body content.

 (thus giving a reference to be able to set a specific node as dirty, **although only for cloned nodes** that are already present in the outline.)
 
The  Good thing in that it could only do it for cloned nodes, is that it's only the cloned nodes that could want to set other \@\<file\> dirty by virtue of containing that same clone. thus it'S only the cloned node that mattered to set dirty.

Sorry for brainstorming in that comment section, but I think my idea could be implemented somehow perhaps...? Please feel free to correct my misunderstandings or misconceptions! :)

edreamleo:
@boltex Thanks for your helpful comments. Would you like to zoom about this topic?

However, I would like to merge this PR now because my testing shows that this PR works for the OP's use case in #4566:

- The `@clean` algorithm marks the (single) cloned node as dirty.
  For the OP, such cloned nodes must be *descendants* of an `@clean` node, not an `@clean` node itself.

- The code in `efc.idle_check_commander` traverses the entire outline looking for dirty nodes,
  thereby adding both cloned *positions* to the `to_do` kwarg passed to `p.v.setAllAncestorAtFileNodesDirty`.

I would like to handle any issues related to `@file` in a separate PR. Actually, I would like to ignore those issues unless somebody complains :-)

edreamleo:
@boltex Heh. There may be a better way, one that avoids `v.setAllAncestorAtFileNodesDirty` entirely.
To be clear, this PR should include the improved `v.setAllAncestorAtFileNodesDirty`.

Why not solve the problem literally at its root, in `efc.idle_check_commander`?

~~At present, this method stops when finding *any* changed external file. Instead, it could scan for *all* changed external files. Then, after calling `c.refreshFromDisk`, it could simply mark all the changed root `@<file>` nodes as changed, without ever calling `v.setAllAncestorAtFileNodesDirty`.~~

~~Job done.  What do you think?~~

All my experiments have failed.  `efc.idle_check_commander` is trickier than I guessed.

boltex:
@edreamleo I agree that the current state adresses the reported issue, and that similar behavior for \@file nodes could and should be a separate issue. 

If this is working as intended (looks like it) let go with  it then 😄

boltex (approved):
Looks good - works for @clean nodes,  and indeed @file nodes will be a separate issue/pr. :)

edreamleo:
@boltex Thanks for your approval. On second thought, rev 8802776 should ensure that all roots containing changed clones will be marked dirty. What do you think?

boltex:
@edreamleo 

yeah! ha! it's Pretty clever: it has no negative effect, and merely ensures an eventual CTRL+S will have the other external files contain the proper change(s) to their cloned parts... (if any clones were really dirty, if not, not harm done.)

Better than the current condition of reloading an external file for changes that containied a clone that should have affected other \@\<file\> nodes, and then saving thinking they now contain the change incuced by the cloned part(s), when in fact, they did not! 

I'm sold! 😁

edreamleo:
@boltex Glad you liked the fix :-)

PR #4567, opened 2026-03-23
PR: Greatly simplify c.deletePositionsInList

edreamleo:
This PR is a huge simplification of the existing code.

The code is a straightforward combination of iterative and recursive code.
We process the top-level nodes *iteratively* in reverse order, guaranteeing that positions are valid when deleted.
Processing each top-level node *recursively* deletes its children.
Aha! We remove nodes from to-be-deleted list when we delete their parent! Why did we never see this trick before?

Unless I am greatly mistaken, the algorithm runs in O(N) time, where N is the number of nodes in the tree.

- [x] Rewrite `c.deletePositionsInList`.
- [x] `c.undoableDeletePositions` is now an alias for  `c.deletePositionsInList`.

### DIFF
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -5516,7 +5516,7 @@ class Commands:
         *Undoably* delete all vnodes corresponding to the positions in aList.
         """
         c = self
-        u, undoType = c.undoer, 'deletePositionsInList'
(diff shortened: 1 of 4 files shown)


boltex (approved):
Now also part of leojs and leo-web :)

PR #4569, opened 2026-03-23
Improved 'extract' to restrict finding defs for JavaScript and TypeScript.

### LINKED ISSUE #4568: 'extract' command tries to match extractDef_patterns on all selected lines instead of the first non-empty line
Edit: See PR #4569 that addresses this issue

When doing 'extract' commands I realized that I sometimes had the wrong headline set and that the first line selected was not made into the headline, and was insead moved in as the first line of the  body instead.

I've drilled down the cause by looking at the code. This array of regex tries to deduce if the selected chunk is a function/definition of some sorts in a couple of languages, and if not, **only then does it use the first line as the headline for the extracted node content**.
```python
extractDef_patterns = (
    re.compile(r'\((?:def|defn|defui|deftype|defrecord|defonce)\s+(\S+)'),  # clojure definition
    re.compile(r'^\s*(?:def|class)\s+(\w+)'),                               # python definitions
    re.compile(r'^\bvar\s+(\w+)\s*=\s*function\b'),                         # js function
    re.compile(r'^(?:export\s)?\s*function\s+(\w+)\s*\('),                  # js function
    re.compile(r'\b(\w+)\s*:\s*function\s'),                                # js function
    re.compile(r'\.(\w+)\s*=\s*function\b'),                                # js function
    re.compile(r'(?:export\s)?\b(\w+)\s*=\s(?:=>|->)'),                     # coffeescript function
    re.compile(r'(?:export\s)?\b(\w+)\s*=\s(?:\([^)]*\))\s*(?:=>|->)'),     # coffeescript function
    re.compile(r'\b(\w+)\s*:\s(?:=>|->)'),                                  # coffeescript function
    re.compile(r'\b(\w+)\s*:\s(?:\([^)]*\))\s*(?:=>|->)'),                  # coffeescript function
)  # fmt: skip
```

The problem is that it does not only look to match this regex with the first initial (non empty ) line, **it tries with all lines selected!** so any chunk that coinicidally contains something resembling a function/class definition in some of those regex makes it have that line's detected name be the headline! (instead of the first line selected fallback)

e.g. When selecting this code and using the extract command, i expect the headline to be **xzy**, not **myFunc** because inline functions assigned to constants are a common occurence in many languages. (in this specific example, its the equal sign followed by parenthesis, etc. but others combinations from other languages happening by conincidence inside a chunk of text selected for the extract could also occur, making using the command unreliable and feeling indirectly unstable because of that '_feature_' ) 

```typescript
xyz
aaa
aaa
const myfunc = (myParameter) => {
  return blablabla + 1;
}
sdfsdf
asf
asf

```

I'll propose a pull request to fix for this (by limiting the match to the first non empty line) very soon.

boltex:
I also left a comment for further additions/improvements to the extract command.
Fixes #4568

### DIFF
--- a/leo/commands/commanderEditCommands.py
+++ b/leo/commands/commanderEditCommands.py
@@ -453,11 +453,11 @@ def extract(self: Self, event: LeoKeyEvent = None) -> None:
     h = lines[0].strip()
     ref_h = extractRef(c, h).strip()
     language = (c.getLanguage(c.p) or '').lower()
-    # Note: Adjust as necessary for other languages, maybe clearing top comments!
+    # Note: Adjust as necessary for other languages, maybe clearing top comments, etc.
     if language in ('javascript', 'typescript'):
         def_h = extractDef_find(c, [lines[0]])  # Only look at the first line for a definition.
     else:
-        def_h = extractDef_find(c, lines)  ## Default to look at all lines for a definition.
+        def_h = extractDef_find(c, lines)  # Default to look at all lines for a definition.
     #
     # First check for reference, then definition, then default to the first line.
     if ref_h:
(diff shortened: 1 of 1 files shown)


edreamleo (approved):
Interesting quirp. I think this PR is safe enough.

PR #4571, opened 2026-03-26
PR fixes #4570: Add setting to bypass file-update dialogs.

### LINKED ISSUE #4570: Make "files changed" dialogs optional
See PR #4571.

Editors such as SciTE update files changed externally automatically and silently. This seems safe enough: the user is likely to already know that this is happening.

edreamleo:
See #4570.

- [x] `efc.ask` does nothing unless a new setting is True.
- [x] Only `efc.idle_check_commander logs 'update' messages.
    The tricky part: print log messages only once!
- [x] Add `@bool raise-file-update-dialogs = False` to LeoSettings.leo.

### DIFF
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -377,7 +377,7 @@ def refreshFromDisk(
     p: Position = None,  # For compatibility with existing scripts.
     *,
     event: LeoKeyEvent = None,  # Required for all commands.
-    silent: bool = True,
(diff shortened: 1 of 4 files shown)


boltex (commented):
@edreamleo 
option for Auto-reload upon external file change detection - hurray! 😄 

I'll remove the config feature users had to activate to get it in leointeg/leojs since there's now a config setting for it  if the user want to change that default.

edreamleo:
@boltex  Excellent!

PR #4579, opened 2026-03-29
Fixed file type detection in 'ask' and updated the leoserver version of 'ask'.

boltex:
Two small fixes in leoserver and leoExternalFiles for the 'ask' method.


edreamleo (approved):
Looks good. We'll test it further in devel.

PR #4576, opened 2026-03-28
PR fixes #4572: Concrete annotations for Widgets and Wrappers

edreamleo:
See #4572. This PR will likely have no effect on LeoJS.

Replacing `Widget` and `Wrapper` by concrete annotations was remarkably easy. See the Aha below.
However, **I do not entirely trust mypy's analysis**.  Why? Because it may not take account of Leo's console and Null guis.

**Aha: rename ivars in subclasses**

An Aha that removed several roadblocks: it's sometimes (not always) possible to rename ivars in a subclass so that the annotations in the base class do not conflict with those in subclasses. In particular, added the `qt` prefix to various ivars in the Qt-related subclasses of the base classes in `leoFrame.py`.

**Documentation**

- [x] Document all breaking changes.
- [x] Update Info item #1585.

**Breaking changes**

- Delete `c.trace_idle_focus`, `c.is_unusual_focus`, and `c.trace_idle_focus`.
- Delete `k.showStateCursor`.
- `LeoQtFrame` class: Remove `configureBar` , `configureBarsFromConfig` and `reconfigureFromConfig`.
- Remove `PlainTextWrapper` and `BaseTextAPI` classes.
- Remove the `name` kwarg from `g.printStats`.

<details><summary><b>Annotate all text wrappers with QTextMixin</b></summary>
<br>

**Phase 1**

- [x] Replace `Wrapper` annotations by  `QTextEditWrapper`  in *all* files.
    Exception:  in `c.widgetWantsFocus`, annotate `w` as ` Union[QTextEditWrapper, StringTextWrapper`.
- [x] Replace `Widget` annotations by concrete annotations where possible.
    *Important*: The few remaining `Widget` annotations seem natural and reasonable.
    Why? Because none of the methods actually *uses* the widgets. Examples: 
    - `c.widget_name` just calls `g.app.gui.widget_name(widget)`.
    - `c.get_focus` just returns `g.app.gui.get_focus(c)`.
    - `AutoCompleterClass.__init__` sets `self.w` to the widget to get focus on exit.
    - etc.

** Phase 2**

- [x] Make `StringTextWrapper` class a subclass of `QTextMixin`.
- [x] A breakthrough: Annotate text wrappers as `QTextMixin` *everywhere*.
- [x] Create `g.check*` functions. Call them to check all text-related annotations.
- [x] Add two revealing unit tests: `TestNullGui.test_annotations` and `TestQtGui.test_annotations`.
[...]

PR #4581, opened 2026-03-30
PR: Generalize g.stat and related functions

edreamleo:
This PR will remain a draft until Leo 6.8.8 is released. It is a work in progress and is really part of PR #4576.

The `g.stat` function maintains `g.app.statsDict`, annotated as `dict[str, Any]`.
At present, `g.stats` inserts only ints, that is counts of how many times a function/method is executed.

This PR allows these legacy statistics, but adds the capability to list the set of other objects.
My tests call `g.stat(obj=w.__class__.__name__)` to show all the object types that an argument `w` has during execution.
These show `w`'s actual types when `w` is annotated as `Any`.

- [x] Remove `name` arg from `g.stat` and `g.printStats`.
- [x] Add a new `obj` arg to `g.stat`.
    Update the count statistic regardless of the `obj` kwarg.
- [x] All intermixed calls to `g.stat()` and `g.stat(obj=whatever)`.
- [x] `g.printStats` prints int stats followed non-int stats.

Imo, this PR is a significant addition to Leo's debugging capabilities.

### DIFF
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -2694,7 +2694,7 @@ def printStats(event: LeoKeyEvent = None, name: str = None) -> None:
 
 
 # @+node:ekr.20031218072017.3136: *4* g.stat
-def stat(name: str = None) -> None:
+def stat(name: str = None, *, obj: Any = None) -> None:
     """Increments the statistic for name in g.app.statsDict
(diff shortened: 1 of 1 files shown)


edreamleo:
I am going to close this PR in favor of PR #4576.

PR #4582, opened 2026-04-02
Boltex/issue4580

### LINKED ISSUE #4580: check-for-changed-external-files option is unaffected by reload-settings.
**Note: see P.R. #4582 for this issue.**

_I discovered this by testing out changing an external file to see if the  **check-for-changed-external-files**  and the recently added  **raise-file-update-dialogs** work well._

Unlike other options, **check-for-changed-external-files** has no effect on behavior if changed from True to False in an @setting node, saving the file and then using the 'reload-setttings' command.

It only have effect if changing it, saving the leo file, and closing, then re-opening the file.

 (no need to close Leo itself entirely, but closing and reopening the file itself makes that change take effect.)

All other leo settings I have tried can directly change the behavior of leo after changing the setting in an @setting node, and then using the 'reload-settings' command.

 But **check-for-changed-external-files** seems to require a close/reopen cycle for some reason.

boltex:
Fixes #4580

### DIFF
--- a/leo/core/LeoPyRef.leo
+++ b/leo/core/LeoPyRef.leo
@@ -121,6 +121,7 @@
 <v t="ekr.20241022090855.1"><vh>@file leoJupytext.py</vh></v>
 <v t="ekr.20190515070742.1"><vh>@file leoMarkup.py</vh></v>
 <v t="ekr.20031218072017.3320"><vh>@file leoNodes.py</vh></v>
+<v t="ekr.20040301205720"><vh>p.__str__ and p.__repr__</vh></v>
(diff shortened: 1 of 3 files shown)


boltex (commented):
I ran unit tests and tried multiple times, seems to work fine now.

edreamleo (changes requested):
Good work. I had forgotten about `c.reloadConfigurableSettings`.

I have just created #4583. It recommends pruning the table in `c.reloadConfigurableSettings`.

However, please don't change LeoPyRef in this PR.

boltex:
@edreamleo Done! :)

edreamleo (approved):

PR #4584, opened 2026-04-03
PR: Restore args to run_test_leo.py

edreamleo:
- [x] Fix reversion in `run_test_leo.py`.
- [x] Fix one unit test (and supporting code) to all `-v` and `--version` to appear on the command-line for  `run_test_leo.py`.
    Fixing this unit test was a bit of a nightmare, but all my local tests pass.


boltex (commented):
Thanks for this. It could be useful when stuck with some error when running tests  :)

edreamleo:
@boltex You're welcome :-)

PR #4586, opened 2026-04-05
Made doShowMarked uniform with the other quicksearch commands.

### LINKED ISSUE #4585: in quicksearchController, the command to find 'marked' nodes does not work
Note: See P.R. #4586 

It's the only one of those that do not work, **(alt+x and type marked-list)** and looking at the code, its the only one that's not setup like the others in that class. 

I'll try my hand at fixing it 😄 

I know there's also an easter egg that shows the marked list when 'm' is pressed alone in the nav text input box - so I'll keep an eye open not to break that !

boltex:
Made the setup of doShowMarked uniform with the other quicksearch commands.
Fixes #4585


edreamleo (approved):
Looks good!

PR #4587, opened 2026-04-09
PR fixes #4471: improve java importer

### LINKED ISSUE #4471: Improve Java Importer
Using File/Import to import a .java source file, the file gets organized into nodes that make no sense and are wrongly named with names that do not match the body.  In my test, most nodes are named like _func if_, _func catch_, and so on.  The conversion into nodes is so bad that I would be better off pasting the entire file into a node and breaking it up into nodes myself.  The outline for my test file looks like this:

```
+ class Codes
	- func getCoord
	- func if
	- func catch
	- func if
	- func if
	- func if
	+ func catch
		- func if
	+ func if
		- func if
	- func catch
	- func if
	- func if
	- func if
	- func if
	- func if
	- func catch
	+ func catch
		- func if
	- func if
```

See_ attached screen shot. 

Image

edreamleo:
See #4471.

- [x] `Importer.find_blocks` ignores regex matches that match compound statements.
- [x] Improve `Java_Importer` regexes.
- [x] Generate more informative headlines for functions.
- [x] Add and revise unit tests accordingly.
- [x] Fix a bug in the base `Importer.compute_headline` method.


edreamleo:
@tbpassin This PR is ready for review, but I won't merge it before releasing Leo 6.8.8.

The changes required were straightforward.Just fixing the regexs suffices.

Using a fancy external parser would not fit in well with Leo's existing importer code.

tbpassin:
I checked out `ekr-4471-java-importer branch, build 6ca5393e3e` and imported a java source file. It all imported into a single node. That isn't intended, is it?

tbpassin:
> Using a fancy external parser would not fit in well with Leo's existing importer code.

It requires subclassing some of the importer classes. The main trouble I've had using a real java parser is determining where blocks end. The parser does a fine job of finding where they start. My current version can split out one level of indentation but inner classes don't go into their own subtrees.

Sorry I've not finished the work but I'm overloaded with a couple of other projects.

edreamleo:
@tbpassin Thanks for testing. Please send me a test file so I can improve the importer as needed.

tbpassin:
Here is a .java file that does get decomposed but some of the node names are wildly wrong, and in some cases the last line or few lines from one function are put into the following node.
[Autotest.java](https://github.com/user-attachments/files/26649469/Autotest.java)

edreamleo:
@tbpassin Thanks for the test file. I'll make it work.

edreamleo:

I tried many regex patterns. Imo, the present patterns are simplest and produce the best results.

edreamleo:
@tbpassin Please review and test this PR.

tbpassin:
Better but not there yet.
Good:
1. Detects Java doc comment block and includes it ahead of the class declaration line;
2. Splits the outer class into a node.

Not so good:
1. Includes the package and import lines at the start of each file in the "Class"'s node.
2. Although the one test file was split up into nodes per method, that didn't happen for some other files I tested with. E.g., for the file included below, the `getCounts()` method was included in the node for `main()` even though they are siblings.

[SplitTest.java](https://github.com/user-attachments/files/26717374/SplitTest.java)

edreamleo:
@tbpassin Thanks for your comments. I'll see what I can do.

edreamleo:
@tbpassin Alright, I bit the bullet and overrode three methods so as to simplify the regexes.

- The new `java_i.find_blocks` ignores regex matches that match compound statements.
- The new `java_i.move_module_preamble` comes from the Python importer.

What do you think?

edreamleo:
@tbpassin I found an easier way: The base `Importer.find_blocks` ignores regex matches that match compound statements.

All subclasses need to do is to override their `compound_statements` list.

tbpassin:
Here are two more .java test files. One of them has two levels of inner classes. They imported all right with this version of the importer:
```
Leo 6.8.9-devel, ekr-4471-java-importer branch, build 9290bc5f28
2026-04-16 06:01:40 -0500
Python 3.13.3, PyQt version 6.9.0
Windows 11 AMD64 (build 10.0.26200) SP0
```
[Innerclass.java](https://github.com/user-attachments/files/26785356/Innerclass.java)
[interface1.java](https://github.com/user-attachments/files/26785358/interface1.java)

One of the files only contains interface definitions. They did not get separated into nodes. I think that is all right because these interface definitions are almost always one-liners. These files, especially the one with inner-inner classes, should be a good test file.

One thing I haven't checked is to make sure that the importer works right no matter whether the opening brace is on the same line or the next one:

```
class C1 {
   //..
}

class C2
{
   //..
}
```

edreamleo:
@tbpassin Thanks for your comment. I'll look into using re.MULTILINE for the patterns.

Actually, re.MULTILINE isn't the right flag, but re.DOTALL isn't going to work either.

edreamleo:
@tbpassin I did a little research:

The base `Importer.find_blocks` applies the block patterns only to a single line, so there is no way to match across lines without significant changes to the existing code. I'd rather leave such changes to a separate PR.

All your test files now import reasonably, imo. Please approve this PR.

tbpassin (approved):
Remember another issue should be raised about having the opening "{" on a following line.

edreamleo:

#4610 suggests supporting functions whose signatures span multiple lines.

PR #4588, opened 2026-04-12
Fix string concatenation in abbreviation toggle message

boltex:
Quick fix for the toggle-abbrev-mode command's output message which was inconsistent.


edreamleo (approved):
Thanks. Good catch.

PR #4597, opened 2026-04-15
(Defensive) Fix potential bug: Improved .leojs file format support for AbbrevCommandsClass 'reload-settings'.

boltex:
Improved .leojs file format (json) support. 

Fixed a potential error message in the log pane, that should not have happenned unless the settings were set to have clipboard in json, AND reloading settings with abbreviations turned on.


edreamleo (approved):
Looks good.

PR #4601, opened 2026-04-15
PR: Fix copy paste

edreamleo:
Fixes PR #4576. `git-bisect` found the culprit: rev cf2d872.

- [x] Set `self.qtLogCtrl` in `LeoQtLog.createTab` and `LeoQtLog.selectTab`.

**Other changes**

- [x] Merge PR #4600.
- [x] Improve `LeoKeyEvent.__repr__`.

**Discussion**

This bug arose from renaming the ivars of the `LeoQtLog` class, and particularly the `qtLogCtrl` ivar. This renaming was a breakthrough, not a mistake. The renaming helped mypy annotate wrappers more concretely.

From the first, I was aware that this renaming could be dangerous&mdash;the distinction between free and bound variables is crucial in mathematical logic. I simply missed a few places where I had to copy (bound) values from the subclass to the base class.

### DIFF
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -129,11 +129,14 @@ def restartLeo(self: Self, event: LeoKeyEvent = None) -> None:
     sys.stdout.flush()
     sys.stderr.flush()
     # Restart Leo with subprocess.run.
-    g.cls()
-    print('')
+    if verbose:
+        print('')
+    else:
+        g.cls()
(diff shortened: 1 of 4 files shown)


edreamleo:

boltex (commented):
dont know enough about python and qt to understand why this fixed it , but I tried it and it works 😄

edreamleo:
@boltex Thanks for the testing. The cause of the bug was indeed subtle.

Issue #1510, opened 2020-02-25
Invalid character can ruin a .leo file


edreamleo:
A `0x01` character in the .leo file produces the traceback given below.
```
bad .leo file: WORK.leo
g.toUnicode: unexpected argument of type ParseError
toUnicode openLeoFile,getLeoFile,readFile,readWithElementTree


Traceback (most recent call last):

  File "E:\git\leo\leo\core\leoCommands.py", line 2278, in executeAnyCommand
    return command(event)

  File "E:\git\leo\leo\core\leoGlobals.py", line 240, in commander_command_wrapper
    method(event=event)

  File "E:\git\leo\leo\commands\commanderFileCommands.py", line 261, in open_outline
    open_completer(c, closeFlag, fileName)

  File "E:\git\leo\leo\commands\commanderFileCommands.py", line 204, in open_completer
    c2 = g.openWithFileName(fileName, old_c=c)

  File "E:\git\leo\leo\core\leoGlobals.py", line 3920, in openWithFileName
    return g.app.loadManager.loadLocalFile(fileName, gui, old_c)

  File "E:\git\leo\leo\core\leoApp.py", line 3288, in loadLocalFile
    previousSettings = lm.getPreviousSettings(fn)

  File "E:\git\leo\leo\core\leoApp.py", line 2241, in getPreviousSettings
    c = lm.openSettingsFile(fn)

  File "E:\git\leo\leo\core\leoApp.py", line 2412, in openSettingsFile
    ok = c.fileCommands.openLeoFile(theFile, fn,

  File "E:\git\leo\leo\core\leoFileCommands.py", line 692, in openLeoFile
    ok, ratio = self.getLeoFile(

  File "E:\git\leo\leo\core\leoFileCommands.py", line 559, in getLeoFile
    v = FastRead(c, self.gnxDict).readFile(fileName)

  File "E:\git\leo\leo\core\leoFileCommands.py", line 60, in readFile
    v, g_element = self.readWithElementTree(path, s)

TypeError: cannot unpack non-iterable NoneType object

Traceback (most recent call last):
  File "E:\git\leo\leo\plugins\qt_frame.py", line 1638, in closeEvent
    g.app.gui.close_event(event)
AttributeError: 'NullGui' object has no attribute 'close_event'
```

tbpassin:
"A 0x01 character in the .leo file produces the traceback given below. There are two possible fixes:

    Prohibit bad characters when writing .leo files.
[...]

PR #4615, opened 2026-04-19
Quick perf pass on core: ~3x faster cold open via leoBridge

vivainio:
## Caveat

This is a quick exploratory pass done with Claude Code — ~1 hour of profiling and patching, not a deep study. Each commit is small and narrowly scoped so you can cherry-pick or reject them individually. All three are measured wins on cold open via `leoBridge`, but wider testing (GUI paths, other `.leo` files, Windows) would be worth doing before merge.

## Measurements

Profiled `leoBridge.controller().openLeoFile('leo/doc/LeoDocs.leo')` (1.6 MB) on Linux / Python 3.12:

| stage | total cold open | delta |
|---|---|---|
| baseline (devel @ 18084b174) | 1784 ms | — |
| + 8114aa7f4 (leoMarkup lazy) | 808 ms | −55% |
| + e002125bf (XML translate skip) | 620 ms | −23% |
| + d66032336 (importer startswith) | 544 ms | −12% |
| **all three** | **544 ms** | **−69%** |

`cProfile` was used to attribute time; numbers are `time.perf_counter()` wall-clock, nullGui, `loadPlugins=False`, `readSettings=False`.

### 8114aa7f4 — `leoMarkup: defer external tool discovery to first use`

`leoMarkup.py` ran four `shutil.which()` calls and spawned `rvm gemdir` at **module import time**. Because `Commands.__init__` imports `leoMarkup`, this fired on every `.leo` file open (twice — settings + file), spending ~800 ms walking PATH and fork/exec'ing a Ruby tool most users don't have.

- Wrapped each lookup in `functools.cache`d helpers; resolution now happens only when an `@adoc`/`@pandoc`/`@sphinx` command is actually invoked.
- Dropped the RVM probe entirely — nothing in-repo depends on it, and the `try: subprocess.check_output(['rvm', 'gemdir'])` fails for any user without Ruby Version Manager, wasting a fork.

Plugin copies (`viewrendered.py`, `viewrendered3.py`) have their own eager `which()` calls at import time but those modules only load when a plugin is activated, so they're not on the core hot path. Left untouched.

### e002125bf — `leoFileCommands: skip XML control-char strip on clean files`

`readWithElementTree` unconditionally ran `str.translate` over the entire `.leo` file contents to strip XML-invalid control chars (issue #1510), costing ~100 ms per MB of file. Virtually no real `.leo` file contains those chars.

- Added a compiled regex pre-check (`bad_xml_chars_re`) that scans for any affected code point. On LeoDocs.leo the check runs in ~3 ms vs ~115 ms for `translate`.
- Behavior is unchanged: when bad chars are present the same `translate` still runs.

### d66032336 — `base_importer: use str.startswith instead of g.match in inner loop`

`delete_comments_and_strings` runs for every line of every `@file` body on load. It called `g.match(line, i, pattern)` inside a per-character loop. `g.match` is a pure-Python wrapper around `str.find`; `str.startswith(pattern, i)` does the same work in C and accepts a **tuple** of alternatives, so `any(g.match(line, i, z) for z in string_delims)` collapses to one native call.

- Replaced four `g.match` call sites.
- Removed ~160k Python-level calls per LeoDocs load.
- Verified behavior equivalence: `string_list` entries are always non-empty (or the whole list is empty, e.g. rust), and the other call sites (`target`, `start_comment`) are already truthiness-guarded — so the `str.startswith('')=True` edge case cannot be hit.

## What I did not touch (but noticed)

- `leoCommands.initObjects` still runs two subcommander wire-ups per open (settings + file). Possibly skippable on the settings pass but a bigger refactor.
- `c.getPathFromNode` is called ~10k times during a full-tree walk; memoization per-VNode for the duration of a load would help but needs care around edits.
- `viewrendered*.py` have the same eager `which()` pattern as pre-fix `leoMarkup.py`.

## Test plan

- [x] Manual: open LeoDocs.leo, leoSettings.leo, and a project `.leo` in the GUI; verify no behavior change.
- [ ] `@adoc` / `@pandoc` / `@sphinx` commands still work when those tools are installed (I can't easily verify — no asciidoctor/pandoc on this machine).
- [x] CI test suite.
- [ ] A `.leo` file actually containing stripped control chars exercises the `bad_xml_chars_re` branch in `readWithElementTree`.

### DIFF
--- a/leo/core/leoFileCommands.py
+++ b/leo/core/leoFileCommands.py
@@ -15,6 +15,7 @@ import io
 import json
 import os
 import pickle
+import re
 import shutil
 import sqlite3
 import tempfile
@@ -171,6 +172,9 @@ class FastRead:
     # @+node:ekr.20180602062323.7: *3* fast.readWithElementTree & helpers
     # #1510: https://en.wikipedia.org/wiki/Valid_characters_in_XML.
     translate_dict = {z: None for z in range(20) if chr(z) not in '\t\r\n'}
+    # Match any control char that must be stripped per #1510. Used as a cheap
+    # pre-check so we skip `str.translate` (~100 ms/MB) on clean files.
+    bad_xml_chars_re = re.compile('[' + ''.join(chr(z) for z in translate_dict) + ']')
 
     bad_path_dict: dict[str, bool] = {}
 
@@ -180,8 +184,9 @@ class FastRead:
         s_or_b: str | bytes,
     ) -> tuple[VNode, Value]:
         contents = g.toUnicode(s_or_b)
-        table = contents.maketrans(self.translate_dict)  # type:ignore #1510.
-        contents = contents.translate(table)  # #1036, #1046.
+        if self.bad_xml_chars_re.search(contents):  # #1036, #1046, #1510.
+            table = contents.maketrans(self.translate_dict)  # type:ignore
+            contents = contents.translate(table)
         try:
             xroot = ElementTree.fromstring(contents)
         except Exception:
--- a/leo/core/leoFileCommands.py
+++ b/leo/core/leoFileCommands.py
@@ -170,22 +170,16 @@ class FastRead:
         return hidden_v
 
     # @+node:ekr.20180602062323.7: *3* fast.readWithElementTree & helpers
+    # PR #4615: Don't use str.translate unless there are bad characters.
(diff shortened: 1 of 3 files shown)


edreamleo (approved):
@vivainio Many thanks for another excellent PR.

edreamleo:
@vivainio Ok. We'll assume that the unfinished tests won't cause problems, hehe.

PR #4617, opened 2026-04-19
Colorizer: fix accidentally-quadratic jupytext check in python_comment (4x speedup)

vivainio:
## The bug

`leo/modes/python.py::python_comment` is called by the syntax highlighter for **every `#` character** that starts a comment, on every line of every Python node. Since Leo 6.8.3 it does this:

```python
# Always colorize the comment line as a *python* comment.
n = colorer.match_eol_span(s, i, kind=\"comment1\", seq=\"#\")

# Leo 6.8.3. Add special case for @language jupytext.
in_jupytext_tree = any(
    z.startswith('@language jupytext')
    for z_p in c.p.self_and_parents()
    for z in g.splitLines(z_p.b)
)
is_any_jupytext_comment = i == 0 and s.startswith('# %%') and in_jupytext_tree
```

The generator walks **every ancestor**, splits **every ancestor body** into lines, and checks **every line** for `@language jupytext`. Because `any()` only short-circuits when it *finds* a match, in the overwhelmingly common case (no ancestor is jupytext), every line of every ancestor gets scanned.

And this runs even when the cheap, gating checks — `i == 0` and `s.startswith('# %%')` — haven't been satisfied. Virtually no Python comment starts with `# %%`, so virtually all of this work is wasted.

## Fix

Move the cheap checks in front of the ancestor scan:

```python
if not (i == 0 and s.startswith('# %%')):
    return n
in_jupytext_tree = any(...)
if in_jupytext_tree:
    ...
```

Semantics are unchanged. When the line really is a `# %%` marker, the ancestor scan runs exactly as before. Otherwise we return with the already-applied comment coloring, just as the old code would have (since `is_any_jupytext_comment` would be False).

## Measurements

Driven headlessly: `JEditColorizer(c, None).mainLoop(...)` over every line of `leo/core/leoGlobals.py` (~278 KB, 8716 lines, ~1408 `#` comments). Python 3.12, Linux.

| | wall time | `str.startswith` calls | `python_comment` cumtime |
|---|---|---|---|
| before | 1511 ms | 12,303,336 | 4.08 s (under cProfile) |
| after | 367 ms | ~0 inside `python_comment` | not in top 25 |

**~4.1× faster** on this file. The ratio scales with (number of `#` comments) × (total lines in ancestor bodies), so bigger outlines or files with more comments benefit more.

## Caveat

Quick fix found while profiling the colorizer with Claude Code. `leo/unittests/core/test_leoColorizer.py` (38 tests) still passes. There are no dedicated jupytext tests, but manual verification of a `@language jupytext` tree with `# %%` / `# %% [markdown]` markers would be worth adding before merge if you want belt-and-braces.

## Test plan

- [x] `leo/unittests/core/test_leoColorizer.py` — 38 passed
- [x] Manual: open a jupytext node with `# %% [markdown]` cells, verify markdown coloring still kicks in.
- [x] Manual: confirm non-jupytext Python files look identical after the change.

### DIFF
--- a/leo/modes/python.py
+++ b/leo/modes/python.py
@@ -336,13 +336,13 @@ def python_comment(colorer, s, i):
     n = colorer.match_eol_span(s, i, kind="comment1", seq="#")
 
     # Leo 6.8.3. Add special case for @language jupytext.
-    # Fast path: virtually every '#' comment fails the cheap checks, so
-    # gate the ancestor/body scan behind them. Without this gate, the
-    # scan fired on every '#' in every Python node, turning colorization
-    # of any large @file node into an O(n_comments * n_ancestor_lines)
-    # operation.
+
+    # PR #4617 (Ville): https://github.com/leo-editor/leo-editor/pull/4617
+    # Make this fast test first.
     if not (i == 0 and s.startswith('# %%')):
         return n
+
+    # This test is expensive!
     in_jupytext_tree = any(
         z.startswith('@language jupytext')
         for z_p in c.p.self_and_parents()
--- a/leo/modes/python.py
+++ b/leo/modes/python.py
@@ -336,13 +336,19 @@ def python_comment(colorer, s, i):
     n = colorer.match_eol_span(s, i, kind="comment1", seq="#")
 
     # Leo 6.8.3. Add special case for @language jupytext.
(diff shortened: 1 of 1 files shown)


edreamleo (approved):
Again, many thanks for this superb PR.

PR #4618, opened 2026-04-19
Colorizer: skip setTag debug trace by default (~65% less time in setTag)

vivainio:
## The bug

`leoColorizer.JEditColorizer.setTag` runs a nested `report()` function **on every tag-set**, to maintain `self.last_trace` — the buffer consumed by the `dump-last-colorizer-trace` diagnostic command:

```python
def report(color: QtGui.QColor) -> None:
    \"\"\"A superb trace. Don't remove it.\"\"\"
    i_j_s = f\"{i:>3}:{j:<3}\"
    matcher_name = g.caller(3)
    rule_name = g.caller(4)
    matcher_s = f\"{self.rulesetName}::{rule_name}:{matcher_name}\"
    s2 = s[i:j]
    state = self.currentState()
    state_repr = self.stateNumberToStateString(state)
    state_s = f\"{self.currentState()}={state_repr}\"
    trace_line = (
        f\"{self.recolorCount:5} {self.currentBlockNumber():<4} {state_s:<25}\"
        f\"{matcher_s:<55} {colorName:7} {full_tag:<20} {i_j_s} {s2}\"
    )
    ...
    self.last_trace.append(trace_line)
    if trace:
        print(trace_line)
```

Per call it walks the stack twice via `g.caller` (which internally calls `g.callers` → `_callerName`), queries colorizer state, and builds a six-field f-string. On a large Python node (`leoGlobals.py`, 8716 lines, ~14 600 tag-sets per colorize pass) this accounts for roughly 40% of total colorize time — for a diagnostic the user is not currently running.

## Fix

`setTag` already has a `trace` flag at the top:

```python
trace = 'coloring' in g.app.debug and not g.unitTesting
```

That flag toggles whether the live trace *prints*. Reuse it to also toggle whether the trace is *recorded*:

```python
self.tagCount += 1
if trace:
    report(color)
self.highlighter.setFormat(i, j - i, format)
```

## Behavior change

Minor and opt-in. Previously, `dump-last-colorizer-trace` always showed the last recolor's tagging timeline. After this change, the user must set `g.app.debug = ['coloring']` before the recolor they want to inspect. The live trace print already required exactly this flag, so the flag is not new — it simply now also gates recording.

If you prefer zero behavior change, the alternative is to store raw tuples in `last_trace_data` and format them lazily in the dump command. Happy to switch to that approach. It's more code churn but the user-visible feature stays always-on.

## Measurements

Driven headlessly: `JEditColorizer(c, None).mainLoop(...)` over every line of `leo/core/leoGlobals.py` (~278 KB, 8716 lines) on Python 3.12 / Linux.

| scenario | mainLoop wall | setTag cumtime (cProfile) |
|---|---|---|
| devel baseline | 1438.6 ms | 624 ms |
| this PR alone | 1403.6 ms | 216 ms |
| PR #4617 alone (jupytext fix) | 366.7 ms | 598 ms |
| this PR + #4617 | **147.9 ms** | 169 ms |

Notes on the numbers:
- **setTag cumtime drops ~65%** (624 → 216 ms) from this change alone.
- Wall time drops only modestly (1438 → 1403 ms) when measured on devel, because the jupytext bug in `python_comment` (see #4617) dominates wall time. With #4617 landed first, this PR cuts colorize wall time from 367 ms → 148 ms — **~2.5× faster** on top.
- Stacked with #4617 the combined improvement on this file is **~10× over devel baseline** (1438 ms → 148 ms).
- Bigger outlines or files with more tags scale linearly; the ratio stays constant.

## Test plan

- [x] `leo/unittests/core/test_leoColorizer.py` — 38 passed.
- [x] Manual: confirm `dump-last-colorizer-trace` still works after setting `g.app.debug = ['coloring']`.
- [x] Manual: confirm the trace print (which was already gated on this flag) behaves identically.

## Caveat

Quick pass with Claude Code. One tiny guard; very easy to revert if the behavior change is unwelcome.

### DIFF
--- a/leo/core/leoColorizer.py
+++ b/leo/core/leoColorizer.py
@@ -1550,11 +1550,9 @@ class JEditColorizer(BaseColorizer):
             format.setForeground(color)
             format.setUnderlineStyle(UnderlineStyle.NoUnderline)
         self.tagCount += 1
-        # Building the trace costs ~40% of colorize time (g.caller stack walks
-        # and f-string formatting, repeated per tag). Gate it behind the same
-        # debug flag that toggles the live trace print; users who want to run
-        # `dump-last-colorizer-trace` must set `g.app.debug = ['coloring']`.
         if trace:
+            # PR #4618: (Ville Vainio) https://github.com/leo-editor/leo-editor/pull/4618
+            # Don't call report by default: It's setup is expensive!
             report(color)
         self.highlighter.setFormat(i, j - i, format)
 
--- a/leo/core/leoColorizer.py
+++ b/leo/core/leoColorizer.py
@@ -1550,7 +1550,12 @@ class JEditColorizer(BaseColorizer):
             format.setForeground(color)
             format.setUnderlineStyle(UnderlineStyle.NoUnderline)
         self.tagCount += 1
-        report(color)  # A superb trace, cached for the dump-last-colorizer-trace command.
+        # Building the trace costs ~40% of colorize time (g.caller stack walks
+        # and f-string formatting, repeated per tag). Gate it behind the same
+        # debug flag that toggles the live trace print; users who want to run
+        # `dump-last-colorizer-trace` must set `g.app.debug = ['coloring']`.
+        if trace:
+            report(color)
         self.highlighter.setFormat(i, j - i, format)
 
     # @+node:ekr.20110605121601.18589: *3* jedit:Pattern matchers


edreamleo (approved):
Many thanks for this excellent PR.

I shortened the comment and added a link to this PR.

My by-hand tests look good.

PR #4620, opened 2026-04-19
Add GitHub Actions CI: ruff lint on push/PR to devel

vivainio:
## Summary
- Adds `.github/workflows/ci.yml` running `ruff check leo` on every push and PR to `devel`.
- Matrix: Python 3.10, 3.11, 3.12 on `ubuntu-latest`.
- Uses the existing ruff configuration already in `pyproject.toml` — no new config, no behavior change to the repo.

## Why
Leo already has ruff, mypy, flake8, pylint, and pytest configured locally, but nothing gates PRs. This is the minimal first step — a fast, zero-dependency lint gate that will be green on day one (verified locally: `ruff check leo` → *All checks passed!*).

Follow-ups (deliberately out of scope here):
- mypy gate (262 errors today, mostly transitive from `leo/plugins/qt_gui.py`; only ~26 inside `leo/core/` — worth a dedicated cleanup PR).
- pytest gate (needs PyQt6 system libs + `QT_QPA_PLATFORM=offscreen`).
- pre-commit config.

### Optional follow-up: release workflow
If there's interest, I can add a separate `release.yml` that publishes `leo` to PyPI automatically when a GitHub Release is created (using `pypa/gh-action-pypi-publish` with OIDC trusted publishing — no long-lived API token needed). Happy to do that as a follow-up PR if maintainers want it.

## Test plan
- [x] CI runs on this PR and all three Python versions pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,23 @@
+name: CI
+
+on:
+  push:
+    branches: [devel]
+  pull_request:
+    branches: [devel]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.10", "3.11", "3.12"]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: pip
+      - run: pip install ruff
+      - run: ruff check leo
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
     steps:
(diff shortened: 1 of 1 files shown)


edreamleo:
@vivainio I am concerned about the closed PR #4621. I don't get any mypy errors ever. Would this PR cause mypy to start complaining?

edreamleo:
@vivainio Also, shouldn't Python 3.13 be included here? `python-version: ["3.10", "3.11", "3.12"]`?

vivainio:
@edreamleo This PR only runs `ruff check` — no mypy is invoked, so it can't cause new mypy complaints. We can add a mypy job later if we want.

Feel free to merge this whenever you like.

edreamleo (approved):
Alright. Let's try this workflow and see how it works.

vivainio:
@edreamleo seems to work fine. Just say the word and I can spin up the release workflow as well, removing the need to upload release packages to github releases manually

edreamleo:
@vivainio Alright. Let's try it. You're pulling me into the modern world :-)

vivainio:
Follow-up: opened #4651 with the `publish.yml` workflow (GitHub Release assets + PyPI trusted publishing) we discussed here.

PR #4621, opened 2026-04-19
Fix all mypy errors on leo/core import surface (QWidget types + small bug)

vivainio:
## Summary
Drives `mypy --config-file .mypy.ini leo/core` from **270 errors → 0**.

Nearly all errors were the same pattern: Qt classes (`QWidget`, `QDialog`, `QFrame`, etc.) used as type annotations but mypy couldn't recognize them as types, because they were assigned as plain variables inside `TYPE_CHECKING` blocks:

```python
QWidget = QtWidgets.QWidget   # mypy: "Variable ... is not valid as a type"
```

This PR converts those to proper type aliases (the `TypeAlias` pattern was already used for a handful of entries — this generalizes it):

```python
QWidget: TypeAlias = QtWidgets.QWidget
```

## Scope
- **`leo/core/`** — leoColorizer, leoKeys, leoVim: replaced the value assignment with a direct `from PyQt6.QtWidgets import QWidget` under `TYPE_CHECKING`. leoQt: added `# type: ignore[no-redef]` to the optional-Qt-module `try: from PyQt6 import X` imports, which redefine names declared by the `if 0:` testing-toggle block at the top of the file.
- **`leo/plugins/`** — qt_commands, qt_frame, qt_gui, qt_layout, qt_text, qt_tree, viewrendered: same `: TypeAlias` treatment. `qt_layout.py` had the assignments at module level; swapped them for direct `from PyQt6.QtWidgets import ...`.
- **`leo/commands/editCommands.py`** — annotated `self.moveSpot` and `self.moveCol` as `int | None` so that `setMoveCol` can assign `int` values to them (mypy had inferred the attributes' type as `None` from the `__init__` default).

Also fixed a latent typo in `qt_frame.py`:
```python
QLayout = QtWidgets.QWidget   # was wrong
QLayout: TypeAlias = QtWidgets.QLayout
```

## Results
- `python -m mypy --config-file .mypy.ini leo/core` → *Success: no issues found in 48 source files*
- `ruff check leo` → *All checks passed!*

## Test plan
- [ ] Re-run both commands in CI (see companion PR #4620 for the ruff gate).
- [ ] Smoke-test Leo launches and core editing works — all changes are in `TYPE_CHECKING` blocks, type-only annotations, or the single intentional `QLayout` typo fix; no runtime behavior change expected.


edreamleo:
I am going to close this PR for now.  Leo supports Python 3.10 and above. Python 3.12 added TypeAlias.

We can consider reopening this PR in a few years, when Leo requires Python 3.12 or above.

PR #4622, opened 2026-04-19
Replace implicit Optional parameters with explicit T | None

vivainio:
## Summary
Applies the [`no_implicit_optional`](https://pypi.org/project/no-implicit-optional/) codemod (`--use-union-or`) to `leo/core`, `leo/commands`, and `leo/plugins`. Rewrites **71 files**, purely mechanically:

\`\`\`python
# before
def foo(x: str = None, y: Callable = None) -> None: ...

# after
def foo(x: str | None = None, y: Callable | None = None) -> None: ...
\`\`\`

## Why
PEP 484 banned implicit Optional in 2021 (mypy's default is now \`no_implicit_optional = True\`). The codebase has accumulated ~950 call signatures in the old form. These are the single largest category of issues masked by Leo's \`.mypy.ini\` setting \`strict_optional = False\`. Fixing them mechanically now:

1. Makes annotations honest about nullability (closer to what the code actually accepts).
2. Unblocks future per-file \`strict_optional\` work — the ~950 implicit-Optional errors would otherwise dominate any strict run.
3. Aligns with mypy's own default and with the style used in new code already in the repo.

## Scope
- `leo/core`: 34 files
- `leo/commands`, `leo/plugins`: 37 files
- Total: 71 files, 1085 / 1085 line changes (symmetric — pure rewrite, no structural changes).

## What this does NOT do
- Does **not** enable \`strict_optional\` in \`.mypy.ini\` — that's a separate follow-up PR. This PR is strictly an annotation cleanup; current mypy gates keep the same error count (262, all pre-existing).
- Does **not** touch the 2 unrelated editCommands.py assignment errors.

## Test plan
- [x] \`ruff check leo\` — clean
- [x] \`python -m mypy --config-file .mypy.ini leo/core\` — same 262 errors as on \`devel\` (no regression)
- [ ] Smoke test: Leo launches and basic editing works (no runtime code paths change — only annotations on parameter defaults)

## How to reproduce
\`\`\`bash
pip install no_implicit_optional
no_implicit_optional --use-union-or leo/core leo/commands leo/plugins
\`\`\`

### DIFF
--- a/leo/commands/abbrevCommands.py
+++ b/leo/commands/abbrevCommands.py
@@ -638,7 +638,7 @@ class AbbrevCommandsClass(BaseEditCommandsClass):
     # @+node:ekr.20150514043850.19: *3* abbrev.dynamic abbreviation...
     # @+node:ekr.20150514043850.20: *4* abbrev.dynamicCompletion C-M-/
     @cmd('dabbrev-completion')
-    def dynamicCompletion(self, event: LeoKeyEvent = None) -> None:
+    def dynamicCompletion(self, event: LeoKeyEvent | None = None) -> None:
         """
         dabbrev-completion
         Insert the common prefix of all dynamic abbrev's matching the present word.
@@ -673,7 +673,7 @@ class AbbrevCommandsClass(BaseEditCommandsClass):
 
     # @+node:ekr.20150514043850.21: *4* abbrev.dynamicExpansion M-/ & helper
     @cmd('dabbrev-expands')
-    def dynamicExpansion(self, event: LeoKeyEvent = None) -> None:
+    def dynamicExpansion(self, event: LeoKeyEvent | None = None) -> None:
         """
         dabbrev-expands (M-/ in Emacs).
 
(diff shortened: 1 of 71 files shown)


edreamleo:
I am going to close this PR for the following reasons:

1. The additional syntax *reduces* the signal/noise ratio with providing any obvious benefits.  Compare:
```python
def dynamicCompletion(self, event: LeoKeyEvent = None) -> None:
def dynamicCompletion(self, event: LeoKeyEvent | None = None) -> None:
```
Imo, [PEP 484](https://peps.python.org/pep-0484/) is misguided in this regard. If anything, mypy should be extended so that it treats the two signatures above as exactly equivalent. *Note* that I am **not** saying that the second annotation is always correct, but it should be the default *implied* annotation.

2. The following assertion is false: "PEP 484 banned implicit Optional in 2021". Here is what the PEP actually says:

QQQQQ
A past version of this PEP allowed type checkers to assume an optional type when the default value is None, as in this code:
```python
def handle_employee(e: Employee = None): ...
```
This would have been treated as equivalent to:
```python
def handle_employee(e: Optional[Employee] = None) -> None: ...
```
This is no longer the recommended behavior. Type checkers should move towards requiring the optional type to be made explicit.
QQQQQ

**Summary**

Saying that "implicit Optional is no longer the recommended behavior" is emphatically *not* the same as banning implicit Optional!

PR #4628, opened 2026-04-21
Defer jupytext import to fix ~0.7s startup regression

vivainio:
## Problem

`jupytext` (when installed) calls `pandoc_version()` at module load time,
spawning 3 subprocesses to check the pandoc version. Because `leoApp.py`
imported `leoJupytext` at the top level, this ran on every Leo startup —
costing ~0.7s (26% of total startup time) even when jupytext is never used.

Profiled with `python -m cProfile launchLeo.py --gui=null --quit`:
- **Before:** 2.7s, with `_winapi.CreateProcess` (pandoc subprocesses) = 0.709s
- **After:** 2.0s — pandoc subprocess calls eliminated from startup path

## Fix

**`leoJupytext.py`** — replace the eagerly-evaluated `try: import jupytext`
block with a `_ensure_jupytext()` lazy loader. The `full_path()` guard (which
all jupytext operations go through) now calls `_ensure_jupytext()` instead of
checking `has_jupytext`.

**`leoApp.py`** — move `from leo.core.leoJupytext import JupytextManager`
under `TYPE_CHECKING`. Safe because `from __future__ import annotations` is
already present (PEP 563).

### DIFF
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -18,10 +18,10 @@ import platform
 from leo.core import leoGlobals as g
 from leo.core import leoExternalFiles
 from leo.core.leoCache import GlobalCacher
-from leo.core.leoJupytext import JupytextManager
 from leo.core.leoQt import QCloseEvent
 
 if TYPE_CHECKING:
+    from leo.core.leoJupytext import JupytextManager
     StringIO = io.StringIO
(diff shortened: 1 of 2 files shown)


edreamleo (approved):
Excellent work. Many thanks for your help.

PR #4633, opened 2026-04-23
goto-prev-marked now tries to loop going up from bottom

### LINKED ISSUE #4632: goto-prev-marked does not loop around like goto-next-marked.
See P.R. #4633 which fixes this issue 😄 

Unlike **goto-next-marked**., In an outline with some marked nodes, **goto-prev-marked** prints 'done' instead of trying to loop around if no marked nodes above the current selection.

boltex:
instead of up from the top! :)

Fixes #4632

### DIFF
--- a/leo/commands/commanderOutlineCommands.py
+++ b/leo/commands/commanderOutlineCommands.py
@@ -978,9 +978,7 @@ def goToPrevMarkedHeadline(self: Cmdr, event: LeoKeyEvent = None) -> None:
             break
         else:
             wrapped = True
-            p = c.rootPosition() # Get last node with while-loop below.
-            while p and p.hasThreadNext():
-                p.moveToThreadNext()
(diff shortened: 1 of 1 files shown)


edreamleo (changes requested):
Good catch.  However, please replace:
```python
 p = c.rootPosition() # Get last node with while-loop below.
 while p and p.hasThreadNext():
     p.moveToThreadNext()
```
with
```python
p = c.lastPosition()
```
The latter is significantly faster and more descriptive.

boltex:
@edreamleo no problem - thanks for pointing out the proper method :)

(i had reproduced the logic of goto-last-node because was looking in commanderOutlineCommands.py for such a method instead of directly in leoCommands.py! haha!)

edreamleo (approved):

PR #4635, opened 2026-04-24
Fixed 'expand_text' method to not restore the insert point when selecting next placeholder that was found..

### LINKED ISSUE #4604: Abbreviations: Having scripting enabled breaks the 'double comma' goto next placeholder.
I tried scripting enabled to see the effect of the abbreviation date;; which replaces with the actual date. I've discovered a bug with another abbreviation feature: double comma to select the next placeholder.

Without scripting enabled, using double comma selects all the text of the next placeholder properly:

example, doing abbreviation `html;;` selects **title** in this output, and doing double comma selects **content**: 
```
<html>
<head>
<title>title</title>
<style>
</style>
</head>
<body>
<|content|>
</body>
</html>
```

But, if I set abbreviation scripting on with **@bool scripting-abbreviations = True**, and  then reload-settings command, then doing the same abbreviations place the cursor at the end, instead of the word **title**

Also has erratic behavior when doing double comma if the next placeholders are in other descendant node's body. (seems ok to select properly in headlines. )

(I discovered this while testing out #4598 which seems unrelated altough being in the same source file.)

boltex:
Fixes #4604

### DIFF
--- a/leo/commands/abbrevCommands.py
+++ b/leo/commands/abbrevCommands.py
@@ -346,7 +346,10 @@ class AbbrevCommandsClass(BaseEditCommandsClass):
                     return
                 p.moveToThreadNext()
         else:
-            self.find_place_holder(p, do_placeholder)
+            if self.find_place_holder(p, do_placeholder):
+                # Don't restore the insert point when selecting next placeholder.
+                self.save_ins = None 
+                self.save_sel = None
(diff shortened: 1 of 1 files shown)


edreamleo (approved):
Many thanks for this difficult fix.

PR #4637, opened 2026-04-25
PR fixes #4634: complete previous work

### LINKED ISSUE #4634: Finish the ctrl-c PR
Completes the work of PR #4631 by fixing various minor reversions.

See PR #4637.

edreamleo:
See #4634.

**Significant code changes**

- [x] Remove the evil `LeoKeyEvent.event` ivar.
- [x] `c.insertCharFromEvent`: remove an evil legacy from [#4624](https://github.com/leo-editor/leo-editor/pull/4624).
- [x] Remove the weird `LeoQtGui.insertKeyEvent` method, moving it to the attic.
- [x] Revise `LeoQtGui.widget_name`. This *is* a significant change!
   It might be better to use the base `LeoGui.widget_name` method.
- [x] `ec.doPlainChar` returns if the character isn't a plain character.
   Previously, it called `g.app.gui.insertKeyEvent`, which makes no sense at all.
   Indeed, `k.masterKeyHandler` should have checked that it calls `doPlainChar` only with plain characters.
   However, I am *not* going to change `k.masterKeyHandler` in any way!
   This was the only call to the retired `insertKeyEvent` method.
- [x] Remove weird `cursorStay` logic and it's associated setting.
- [x] `QLineEditWrapper.setAllText` strips '\n' and '\r' in single-line widgets.
   This restores reasonable code from the buggy PR.

**Other code changes**

- [x] Improve tracing in `k.masterKeyHandler` and its helpers.
- [x] `LM.traceShortcutsDict` gives a verbose trace by default.
- [x] Simplify `c.check_event`.
- [x] Fixed minor bugs in `LM.traceShortcutsDict` and `BindingInfo.__repr__`.
- [x] Minor breaking changes:
   Require kwargs in `LeoKeyEvent.__init__`, `KeyEvent.__init__`,  and `create_key_event`.

**Annotations and comments**

- [x] Replace `KWargs` and `Args` annotations with `Any`.
- [x] Replace  'Value' annotations with 'Any'.
- [x] Improve annotation in `td.add_to_list`.
- [x] Improve headline comments for several classes.

**Testing**

- [x] Fix`TestQtGui.test_bug_4626`: It tests that Leo delivers key events to the expected widgets.
  Testing the eventFilters was more challenging than expected. I'll leave well enough alone!

### DIFF
--- a/leo/commands/editCommands.py
+++ b/leo/commands/editCommands.py
@@ -2207,6 +2207,9 @@ class EditCommandsClass(BaseEditCommandsClass):
     ) -> None:
         c, p = self.c, self.c.p
         isPlain = stroke.find('Alt') == -1 and stroke.find('Ctrl') == -1
+        if not isPlain:
+            g.trace(f"Not plain: {stroke!r} {g.callers()}")
+            return
         i, j = oldSel
         if i > j:
             i, j = j, i
@@ -2215,8 +2218,6 @@ class EditCommandsClass(BaseEditCommandsClass):
             w.delete(i, j)
         elif action == 'overwrite':
             w.delete(i)
-        if not isPlain:
-            return
         ins = w.getInsertPoint()
         if self.autojustify > 0 and not inBrackets:
             # Support #14: auto-justify body text.
--- a/leo/commands/editCommands.py
+++ b/leo/commands/editCommands.py
@@ -2215,32 +2215,31 @@ class EditCommandsClass(BaseEditCommandsClass):
             w.delete(i, j)
         elif action == 'overwrite':
             w.delete(i)
-        if isPlain:
-            ins = w.getInsertPoint()
-            if self.autojustify > 0 and not inBrackets:
-                # Support #14: auto-justify body text.
-                s = w.getAllText()
-                i = g.skip_to_start_of_line(s, ins)
-                i, j = g.getLine(s, i)
-                # Only insert a newline at the end of a line.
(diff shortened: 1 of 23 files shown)


tbpassin (commented):
leoGui.py
-----------------
line 320, `create_key_event()` - this change is not only requiring the use of keywords in the args. It also changes the args because `event` is no longer included. I'm concerned that consumers of this method may not all get changed. The line item mentioning this change in the initial remarks for the PR imply no changes except for requiring keywords but this isn't correct.

Likewise, at around lines 354, the event arg has been removed.

There are other places that the event arg has been removed, e.g., `cursesGui2.py`. The same comment applies to them.

qt_events.py
----------------
line 110, `    d[id(obj)] = 1 + d.get(id(obj), 0)`
Let's evaluate `id(obj)` once not twice.

General comments
-------------------------
There are too many different kinds of changes in this PR. It would be better to put them into PRs of their own.  Examples are changing notation types to "Any", and the changes in `qt_gui.py` to `widget_name()`. This PR was already complicated without these other kinds of changes, none of which were needed to fix up the functionality which was the main purpose of the PR.

edreamleo:

- `create_key_event`: I agree that this is a significant change, but Imo it's crucial the eliminate the evil `event` arg.
- I see no way to split this PR into stand-alone parts.

In fact, I see no way to truly evaluate this PR except by merging it and waiting for unintended side effects.
No, this is far from a perfect strategy, but I've spent days testing and days thinking about all possible implications of even the slightest change.

Issue #1585, opened 2020-04-28
Notes re Leo's code base


edreamleo:
This issue will discuss why Leo's code base is as it is, and what I might do differently if I were to rewrite Leo now.

**Background**

Clearly, Leo's code base could be improved.  However, statements like "Leo's code base sucks" show a lack of engineering maturity. To be of any use, criticism must be specific.

**Factors complicating the code**

1. At startup, Leo must handle Leo settings embedded in various outlines.
2. Leo must define not only 'c', 'g', and 'p', but other "official" ivars such as
    'c.frame', 'c.frame.body', c.frame.log', 'c.frame.body.wrapper' etc.
3. Leo's core is almost completely separate from Leo's gui code.
    As a result, Leo supports multiple guis in a straightforward way.
4. Whenever possible, I have tried to maintain compatibility with older versions of Leo.
5. Scripts may update all parts of the gui.
6. Leo's clones complicate the tree code.
7. Leo's gui is complex, with support for multiple body editors and many kinds of syntax coloring.

**Improvements based on pyzo's code**

I think of Almar Klein as the "young genius" :-) It's always worth studying his code.

Leo uses the "has-a" design pattern, but pyzo's code shows the benefits of the "is-a" pattern. The code in pyzoSourceStructure.py is only one of many possible examples. 

The PyzoSourceStructure class *is* a QWidget, it doesn't *have* a QWidget. This simplifies everything. To make this work, all ivars unique to the PyzoSourceStructure class itself must start with a single underscore. Imo, this is a small price to pay.

The PyzoSourceStructure class uses QBoxLayout.addStretch to simplify layout. I wish I had known about this trick.

Most importantly, pyzo's code uses messages much more effectively than Leo does. This decouples pyzo's code more thoroughly than in Leo.

**Summary**

I have briefly mentioned several ways that I might have simplified Leo. There are many more.

Alas, there is no easy way of making wholesale changes to Leo's qt gui code. A rewrite would take months, at the least, and would surely introduce many bugs. Some of these bugs could cause data loss, as I know all too well from experience.

tbpassin:
> Alas, there is no easy way of making wholesale changes to Leo's qt gui code. A rewrite would take months, at the least, and would surely introduce many bugs. Some of these bugs could cause data loss, as I know all too well from experience.

[...]

PR #4639, opened 2026-04-27
PR: Fix copy/paste from the menu

edreamleo:
The bug arises from confusions between wrappers and widgets.

There are only a few places left in Leo's code where using `event.widget` is correct.
Almost all parts of Leo's *core* code should use the `event.w`.
`event.w` will always be some kind of wrapper, but *not* necessarily a *text* wrapper.

- [x] (The fix) `LeoKeyEvent.__init__`: make `event.w` a wrapper if necessary.
   The Aha! Make new wrappers *only* for Qt widgets. 
   Testing for `isTextWrapper` would be very wrong.
- [x] At last! `k.checkKeyEvent`: checks `event.w` correctly!
- [x] Use 'event.w' instead of 'event.widget` in the following methods:
   - `c.insertCharFromEvent`.
   - `g.openUrlHelper`.  Also, add guards.
   - `k.fullCommand`.
   -  `LeoFrame`:`cut/copy/pasteText`.
   - `LeoTree.onHeadlineKey`.
- [x] `QTextEditWrapper.mouseReleaseEvent`: allocate a `QTextEditWrapper` to replace the `QMouseEvent`.
    How did the old code ever work??
- [x] `k.callAltXFunction`: reset `event.w` and `event.widget` separately.
- [x] Remove the evil method called by `k.masterKeyHandler`.
- [x] **Won't do**: rename `x.edit_widget` to `x.edit_wrapper`.
    Global renames are surprisingly dangerous, as we have seen in the buggy PR.

**Other changes**

- [x] In four files: annotate `w` and `self.w` as a `QTextMixin`.
- [x] In class `QTextEditWrapper`: annotate (correctly!) the `widget` arg as `Any`.
    The widget may be other Qt objects besides `QWidget`.
- [x] `QTextEditWrapper.__init__`: All args must be kwargs. Require the `c` kwarg.
- [x] `QTextEditWrapper.setSelectionRange`: Add some guards.
- [x] Transfer the testing checklist to #1585.

**Changes for new PRs**

As the result of this work, I have opened the following issues:

- #4641.
- #4642.
- #4643.
- #4644.

**Testing checklist**

To be done regularly, especially just before merging:

- [x] Test copy/paste from the menu.
- [x] Test copy/paste from all text panes, including the 'Completion' tab.
- [x] Test copy/paste within headlines from the menu.
- [x] Test clicking URLs in body text.
- [x] Test typing in 'Log' and 'Completion' panes.
- [x] Test console gui.

### DIFF
--- a/leo/commands/abbrevCommands.py
+++ b/leo/commands/abbrevCommands.py
@@ -251,7 +251,7 @@ class AbbrevCommandsClass(BaseEditCommandsClass):
     # @+node:ekr.20150514043850.11: *3* abbrev.expandAbbrev & helpers (entry point)
     def expandAbbrev(self, event: LeoKeyEvent, stroke: g.KeyStroke) -> bool:
         """
-        Not a command.  Expand abbreviations in event.widget.
+        Not a command.  Expand abbreviations..
 
         Words start with '@'.
 
--- a/leo/commands/abbrevCommands.py
+++ b/leo/commands/abbrevCommands.py
@@ -59,7 +59,7 @@ class AbbrevCommandsClass(BaseEditCommandsClass):
         self.save_sel = None  # Saved selection range.
         self.store = {'rlist': [], 'stext': ''}  # For dynamic expansion.
         self.tree_abbrevs_d: dict[str, str] = {}  # Keys are names, values are (tree,tag).
-        self.w = None
+        self.w: QTextMixin = None
 
     # @+node:ekr.20150514043850.5: *4* abbrev.finishCreate & helpers
     def finishCreate(self) -> None:
--- a/leo/commands/baseCommands.py
+++ b/leo/commands/baseCommands.py
@@ -87,8 +87,7 @@ class BaseEditCommandsClass:
     def editWidget(self, event: LeoKeyEvent, forceFocus: bool = True) -> QTextMixin:
         """Return the edit widget for the event. Also sets self.w"""
         c = self.c
-        w = event and event.widget
-        # wname = c.widget_name(w) if w else '<no widget>'
+        w = event.widget if event else None  # event.widget is correct.
         if w and g.isTextWrapper(w):
             pass
         else:
--- a/leo/commands/bufferCommands.py
+++ b/leo/commands/bufferCommands.py
@@ -14,6 +14,7 @@ if TYPE_CHECKING:  # pragma: no cover
     from leo.core.leoCommands import Commands as Cmdr
     from leo.core.leoGui import LeoKeyEvent
     from leo.core.leoNodes import Position, VNode
+    from leo.plugins.qt_text import QTextMixin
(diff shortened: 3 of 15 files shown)


edreamleo:
@boltex @tbpassin At last, a small PR. The crucial changes are to `LeoKeyEvent.__init__` in `leoGui.py`.

What do you think?

edreamleo:
@tbpassin @boltex All my tests pass. Please review and approve.

tbpassin (commented):
leoFrame.py
---------------
Line 1033 `w = event.widget`
Using `w` for a `widget` conflicts with other recent changes to make it a `wrapper`. Suggest use a different name, maybe `widget` to avoid mixups later.

leoGui.py
------------
Line 376 - `edit_wrapper = c.edit_widget(c.p)`
This name is very misleading since it's not a widget but a wrapper. In view of all the effort to use widget and wrapper correctly, and the confusion it's caused in the past, it would be very desirable to rename  `c.edit_widget` to something more wrapper-like.

leoKeys.py
--------------
lines 3459, 3460 -
```
        # Important: event.w does *not* have to be a *text* wrapper.
        #            event.w can be *any* non-Qt widget.
```
Since we've gone to a lot of effort to make clear that wrappers are not widgets, it's very confusing to be told that any widget is acceptable here. And what distinguishes a non-Qt widget from a Qt one (aside from the obvious Qt interface? Please clarify this comment.

edreamleo:
@tbpassin Thanks for your reply.

> ## leoFrame.py
> Line 1033 `w = event.widget` Using `w` for a `widget` conflicts with other recent changes to make it a `wrapper`.

I think you've found a bug! I'll change to `w = event.w` and run a full set of tests.

> ## leoGui.py
> Line 376 - `edit_wrapper = c.edit_widget(c.p)` This name is very misleading since it's not a widget but a wrapper. In view of all the effort to use widget and wrapper correctly, and the confusion it's caused in the past, it would be very desirable to rename `c.edit_widget` to something more wrapper-like.

Will do. But I must retain an alias to `edit_widget` for compatibility.

> ## leoKeys.py
> lines 3459, 3460 - [Comment lines]
>     Important: event.w does *not* have to be a *text* wrapper.
>                       event.w can be *any* non-Qt widget.

> Since we've gone to a lot of effort to make clear that wrappers are not widgets, it's very confusing to be told that any widget is acceptable here.

Imo, the only correct check is that `event.w` *isn't* a Qt Widget.
But to your point, I'll add a comment referencing the code in `LeoKeyEvent.__init__`.

Many thanks for these excellent comments.

edreamleo:
@tbpassin Renaming the `edit_widget` method is surprisingly tricky. It deserves a separate PR.

edreamleo:
@tbpassin @boltex I've responded to all of Thomas's comments. I think this PR is ready to be merged.

Issue #4644, opened 2026-04-28
Rename c.edit_widget to edit_wrapper


edreamleo:
Fixed via PR #4646.

Throughout Leo, there are lines such as:

`w = c.edit_widget(p)` or `gui_w = c.edit_widget(p)`.

In *all* cases, `w` should be a `QTextMixin`, so the name `edit_widget` is confusing.

PR #4646, opened 2026-04-28
PR fixes #4643: clean Leo's gui code

### LINKED ISSUE #4643: Remove BaseEdit.editWidget
See PR #4646.

Many commands in Leo's core start with:
```python
 w = self.editWidget(event)
 if not w:
     return
```
Where `BaseEdit.editWidget` is defined as follows:
```python
def editWidget(self, event: LeoKeyEvent, forceFocus: bool = True) -> QTextMixin:
    """Return the edit widget for the event. Also sets self.w"""
    c = self.c
    w = event.widget if event else None  # event.widget is correct.
    if w and g.isTextWrapper(w):
        pass
    else:
        w = c.frame.body and c.frame.body.wrapper
    if w and forceFocus:
        c.widgetWantsFocusNow(w)
    self.w = w
    return w
```
This pattern obscures what is happening. In almost cases the proper replacement would be:
```python
w = event.w
if not g.isTextWrapper(w):
    return
```
Furthermore, the call to `c.widgetWantsFocusNow(w)` should rarely, if ever, be needed.

edreamleo:
This PR fixes the following issues:
- #4643.
- #4644.

OMG: `BaseEdit.editWidget` was the worst kind of faux helper!
OMG: Without `editWidget`, we can get rid of `event.widget`!!!.

`TestEditCommands.run_test` should call `c.doCommandByName` with a proper `event` arg!
```python
 event = LeoKeyEvent(c, w=w)  # Leo 6.8.9.
 c.doCommandByName(command_name, event)
```
After that, all edit commands can simply do:
```python
w = event.w if event else None
if not g.isTextWrapper(w):
    return <maybe some value>
```

**Significant Changes**

These three changes are milestones in Leo's codebase.

- [x] `TestEditCommands.run_test` calls  `c.doCommandByName` with a proper `event` arg!
- [x] Remove `LeoKeyEvent.widget`!
- [x] Replace all calls to `BaseEdit.editWidget` with the code shown above, with a few special cases.
 
**Other changes**

- [x] Create wrapper widgets for the find and change text in the Find tab.
- [x] Move `BaseEditCommandsClass.getRectanglePoints` to the `RectangleCommandsClass` class, where it belongs.
- [x] Remove unused `BaseEdit.getWSString`  and  `BaseEdit.keyboardQuit` methods.
- [x] Rename  `BaseEditCommandsClass._checkSel` to `_checkSelection`.
- [x] 'test_editCommands.py': The headlines of all tests start with `test_`.
- [x] Improve various other headlines.
- [x] Per [#4644](https://github.com/leo-editor/leo-editor/issues/4644), rename `c.edit_widget` to `headline_wrapper`, with aliases for compatibility.
- [x] Replace `gui_w` (or `e`) by `w` in several places. 
- [x] Improve headline comments of classes: Use the full class name, not some too-short abbreviation.
- [x] Simplify `x.widget_name` and ensure that `x.getName` returns a string, never 'None'.
    By default, these methods now return an empty string, not `repr(w)` or `<no name>`.
    This new convention helps when creating debugging traces.

**Testing**

- [x] Test copy/paste from the menu.
- [x] Test copy/paste from all text panes, including the 'Completion' tab and VR pane.
- [x] Test copy/paste within headlines from the menu.
- [x] Test clicking URLs in body text.
- [x] Test typing in 'Log'  and 'Completion' panes.
- [x] Test console gui.

### DIFF
--- a/leo/commands/abbrevCommands.py
+++ b/leo/commands/abbrevCommands.py
@@ -262,7 +262,7 @@ class AbbrevCommandsClass(BaseEditCommandsClass):
         # Verbose only for *both* 'abbrev' and 'verbose'.
         verbose = all(z in g.app.debug for z in ('abbrev', 'verbose'))
         c, p = self.c, self.c.p
-        w = self.editWidget(event, forceFocus=False)
+        w = event.w if event else None
         w_name = g.app.gui.widget_name(w)
         if not w:
             if trace and verbose:
--- a/leo/commands/abbrevCommands.py
+++ b/leo/commands/abbrevCommands.py
@@ -264,7 +264,7 @@ class AbbrevCommandsClass(BaseEditCommandsClass):
         c, p = self.c, self.c.p
         w = event.w if event else None
         w_name = g.app.gui.widget_name(w)
-        if not w:
+        if not g.isTextWrapper(w):
             if trace and verbose:
                 g.trace('no w')
             return False
@@ -650,7 +650,7 @@ class AbbrevCommandsClass(BaseEditCommandsClass):
         """
         c, p = self.c, self.c.p
         w = event.w if event else None
-        if not w:
+        if not g.isTextWrapper(w):
             return
         s = w.getAllText()
         ins = ins1 = w.getInsertPoint()
@@ -685,7 +685,7 @@ class AbbrevCommandsClass(BaseEditCommandsClass):
         all possible completions if the prefix is the same as the word.
         """
         w = event.w if event else None
-        if not w:
+        if not g.isTextWrapper(w):
             return
         s = w.getAllText()
         ins = ins1 = w.getInsertPoint()
--- a/leo/commands/abbrevCommands.py
+++ b/leo/commands/abbrevCommands.py
@@ -649,7 +649,7 @@ class AbbrevCommandsClass(BaseEditCommandsClass):
         This corresponds to C-M-/ in Emacs.
         """
         c, p = self.c, self.c.p
-        w = self.editWidget(event)
+        w = event.w if event else None
         if not w:
             return
         s = w.getAllText()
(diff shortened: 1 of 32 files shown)


edreamleo:
@boltex @tbpassin This PR is ready for review. I'm pleased with it!

tbpassin (commented):
editCommands.py
-------------------------
Line 543 `w = c.frame.tree.edit_widget(p)`

Shouldn't this be changed to 
` w = frame.editWidget(event)`

as in line 714?

if not, then add comment that `w` is supposed to be a __wrapper__ not a widget (If it is supposed to be a widget then change this variable name to `widget`). Note there is an inconsistency in the use of the name `edit_widget` between the tree and body, and this is likely to cause confusion later.

leoKeys.py
-------------
line 2608 `c.widgetWantsFocusNow(event.w)` -

Has this method been changed to know that `event.w` is a wrapper not a widget?

[note - I didn't review `test_editCommands.py`]

edreamleo:
@tbpassin I don't understand your first question:
- This PR eliminates `editWidget` entirely.
-  Furthermore, there was a comment, now deleted, saying to get `w` from `c.frame.tree.edit_widget`.

Your question about `c.widgetWantsFocusNow(event.w)` is a good one.

 `c.widgetWantsFocusNow` will work with either wrappers or widgets. Indeed, this method eventually calls `LeoQtGui.set_focus`:
```python
def set_focus(self, c: Cmdr, w: QWidget) -> None:
    """Put the focus on the widget."""
    if not w:
        return
    if getattr(w, 'widget', None):
        if not isinstance(w, QtWidgets.QWidget):
            # w should be a wrapper.
            w = w.widget
    w.setFocus()
```

PR #4651, opened 2026-04-30
Add publish.yml GHA workflow for GitHub Release + PyPI uploads

vivainio:
Adds a `Publish` GitHub Actions workflow that fires when a GitHub Release is created/published (via UI or `gh release create`). It builds an sdist and a wheel and:

1. Attaches them to the GitHub Release as downloadable assets.
2. Uploads them to PyPI via Trusted Publishing (OIDC, no API tokens).

This was discussed and agreed in earlier PR comments.

## Triggers

- `release: [published, created]` — the natural trigger; works for both `gh release create` and the web UI, draft or non-draft.
- `workflow_dispatch` — manual re-run against an existing release tag if a build fails.

## Required one-time setup before PyPI upload works

A PyPI owner of the `leo` project (e.g. Edward) needs to configure a Trusted Publisher:

1. Go to https://pypi.org/manage/project/leo/settings/publishing/
2. Add a publisher with:
   - **PyPI Project Name:** `leo`
   - **Owner:** `leo-editor`
   - **Repository name:** `leo-editor`
   - **Workflow name:** `publish.yml`
   - **Environment name:** `pypi`

The `pypi` GitHub environment is auto-created on first workflow run — no manual repo setup needed.

Until PyPI is configured, the `pypi` job will fail with an OIDC error, but the `github-release` job still succeeds and attaches assets to the Release. After PyPI is configured, just re-run the workflow.

## Test plan

- [ ] Edward configures the PyPI trusted publisher.
- [ ] Cut a test/pre-release via `gh release create` and verify assets appear on the Release page and on PyPI.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,64 @@
+name: Publish
+
+on:
+  release:
+    types: [published, created]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Existing release tag to build and upload assets for (e.g. v6.8.5)'
+        required: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.tag || github.event.release.tag_name }}
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+          cache: pip
+
+      - run: python -m pip install --upgrade build
+      - run: python -m build
+      - run: ls -la dist/
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: dist
+          path: dist/
+
+  github-release:
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          name: dist
+          path: dist
+
+      - env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          TAG="${{ github.event.inputs.tag || github.event.release.tag_name }}"
+          gh release upload "$TAG" dist/*.whl dist/*.tar.gz \
+            --clobber --repo ${{ github.repository }}
+
+  pypi:
+    needs: build
+    runs-on: ubuntu-latest
+    environment: pypi
+    permissions:
+      id-token: write
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          name: dist
+          path: dist
+
+      - uses: pypa/gh-action-pypi-publish@release/v1


edreamleo:
@vivainio I don't remember agreeing to this.

I use a lengthy by-hand checklist to create the assets. My concern is that failures in the automatic process may be difficult to recover from. I'd like to discuss this issue with you further.

vivainio:
@edreamleo  this is what I referred to in https://github.com/leo-editor/leo-editor/pull/4620

Note that you can still do all the checklist locally, but only upload by creating that release in "Github releases". In many ways it's less error prone that doing these things locally, and it's very fast to roll out patch releases as needed. I'm publishing all my python apps like this nowadays.

edreamleo:
@vivainio Thanks for the comments. I have several questions:
- Is there an instruction page?  
- Can I pretest with TestPyPi automatically first?
- Should I be logged in to PyPi first?
- How does the workflow know how to build Leo?

Stuff like that :-)

vivainio:
@edreamleo good questions:

> **Is there an instruction page?**

PyPI's official trusted-publishing docs: https://docs.pypi.org/trusted-publishers/ and the action's README: https://github.com/pypa/gh-action-pypi-publish. The whole flow is well-trodden — most modern Python projects publish this way now.

> **Can I pretest with TestPyPI automatically first?**

Yes. Two common patterns:

1. **GitHub "pre-release" toggle** — when you mark a Release as pre-release, route it to TestPyPI; non-prerelease goes to PyPI. We add a small `if` per job.
2. **Manual `workflow_dispatch`** with a `target: testpypi|pypi` input — you can fire a TestPyPI test build at any time without cutting a real release.

TestPyPI has its own trusted-publisher config (https://test.pypi.org/manage/account/publishing/) — same fields, separate registration. I can add this to the workflow if you want.

That said — for a project with a long release history, an even smaller risk path is to publish a **pre-release** version (e.g. `v6.8.5rc1`) directly to real PyPI. `pip install leo` won't pick it up unless users pass `--pre`, so it's effectively invisible to normal users while exercising the full automated path against the real registry.

> **Should I be logged in to PyPI first?**

Not at release time. Trusted Publishing uses OIDC — GitHub Actions presents a short-lived token to PyPI, no API tokens, no `~/.pypirc`. You only need to be logged into pypi.org **once** to register the trusted publisher in the project settings. After that, releases authenticate themselves.

> **How does the workflow know how to build Leo?**

It doesn't — `pyproject.toml` does. The workflow runs plain `python -m build`, which reads the existing `[build-system]` and `[tool.setuptools]` sections in `pyproject.toml` and produces an sdist + wheel — exactly what you'd get running `python -m build` locally. The workflow is generic; nothing in it is Leo-specific. If `python -m build` works on your machine today, it works in CI.

### About recovery from failures

A few reassurances:

- A failed PyPI upload doesn't break anything — PyPI rejects duplicate filenames, so a re-run after fixing the issue just works. There's no "half-published" state.
- If something goes truly wrong, you can always fall back to your manual checklist — this workflow doesn't replace anything, it just automates the common case.
- The GitHub Release assets and PyPI upload are independent jobs — if PyPI fails, the wheel is still attached to the Release and you can `twine upload` it by hand.

edreamleo:
@vivainio Thanks for your detailed and reassuring answers. I'm convinced.

edreamleo (approved):

vivainio:
Cool. After you have done the preparatory steps on pypi, let's see how well this works - try to create some pre-release build in "Releases" pane and we'll see what kind of success we get. I can fix it if the workflow fails somehow

edreamleo:
@vivainio Sounds like a plan.

PR #4658, opened 2026-05-03
PR fixes #4645: diff quirps

### LINKED ISSUE #4645: Improve Leo's git-diff-pr command
See PR #4658 

Leo's diff-related commands should clone diff nodes only if node's branch matches the branch from which Leo is running.

edreamleo:
See #4645.

- [x] Clone nodes only if they are in the current branch.

### DIFF
--- a/leo/commands/editFileCommands.py
+++ b/leo/commands/editFileCommands.py
@@ -1315,6 +1315,11 @@ class GitDiffController:
         """Create nodes describing the changes."""
         if not d:
             return
+        branches_match = (
+            branch == rev2 or
+            rev2 == 'HEAD' or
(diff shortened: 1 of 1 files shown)


edreamleo:
@boltex This PR probably doesn't affect LeoJS, but I suppose it might.

boltex:
@edreamleo  I've ported this to LeoJS. Thanks! :)

edreamleo:
@boltex And thanks for your quick work.

It looks like PR #4660 will also affect LeoJS. The PR fixes bugs in the `cff` and `find-all` commands and improves the `find-all` commands. Stay tuned.

PR #4666, opened 2026-05-05
PR fixes #4664: crash in leoclient.py

### LINKED ISSUE #4664: crash in leoclient.py
Fixed via PR #4666.

leoclient.py crashes, even with leoserver.py running:

```console
C:\Repos\leo-editor>python -m leo.core.leoclient
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Repos\leo-editor\leo\core\leoclient.py", line 265, in <module>
    main()
    ~~~~^^
  File "C:\Repos\leo-editor\leo\core\leoclient.py", line 254, in main
    loop = asyncio.get_event_loop()
  File "C:\Python\Python3.14\Lib\asyncio\events.py", line 715, in get_event_loop
    raise RuntimeError('There is no current event loop in thread %r.'
                       % threading.current_thread().name)
RuntimeError: There is no current event loop in thread 'MainThread'.
```
Further testing reveals problems with leoserver.py.

edreamleo:
See #4664.

- [x] Fix various crashers.
- [x] Catch more exceptions and report them more clearly.
- [x] Improve traces and exception handling.
- [x] `leo/modes/python.py`: colorize `ConnectionRefusedError`.
- [x] Improve module-level docstrings from `leoserver.py` and `leoclient.py`.
- [x] Suppress confusing message when instantiating the `LeoServer` class from `leoclient.py`.

**Testing**

- [x] Test with Python 3.10.
- [x] Test with Python 3.14.

### DIFF
--- a/leo/core/leoclient.py
+++ b/leo/core/leoclient.py
@@ -198,7 +198,7 @@ async def client_main_loop(timeout):
                     "param": param,
                 }
                 if trace:
-                    print(f"{tag}: send: id: {n} package: {request_package}")
+                    print(f"{tag}: send: id: {n:4} {action:>40} {param}")
                 # Send the next request.
                 request = json.dumps(request_package, separators=(',', ':'))
                 await websocket.send(request)
--- a/leo/core/leoclient.py
+++ b/leo/core/leoclient.py
@@ -24,6 +24,7 @@ verbose = False
 timeout = 0.1
 times_d: dict[int, float] = {}  # Keys are n, values are time sent.
 tot_response_time = 0.0
+n_async_responses = 0
 n_known_response_times = 0
 n_unknown_response_times = 0
 
@@ -164,11 +165,6 @@ def _show_response(n, d):
 
 
 # @+node:ekr.20210205144500.1: ** function: client_main_loop
-n_async_responses = 0
-n_known_response_times = 0
-n_unknown_response_times = 0
-
-
 async def client_main_loop(timeout):
     global n_async_responses
     uri = f"ws://{wsHost}:{wsPort}"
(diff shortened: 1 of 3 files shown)


edreamleo:
@boltex This PR is ready for review. Please test!

boltex:
@edreamleo Is that what I'm supposed to see?
image

boltex:
@edreamleo It works if I pre-start a server independantly first. Then leoclient connects to it (while still starting its own useless server instance on the next available port) 

and finishes without errors :
image

boltex:
@edreamleo 
Maybe starting an instance of LeoServer is required to use "inspect.getmembers" on it? 

If its the case then there should be some comment or 'print' call that tells the user to pre-start a leoserver the proper way. 

 (creating an instance of LeoServer class does not work because that does not invoke main() in leoserver.py)
 
 In fact, thinking more about it, I guess this leoclient program needs to start with a "print" statement saying 'This leoclient program is intended for 'reason' ... to use it please do 'xxx' . etc.. 
 
**Note**: I tried with python 3.11 and 3.14 

 **Note2**: I noticed there was small modifications to leoserver so I tried LeoInteg, no problem there 😄

edreamleo:
@boltex Thanks for your comments!

Remember that leoclient.py is an *example* of a Leo client:
- It uses hard-coded constants. A better client would use command-line arguments to override those constants.
- It executes a fixed set of commands, the last of which shuts down the server!

Question 1: Yes, the client assumes that the server is already running.
Questions 2 & 3: I wasn't aware that the client tries to start the server. I'm not entirely sure that's true. I'll look into this.

boltex:
@edreamleo Thanks for those explanations. 

I see now that the leoclient does not actually start a websocket server, it merely emits one of its starting string when making an instance of the LeoServer class (not the whole server) to call "inspect.getmembers" on it.  All good :)

boltex (approved):
Looks good :)

edreamleo:
@boltex Thanks for your review and approval.

edreamleo:
@boltex The last two revs improve docstrings and suppress the confusing message from the LeoServer class when the client starts.

PR #4670, opened 2026-05-06
PR fixes #4662: Remove shell=True everywhere

### LINKED ISSUE #4662: Potential command injection vulnerability in adoc & pandoc commands
Found by **Hideaki Takahashi**, (@Koukyosyumei) a PhD student working on system security at Columbia University.

**Comments by EKR**

PR #4670 almost completely removes `shell=True` from Leo's codebase, including within `g.execute_shell_commands`.
Doing so is a minor breaking change to Leo's scripting API.  It's well worth doing.

In practice, however, the PR will have very little effect on Leo's security. See Leo's info issue #4455 for details.
The overriding principle is:

## Do not open .leo files created other people unless you trust them!

<details><summary><b>Original Report</b></summary>
<br>

A .leo outline can carry an @adoc or @pandoc headline whose suffix becomes the input path of a subprocess.Popen(..., shell=True) invocation without quoting. A user who opens such an outline and runs the documented `adoc` (or `pandoc`) command on the tree will execute the embedded shell payload as themselves.

In leo/core/leoMarkup.py:387 and :398:, 

```python
def run_asciidoctor(self, i_path: str, o_path: str) -> None:
    ...
    command = f"{prog} {i_path} -o {o_path} -b html5"
    g.execute_shell_commands(command)        # → subprocess.Popen(command, shell=True)

def run_pandoc(self, i_path: str, o_path: str) -> None:
    ...
    command = f"pandoc {i_path} -t html5 -o {o_path}"
    g.execute_shell_commands(command)
```

In MarkupCommands.command_helper (leoMarkup.py:299-320):

```python
for p in roots:
    i_path = self.filename(p)                   # ← attacker: node headline
    i_path = c.expand_path_expression(i_path)   # only ~ and $VAR — NOT $(...)
    n_path = c.getPath(c.p)
    i_path = g.finalize_join(n_path, i_path)    # path concat — NO escaping
    with open(i_path, 'w', encoding='utf-8', errors='replace') as ...:
        self.write_root(p)
        i_paths.append(i_path)
...
for i_path in i_paths:
    o_path = self.compute_opath(i_path)
    if kind == 'adoc':
        self.run_asciidoctor(i_path, o_path)    # ← reaches the shell sink
```

self.filename(p) returns the substring of the headline after @adoc / @pandoc. c.expand_path_expression (leoCommands.py:3572) only runs os.path.expanduser and os.path.expandvars — $(...) command-substitution is left intact. g.finalize_join does no shell escaping. Shell meta-characters from the headline survive verbatim into the f-string passed to subprocess.Popen(..., shell=True).

If a user opens an attacker-supplied .leo outline containing a headline such as (see the full executable PoC at the end of this e-mail) "@adoc evil;cd;touch leo-pwned;#", and then runs the adoc command, the injected shell command executes.

To fix this bug, we need to avoid shell command strings for these calls. Build argv lists and run with shell=False, for example:

```python
subprocess.run([prog, i_path, "-o", o_path, "-b", "html5"], check=False)
subprocess.run(["pandoc", i_path, "-t", "html5", "-o", o_path], check=False)
```

This avoids needing shell escaping because argument boundaries are preserved by the subprocess.

</details>

edreamleo:
See #4662.

- [x] Remove `shell=True` everywhere except where executing knows Windows commands.

**Breaking changes**

-  `g.execute_shell_commands` no longer sets `shell=True`. 
- The `at_produce.py`, `mime.py`, and `run_nodes.py` plugins may be affected.

**Other changes**

- [x] Use the following pattern to execute Python code within Python scripts.
```python
python = sys.executable
command = f"{python} -m whatever {args}"
g.execute_shell_commands(command)
```

**Testing**

- [x] Check that removing `shell=True` does not affect the scripts in `leo/scripts`.
   Trivial check: all contains fixed arguments to `subprocess.Popen'.

### DIFF
--- a/leo/core/leoBackground.py
+++ b/leo/core/leoBackground.py
@@ -245,20 +245,12 @@ class BackgroundProcessManager:
         Start or queue a process described by command and fn.
         """
 
-        # Note: we can't set shell=True (at least on Windows) because the process
-        #       terminates immediately.
-        #
-        #       Anyway, setting shell=True is supposedly a security hazard.
-        #       https://docs.python.org/3/library/subprocess.html#security-considerations
-        #
-        #       However, we do not expect tools such as pylint, mypy, etc.
-        #       to create error messages that contain shell injection attacks!
         def open_process(data: ProcessData) -> Popen:
             g.es_print(f'{data.kind}: {g.shortFileName(data.fn)}')
             self.process_return_data = []
             proc = subprocess.Popen(
                 command,
-                shell=False,
+                shell=False,  # #4662.
                 stderr=subprocess.PIPE,
                 stdout=subprocess.PIPE,
                 universal_newlines=True,
--- a/leo/core/leoBackground.py
+++ b/leo/core/leoBackground.py
@@ -250,7 +250,6 @@ class BackgroundProcessManager:
             self.process_return_data = []
             proc = subprocess.Popen(
                 command,
-                shell=False,  # #4662.
                 stderr=subprocess.PIPE,
                 stdout=subprocess.PIPE,
                 universal_newlines=True,
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -1081,7 +1081,7 @@ class Commands:
                 first_line = f.readline()
             return first_line.startswith('#!')
 
-        # @+node:tom.20241014154415.20: *4* runFile
+        # @+node:tom.20241014154415.20: *4* runFile @cmd c.execute-external-file
         def runfile(fullpath: str, processor: str, terminal: str) -> None:
             direc: str = os.path.expanduser(os.path.dirname(fullpath))
             if g.isWindows:
@@ -3392,7 +3392,7 @@ class Commands:
         try:
             proc = subprocess.Popen(
                 final_command,
-                shell=True,
+                shell=False,
                 stdout=subprocess.PIPE,
                 stderr=subprocess.PIPE,
             )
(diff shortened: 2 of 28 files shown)


edreamleo:
@Koukyosyume @tbpassin @boltex This PR is ready for review.

tbpassin:
An important question is whether any real-life uses include arguments that expect the shell to expand them. This usually happens when the shell is supposed to deal with quoted and escaped arguments, I think.

edreamleo:
@tbpassin I agree: that's the important question.

I've reviewed all the calls that used to have `shell=True':

- In most cases, the args have already been computed, so there is clearly no need for `shell=True`.
- In a few places, the intention seems to be execute a console command in Windows.
  I've left such code unchanged: the Windows console should know how to deal with injection attacks.
- I've removed `shell=True` in two or three obscure plugins.
  If someone complains I'll discuss the situation with them, but for now I don't want `shell=True` anywhere dubious.

Does this plan seem reasonable to you, Thomas?

tbpassin:
I was thinking more about user scripts. I remember a few times I couldn't get a shell-out command to work unless I used `shell=True`. I don't *think* I have any of those that I use all the time. But maybe someone else does.

I think I would prefer creating a method with a new name that always sets `shell=False`. All known instances (but for those few exceptions you mentioned) would get converted to use the new method. The old method would be marked as deprecated.

If no one actually depends on `shell=True` in their own scripts, then this will be perfectly safe. If they do, at least they can get warned about it.

edreamleo:
@tbpassin Thanks for your comments. Imo, no further action is needed.  If somebody wants `shell=True` their scripts can do so explicitly:
```python
proc = subprocess.Popen(command, shell=True)
proc.communicate()
```
The docs for 6.8.9 will mention the change as a breaking change.

boltex:
@edreamleo 

I think the at_produce.py, mime.py, and run_nodes.py plugins should have been left alone- its their specific purpose to run things the way they did, they are not security issues related to the original issue that Koukyosyume had raised.

edreamleo:
@boltex Thanks for your comments, but I'm not convinced. Let's see if anyone complains about the change. For now, I'd rather use `shell=True` only if obviously necessary.

Issue #4590, opened 2026-04-14
Abbreviations : "_NEXT_PLACEHOLDER" inserted using ",," in headline


boltex:
Note: See PR #4598 

Using double comma to go to next match to replace works well **from body** to headline, and **from body** to body, but when doing **from a headline** it inserts **_NEXT_PLACEHOLDER** instead.

boltex:
@edreamleo I just realized I've got the same bug in leo-web as I'm implementing the abbreviations and reproduced the same behavior 😄 

... so if I find the source, I'll fix it in Leo also.

boltex:
This was introduced by #4462

Issue #4529, opened 2026-02-14
Simple abbreviations in headlines do not work


edreamleo:
Fixed via PR #4530.

Rev 477ec61 (PR #4462) is buggy.

With these abbreviations:
```console
alr;;=@language rest
date;;={|{x=time.strftime("%Y/%m/%d")}|}
```

The `date;;` abbreviation is working for me everywhere, but the `alr;;` abbreviation inserts only `alr;` into the headline.

edreamleo:
Fixed via PR #4530.

jknGH:
Sorry to be a pain, but I have found some behaviour that is (ahem) 'less performant' than it used to be, and which probably relates to this work.

I used to be able to do something like this:

 1. create a new node CTRL-RETURN; enter 'highlight editing mode'
 2. start to edit node headline
 3. realise that I want this to be a child node; press SHIFT-ALT-RIGHTARROW (bound to command 'move-outline-right')
   3a. node is indented; still in highlight editing mode
4. finish editing headline; press RETURN to exit mode
5. ready to press RETURN to edit node body, etc.

Now, after (3), the node is indented but 'edit highlight mode' is exited.

The current behaviour is not a massive pain but I think the earlier is how it should be...

Thanks, J^n

edreamleo:
I have just confirmed (in the master branch) that after step 3 (that is, after moving the new node right) the headline is *not* in editing mode. Indeed, it would be quite wrong to attempt to move a node without exiting headline editing mode.

Why? The short answer is that changing the headline-related low-level code is extremely dangerous. It's way too easy to lose headline code when changing code. There is no way I'm going to change Leo's headline-related code.

I am going to close this issue. The important case is that adding an abbreviation to a headline *does* continue headline-editing mode.

jknGH:
Well ... OK. I am pretty sure that it used to work this way, but perhaps this was by lucky accident, and I'm not going to scream about the loss of this feature, if you think it too dangerous to reinstate.

I agree that the abbreviation fix is the most germane.

PR #4464, opened 2025-10-23
PR for #4462: abbrevs in headlines

edreamleo:
See #4462.

- [x] Handle headlines more simply.

PR #4598, opened 2026-04-15
PR fixes #4590: abbreviations

boltex:
Fixes #4590

PR #4675, opened 2026-05-09
PR fixes #4614: abbreviations (again)

### LINKED ISSUE #4614: Sample abbreviations in leoSettings: "importer;;" and "demo;;" samples do not work
The "importer;;" abbreviation sample does not work properly. It replaces properly the scripting abbreviations delimited with `{|{` and `}|}` in body panes, _but not in headlines_.

It also does properly set selection on the next replacable placeholder delimited with `<| `and `|>` . 

So the problem is really just the replacement of parts delimited with ` {|{ `and `}|}`  in the headlines.

e.g.: in the screenshot below showing the definition of the "importer;;" abbreviation, the `{|{x=cap_name}|}` in the headline do not get replaced, but the ones in the body pane do.

Image

edreamleo:
See #4614.  This PR supersedes PRs:  #4464,  #4590, #4529.

### Overview: The nerdy `,,` convention is gone!!

When expanding a tree abbreviation, Leo sets up a regex find for the substitution template , `<|...|>`.
Doh! F3 finds the first/next template. F2 finds the previous template.

### Significant changes to leoAbbrev.py

- [x] Remove all code related to the `,,` logic. Many difficult methods are gone!
- [x] Completely rewrite and refactor `expandAbbrev`:
   It has collapsed in complexity. No more evil kwargs!
   The new code *looks* easy, but setting the insertion point is extremely challenging.
- [x] Make tree abbreviations undoable.
- [x] `LeoSettings.leo`: retire `@string abbreviations-next-placeholder = ,,`.
- [x] Create 3 new unit tests. Retire tests related to `,,`.

### Suggestions from Félix:

- [x] Do the first search automatically when changing the find settings.
- [x] Search for `<|...|>` in non-tree abbreviations.
- [x] Change the Find settings only if the expanded text contains <|...|>.
    Extra: write a message when updating the find settings.
- [x] Evaluate scripting substitutions in `p.b` before `p.h`. This fixes the `importer;;` abbrev.
- [x] Fix tree-related bugs: rev 70834ff.
- [x] Unit test: verify that the initial find command selects `<|...|>`.

### Other changes

<details>
<br>

**Testing**

- [x] Move two tests from `test_editCommands.py` to `test_leoAbbrev.py`.
- [x] Test demo.py. More work is needed, but that does not affect this PR.

**Other changes to leoAbbrev.py**

- [x] Add `__slots__.
- [x] Improve `abbrev.listAbbrevs`.
- [x] Remove several unused ivars and methods.
- [x] `init_ivars`: set defaults to the defaults in `LeoSettings.leo`.
   This allows unit tests to work independently of settings files (which they don't read!).
- [x] `replace_selection` updates selections immediately, removing the evil selection caching ivars.
- [x] `abbrev._substitution_helper`: avoid recursive substitutions.
- [x] Match `<|` and `|>` only if they occur on the same line.

**leoApp.py**

- [x] Call `abbrev.listAbbrevs` on startup when `--trace=abbrev`.

**LeoFind.py**

- [x] `LeoFind.interactive_search_helper`: Add the`dry_run` kwarg and make other improvements.

**qt_text.py**

- [x] `QTextEditWrapper` class: Remove evil (and buggy) kwargs from `setInsertPoint` and `setSelectionRange`.

**leoUndo.py**

- [x] Fix several quirps.

**LeoSettings.leo**

- [x] Fix the `demo;;` abbreviation.
- [x] Add `details;;` abbreviation.
- [x] Remove an overly wonky comment.

**demo.py**

- [x] `Demo.start`: require kwargs.

</details>


edreamleo:
@boltex Please test and review. I want this PR to be the *last* abbreviation-related PR!

boltex:
@edreamleo 

I tested the two main behaviors to be addressed .

### 1-scripted replacements with **{|{** and **}|}** in the headlines

I tried the `importer;;` abbreviation and the problem remains: the scripted replacements (those delimited with **{|{** and **}|}** ) in the headlines are not being replaced. 

Note that when looking at the given 'importer;;' sample tree-abbreviation given in leoSettings.leo, that the `x=get_language()` placeholder-script **is in the body of the first node**, while its headline requires the chosen name (with that `{|{x=name}|}` placeholder-script). This implies that the order when trying to interpret and replace those placeholders between **{|{** and **}|}**  seems to be body first, then the headline.

### 2-Using double comma to go to next match from within a headline

The modifications here seem to have broken something else that was fixed recently: Using the 'next-placeholder' abbreviation (double commas **,,**) in a headline does not work: it moves the cursor or prints '__NEXT_PLACEHOLDER' directly in the headline. (which is a regression from https://github.com/leo-editor/leo-editor/pull/4598/changes which had fixed this issue.)

edreamleo:

At present, abbreviations in headlines are limited: only simple abbreviations make sense in headlines.

I'll rerun my tests soon with your comments in mind.

boltex:
@edreamleo Thanks! :)

To be clear, when I said the placeholder scripts should be interpreted in the body first, then the headline, I meant still in natural tree order. (a node's body then its headline, then the next node's body, and its headeline, etc.) not all bodies then all headlines.

edreamleo:
@boltex Recent revs simplify the code. All my tests pass:

-  `{|{` and `}|}` work as expected in headlines.
- To repeat, I see no reason to support `,,` when replacing text in headlines.
- But `,,` does find the next `<|...|>` in all other cases, looking first in headlines and then in the body. 
  See `abbrev.find_place_holder`.

The terminology is confusing, but imo it makes no sense to say:

    "placeholder scripts should be interpreted in the body first, then the headline."

`make_script_substitutions` can *only* change the current widget, either `c.p.h` or `c.p.b`.
But maybe you were referring to `,,`?

In short, I think all is well. What do you say?

boltex:
@edreamleo 
First off, I'd just would like to thank you for taking the time to fix the abbreviations. I only had learned about abbreviations a month ago when I ported it to leo-web, I had never used it in the past - But after playing with it since then, I totally see why people where asking for it: It's a killer feature! 😄  

**I have only had time to test it and not to review the code. I'll do review the code soon when I have a bit more time.** 

But after briefly trying out the latest modifications, I've noted the following : 

1- using double comma `,,` to goto and select the next 'placeholders' delimited by `<|`and `|>`, now goes to the next one, and so on, including the ones in headlines, **but they leave behind a single comma**, (I guess the slice must be _off by one_ when doing the cleanup before jumping to the next one ... heh)

2- When doing scripted "string" abbreviations (just text no tree insertion ), **that works well**!  (tested with the date;; abbreviation in both bodies and headlines.)

3- When doing scripted "outline" abbreviations (those defined in `@outline data tree-abbreviations` which insert a chunk of outline ) there seems to be some problems. (I tested with the `importer;;` abbreviation) : **It does not replace scripted placeholders in headlines.** the sample `importer;;` abbreviation make it clear that they are supposed to be replaced the same way as in the body pane. (if I understand it correctly)

   Note: I acknowledge that those "outline" abbreviations defined in `@outline data tree-abbreviations`  can and should only be triggered while typing in the body pane. (unlike text only scripted abbreviations like `date;;`) and that's totally expected. I'm talking about the replacement of those scripted placeholders defined by `{|{` and `}|}` that are in headlines once the tree has been inserted by typing the abbreviation.
   
   The problem (and the source of my intuition that the replacement of scripted placeholders defined by `{|{` and `}|}` should be done in natural tree order but bodies first then the headline when going node to node) is apparent when looking at the "get_language" abbreviation subst in the **@data abbreviations-subst-env** node in leoSettings.leo in cunjunction with the 
 tree-abbreviation that uses it: `importer;;` In particular, the first node inside that `importer;;`
 
 After looking at what the _"get_language" abbreviation subst_ does, (it sets variables like name and cap_name used in  scripted placeholders defined by `{|{` and `}|}`) , and then looking at the body of the first node in that   `importer;;` abbreviation, you will notice the presence of that **get_language abbreviation subst ** first line of body text doc string: `The @auto importer for the {|{x=get_language()}|} language.`. That get_language script will run when trying to replace it, and it will set the nam and cap_name variables availaable for all other scripted substitution of that abbreviation. But if you now look at the headline of that same node: it uses the **name** variable: 
 
 **@@file importers/{|{x=name}|}.py**
 
 So that is why I deduced that the order of replacing all scriping placeholders, (defined by `{|{` and `}|}` ) should be the natural tree order but do the body and then headline when going from node to node. (that first node in the `importer;;` needs a replacement in its headline with a variable that gets created when running the _"get_language" abbreviation subst_ in its body.
 
hehe, anyways- Hope this makes sense, I tried to use better nomenclature than that my last mention of this a few days ago, which I admit was quite vague :)

I'll do more testing soon - dont hesitate to ask for more details if any of my explanations are unclear.

Félix

edreamleo:
@boltex Thanks for your lengthy comment. I have greatly enjoyed this project.

The old code was the product of many people, and many hacks. The new refactoring was urgently needed. The code will never be straightforward, but now it's at least sane.

The terminology is extremely confusing. Alas, most of the confusion is grounded in the names of user settings, which for obvious reasons I do not wish to change. To ease confusion, I've added comments to the code. For a start, here are some lines from `abbrev.init_settings`
```python
self.next_placeholder = getString("abbreviations-next-placeholder") or ''  # ,, by default
...
c.abbrev_place_end = getString('abbreviations-place-end') or ''  # |> by default.
c.abbrev_place_start = getString('abbreviations-place-start') or ''  # <| by default.
c.abbrev_subst_env = {'c': c, 'g': g, '_values': {}}  # May be augmented in init_env.
c.abbrev_subst_start = getString('abbreviations-subst-start') or ''  # '}|}' by default.
c.abbrev_subst_end = getString('abbreviations-subst-end') or ''  # '{|{' by default.
```
I'll probably similar comments in various helpers as reminders so I won't get so confused.

I do not see the problems you report. I'm not sure what's going on.
Also, I'm good with having limited functionality in headlines.

I'm deep into unit testing now. My plan is to merge this PR into devel fairly soon so more people can test the new code.

In short, this PR is a big step forward. I'll be happy to fix any remaining issues as they arise.

boltex:
@edreamleo I only had a bit of time before bed tonight so I pulled and tried the provided sample abbreviations that are provided in leoSettings.leo, to see if I could find bugs or glitches :)

## 1- regular text abbreviation

Here's an illustration of what I briefly mentionned yesterday. I tested with the provided sample global abbreviation `html;;`

here's what I see in the body pane after typing `html;;`
image

I also see **Found earlier placeholder** strangely printed in the log pane even though I have no prior nodes nor text in the body pane.

Then, I typed foobar and two commas like so `foobar,,`
image

As you can see, it left a comma after the foobar when the double comma triggered the next placeholder selection.

I had  **@bool scripting-abbreviations = False**  so I tried again with **@bool scripting-abbreviations = True** just in case, but I got the same result. (although this should not have had any effect on that abbreviation anyways as it contains no scripting.)

## 2- The @bool scripting-abbreviations itself

I tried the   `importer;;` tree-abbreviation and the `date;;` abbreviation with the **@bool scripting-abbreviations** setting set to **False**, then **True** to see if that setting was applied and enforced correctly. Here are the results of my tests below. 

### The  `importer;;` tree-abbreviation

When testing the `importer;;` tree-abbreviation with **@bool scripting-abbreviations = True**, none of the abbreviations-subst are interpreted nor replaced. (neither in body panes nor headlines) I tried with  **@bool scripting-abbreviations = False** just in case some condition was inverted but I got the same result, so it seems like scripting is just not applied to tree-abbreviations.

Also, the cursor and selection were left in the top node instead of being placed on the first placeholder inside that outline. 

### The  `date;;` abbreviation

When testing the `date;;` abbreviation (which uses scripting) and **@bool scripting-abbreviations = True**, it did work as expected. 

But when trying it again with **@bool scripting-abbreviations = False**, which should not try to interpret anything and instead just replace with `{|{x=time.asctime()}|}`, I got this error:
image

when done in the body pane, it removed the "date;;" that I typed. Alternatively, when done in a headline it did correctly leave the `{|{x=time.asctime()}|}` as replacement, but I still got that _"name 'time' is not defined"_ error printed in the log pane.

Alright, I'm going to bed now,  let me know when you'd like me to do some more tests - I kind of like playing with abbreviations since I discovered that cool feature of Leo! 😁

edreamleo:
@boltex Many thanks for your comments. 

- 'Found earlier placeholder'. Will remove this useless message.
- Unwanted comma. At last I see what you describe. To summarize the picture, when expanding the `html;;` shortcut, when title is selected, typing two commas (with or without typing anything else previously), there is an unwanted comma left behind in the "title" area.
- It seems that scripting is not applied to tree abbreviations.  OOPS.
  ~~Such an enhancement is beyond the scope of this PR.~~
- "name 'time' is not defined": I see the behavior you describe.

edreamleo:
@boltex "name 'time' is not defined".

Wow, The abbrev code shouldn't execute any `exec` statement at all if ` @bool scripting-abbreviations` is False!!!

My present opinion that ensuring security within Leo is impossible if people run outlines given to them by bad actors. But we can still try to remove obvious security bugs.

edreamleo:
@boltex I do see the traceback concerning 'time' when `@bool scripting-abbreviations` is false.

Clearly, both `make_first_headline_substitution` and `make_script_substitutions` should do nothing when  `@bool scripting-abbreviations` is false. That solves the problem at its root. Many thanks for reporting this issue.

Happily, there may be a way to remove *all* special cases involving headlines. Stay tuned.

edreamleo:
@boltex Work is going well, but quirps remain. See the first comment for remaining issues.

edreamleo:
@boltex This PR is ready for review. I have rewritten almost every line of the code.

boltex:
@edreamleo 
Thanks for this major abbreviation fix effort!!

I very much like the replacement of the double-comma hack with F3, makes a lot of sense... Why did nobody think of that before? haha! 😄 

# Preambule

I've tested with the following abbreviations given as default examples in the leoSettings.py file:

(_non-scripted_) **details;;** and **html;;** 
(_non-scripted tree-abbreviation_) **demo;;** 

(_scripted_) **date;;**
(_scripted tree-abbreviation_) **importer;;**

I'd like to suggest that those 5 sample abbreviations given as examples in leoSettings.leo are the abbreviations against which I think we should 'test' this module's implementation. 

# Tests Results

I know what the amount of work you did represent, so I systematically and carefully tested each of those (which cover pretty much all use-cases) in order to map and target the specific remaining small quirps and bugs more easily. 

I've categorized my findings in 'Misc', 'Non-Scripted' and 'Scripted' sections below.

### Misc Abbreviations

- (bug) tree-abbreviations replace with their container node instead of their first child. So when trying with **demo;;** or **importer;;** I'm left with the node from leoSettings.leo that contained the tree to be replaced, instead of the intended tree itself. (e.g.: **demo;;** should leave me with an `@@button` node, **importer;;**  should leave me with an `@@file` node)

- (bug) I'm getting weird behavior when using a tree-abbreviation in the very first node of the outline (maybe its related to the bug described above and it will be fixed automatically if the bug above is fixed) When doing a tree-abbreviation as the  very top node of an outline, the tree is put below or inside the mode below itself, instead of simply being replaced-in-place.

Here's a screenshot to illustrate as its hard to describe. Here i'm about to finish typing **demo;;** in the first node of the outline:
image

After doing so I'm left with the node being replaced below that second node instead of being still the fist node of the outline:
image

### Non-Scripted Abbreviations

- (bug) non-scripted regular abbreviations replace the text but leave the cursor at the very-end instead of selecting the first placeholder (when placeholders are present)

- (bug) non-scripted abbreviations only seem to trigger the find text of `(<\|.*?\|>)` **for tree-abbreviations**. So regular **non-tree** abbreviations replace the text but without replacing the 'find-panel' text with `(<\|.*?\|>)`. (as they should, just like for tree abbreviations)

- (bug) When the find-panel does have its find text replaced with `(<\|.*?\|>)`, it should actually also do an actual find action to find the first match, in order to leave the user with the first match found & selected.

- (Suggestion) If the content in the abbreviation (regular or tree abbreviation) did **not** contain the opening placeholder text (default `<|`), dont bother replacing the find-panel text nor try to find the first match. (E.g. the **demo;;** abbreviation is such a case)

### Scripted Abbreviations

- (bug) **non-tree** scripted-abbreviation like **date;;** used in the body pane leave the cursor 2 characters to the left of were it should be. (should be after the last character it just replaced if there was no placeholder to be found&selected)

- (bug) When resolving the scripts of tree-abbreviations, the scripts in a node's body should be done before its headline. (Otherwise that subtle bug apparent with the **importer;;** abbreviations happens) 

I know how UI related code and bugs can be a **pain in the a\*\*** to get right, so I'm thanking you again for this work you've done so far with this abbreviations killer-feature re-vamp. ...And let me know if that last bug that I've listed is unclear at all. I'll illlustrate it clearly so its more apparent as to why the nodes' body script-placeholders should be resolved before their headline.

edreamleo:
@boltex Many thanks for your comments and testing. Here are my first thoughts:

### Tree abbreviations

You have probably found one or two serious bugs in tree abbreviations. I'll ensure that everything works at the top level, and when the node containing the tree abbreviation is the first node of the outline.

### Find commands (Non-scripted abbreviations)

- I agree: Leo should not change the find text if `<|...|>` does not appear in the expansion.
- It's easy to search automatically: In `init_place_holder_search`, just remove the `dry_run` kwarg in the call to `interactive_search_helper`.

My experience shows that messing with the search args quickly becomes annoying.
The annoyance almost never happens to me in practice because:
-  I almost never use tree abbreviations, except while testing ;-)
-  None of the non-tree abbreviations that I use all the time contain  `<|...|>`.

The "platinum" solution would be to add a new `revert_settings_on_fail` kwarg to `LeoFind.do_find_next`.
It would be straightforward to make that change, but for now I don't think it's worth doing.

### Scripting issues

There is a bug that you may not have discovered in the importer;; script. That bug exists in the "devel" branch. Leo does not expand `@@file importers/{|{x=name}|}.py` at all. The PR catches a NameError exception to mask this error.

I've tried to find the source of the bug. I don't think making scripting substitutions first in body text will help. But I'll look into your suggestion.

### Cursor issues

Yesterday, I tried and failed to put the cursor where it belongs in non-tree abbreviations. The code is fraught.
- It's nearly impossible to calculate where the cursor *should* be when using script expansion, as in date;;.
  Why? Because `_substitution_helper` would need to do some AI and then tell the callers where the insert point should be. This bug never bites me, because I only ever use date;; in empty headlines or bodies.
- Correctly computing selection ranges and insert point isn't enough.
  Somehow (!!) the abbrev code must prevent the rest of Leo from clobbering the selections!

### Summary

Thanks again for your comments. They have been helpful.

What I'll do:

- Fix the two bugs concerning tree abbreviations and add one or two new unit tests.
- Change the settings for the find command only if the expanded text contains `<|...|>`.
- Do the first search automatically when changing the find settings.

What I *won't* do:

- Obsess about placing the cursor. I've tried at least dozens of fixes with no joy.
  For me, abbrevs such as date;; and (especially) fmt;; work fine. The cursor goes to the end of the line.
  Otoh, some new inspiration may appear.
- Change the hack (in `expand_tree`) that makes scripting abbreviations undoable.

boltex:
@edreamleo 

> Leo does not expand @@file importers/{|{x=name}|}.py at all.  I've tried to find the source of the bug. I don't think making scripting substitutions first in body text will help. But I'll look into your suggestion.

That is the bug I tried to explain a few posts back, that will arrise if you convert the expressions in the headline first:  take a look at the **get_language** subst code... It sets the **globals** such as '**name**' . This happens in the first body (the body of the first node). so obviously its headline cannot replace the 'name' if that 'get_language' subst has not run yet in the body. 

😄 

Hence my observation from a few days ago: For that **importer;;** abbreviation to have worked in the past means that the resolution of script subst. sould be done in the natural tree order, **but the node bodies before their headlines.**

edreamleo:
@boltex Rev 597462dd4 does indeed fix the scripting problem with the `importer;;` abbreviation.

There is no special reason why bodies should be evaluated first, but no matter&mdash;I'm good with this hack for now.

boltex:
@edreamleo 

In **theory**, you are right, hehe, that there is no special reason why bodies should be evaluated first.

But In **practice**, as the saying goes, it is different... 😉 

Now, to clarify, this is only true when we consider the 5 given sample abbreviations given as examples to the user in leoSettings.leo, and, the suggestion I made to consider those 5 abbreviations to be the ones against which we should **implement & test** this module's implementation.

More precisely, what I mean by that is the reason I knew from the start that this would be required to make that **importer;;** abbreviation work, is that I studied the provided sample **@data abbreviations-subs-env** node in leoSetting.leo. In particular, its child called **get_language**.

image

The **importer;;** uses it once, at a crucial place: the body of its first node. 

So a reference to 'name' cant be resolved in its first node's headline unless that body has resolved its scripting substitutions first. 

Hope that helps clarifying this specific part of this abbreviation's behavior.

As always - thanks a lot for making all those stabilization and implementation efforts to this abbreviations module! Like I said a few weeks ago, it's a small module in leo's codebase but its a killer feature,  so I'll make sure to re-test everything systematically when you're done with this next batch of modifications and report back again with whichever quirp remains 😄

edreamleo:
@boltex This PR seems ready for review.

 Recent revs complete the recent round of works. These were valuable improvements. Thanks for suggesting them.

boltex:
@edreamleo 

Many Thanks, 

Again, I can appreciate what all that work represents, so I'll make sure to review it carefully.

boltex:
@edreamleo 

I gave it a round of testing. Found only very few small unresolved quirps now:

1 - Trying out the simple non-script, non-tree abbreviations **details;;** and **html;;**, I see that they leave the cursor flashing at the very end instead of having selected the first instance of `<| ,,, |>` placeholders found in them. (like the tree abbreviations do) **i.e.** this should leave that first  `<| ,,, |>` placeholder selected for the user to quickly type its first replacement string after triggering the abbreviation, so he can then press F3 to go to the next, (and so on in rapid succession to the next, if any more.)

2 - (_New quirp that was not present before_) When doing a **non-tree** scripted abbreviation such as **date;;**, in a _headline_, I now see my cursor end up at the very end of the body instead of the end of the headline where it should normaly rest.

3- Tree-abbreviations still have the bug where they replace with their container node instead of their first child.  (e.g.: **demo;;** should leave me with an `@@button` node, not a `demo;;` node. (that "demo;;" headline is the container of the **\@outline-data tree-abbreviation** node in leoSettings.leo, only its child should be used, not that node itself.)

4- When doing a tree-abbreviation such as demo;; or importer;; in the body of **the very first node of the outline**, it switches place with the second node instead of replacing-in place.

5- scripted Tree-abbreviations now replace everything perfectly inside of them and also perfectly select the first placeholder they contain. But a bug remains: (This is probably related to the bug **#** 4 above) If I do an tree-abbreviation such as **demo;;** or **importer;;** in an outline where the node above it is expanded, the abbreviation places the new node as last child of that node - instead or replacing the current node being typed in by the user.
 
Like I said, I suspect bug **#** 4 and **#** 5 to be related and there's probably a single bug underneath that provokes those two behaviors.

edreamleo:
@boltex Thanks for your testing.

Rev 953ab48 fixes bugs 1 and 2.

### Bugs 3 and 4

I see the behavior you describe, in an outline with only one node (a`demo;` node) and in my main testing outline.

Imo this behavior is essential for proper operation of undo/redo. I don't want to change it.

Furthermore, bugs 3 and 4 are really minor quirps. The user can easily see what's going on.

edreamleo:
@boltex Rev e725ac7 adds a unit test that would have caught the recent bugs.

Please review and approve this PR.

boltex:
@edreamleo 

Many thanks for this work! I tried the unittest you've added: It uses **QTextMixin** 
```python
if TYPE_CHECKING:
    from leo.plugins.qt_text import QTextMixin
```
which is not available on my machine, perhaps an `else: QTextMixin = any` would be required. 

 ... and that **"Searching for {start_pat}...{end_pat}"** log-pane message every time we use an abbreviation could be removed - but that's just a small detail.

## Code Review

I've now finally looked at the code and can confirm what you did **greatly simplifies** the abbreviation codebase! Wow! This is incredible!

Not only the code-reorganisation and cleanup, but the usage of **F3** instead of double-commas, and the whole thing being undoable actions sandwiched between **beforeChangeGroup** and **afterChangeGroup** making this a single clean undo bead is great!

## My attempt at the 3 remaining quirps...

Since this is now so close to perfection, I took a bit of time to see if those 3 remaining quirps (**# 3**, **# 4** and **# 5** above) stemming from the **expand_tree** method, could be fixed.

After extensive testing, I narrowed it down to a total of 5 things to check, so I just set 5 flags before your usual **expand_tree** process, and then adust the final position taking those flags into account. 

The final position of the pasted node is then correct in all cases. I also added 6 lines that leaves the proper node in place _instead of its 'container' node_, finishing the whole process as idealy required. 

## Success! All remaining quirps are gone!

So with this dozen of lines added, all remaining quirps are gone - and as far as I can test with that, the behavior seems perfect. - and still undoable as a single bead - Here's my fixed version of **expand_tree** :

```python
    def expand_tree(self, i: int, j: int, word: str, expansion: str) -> None:
        """
        Paste `expansion` as children of c.p.
        This happens *before* any substitutions are made.
        """
        c = self.c
        u, undoType = c.undoer, 'Expand Tree Abbreviation'
        if c.p.hasChildren():
            g.es_print('tree abbreviations must not have children', color='blue')
            return
        if not c.canPasteOutline(expansion):
            g.es_print(f"bad copied outline: {expansion}")
            return

        # Replace the old node with a new node as a group of 'undoable' beads
        u.beforeChangeGroup(c.p, command=undoType, verboseUndoGroup=True)
        self.replace_selection(i, j, '')
        isRoot = c.p.isRoot()            
        wasHoisted = len(c.hoistStack) > 0
        noSiblings = c.p.getParent() and c.p.getParent().numberOfChildren() == 1
        isFirstChild = c.p.getParent() and c.p.getParent().firstChild() == c.p
        prevSibling = c.p.copy().moveToBack()
        prevSiblingExpanded = prevSibling and prevSibling.isExpanded()
        if c.canDeleteHeadline():
            if prevSiblingExpanded:
                prevSibling.contract() # To prevent pasting as last child of prevSibling
            c.deleteOutline(op_name="Cut Node")
            c.pasteOutline(s=expansion)
            if noSiblings:
                c.moveOutlineRight() # inserted below instead of as child, so move right
            if isRoot:
                c.moveOutlineUp() # delete & paste made it second position, so move up
        else:
            c.pasteOutline(s=expansion)
            c.selectPosition(c.p.moveToBack())
            c.deleteOutline(op_name="Cut Node")
            if wasHoisted: 
                c.selectVisNext()
        # Finish by replacing container node with its first child.
        child = c.p.copy().moveToFirstChild()
        if child:
            c.selectPosition(child)
            c.moveOutlineLeft()
            c.goToPrevSibling()
            c.deleteOutline(op_name="Cut Node")
            if isFirstChild or (not isRoot):
                c.selectVisNext()
        if prevSiblingExpanded:
            prevSibling.expand() # Re-expand the previous sibling
        u.afterChangeGroup(c.p, undoType=undoType)
        c.redraw(c.p)
```

By the way, I could not have found this final fix if this whole class had not been rewritten and simplified like you did in the first place, So thanks again for that!! 😄

edreamleo:
@boltex We're getting close!

### from leo.plugins.qt_text import QTextMixin

Huh? Of course `leo/plugins/qt_text.py` exists! Nothing would work without it.

### "Searching for" log-pane message

I'll remove it.

### Code review

You're welcome. I would never have done the work without your prompting and support.

### expand_tree

Excellent. I'll commit your version.

boltex:
@edreamleo 

I'm getting this error:

```
..........................................................
======================================================================
ERROR: test_find_command_selects_place_holder (leo.unittests.commands.test_leoAbbrev.TestAbbrev.test_find_command_selects_place_holder)
----------------------------------------------------------------------
  File "C:\prog\leo-editor\leo\unittests\commands\test_leoAbbrev.py", line 114, in test_find_command_selects_place_holder
    def test_selection(kind, w: QTextMixin, expected: str) -> None:
                                ^^^^^^^^^^
NameError: name 'QTextMixin' is not defined

----------------------------------------------------------------------
Ran 869 tests in 17.279s

FAILED (errors=1, skipped=3)
```

Altough you are right that this is done in many places, it is not done in the unit tests: (maybe TYPE_CHECKING is false in unit tests?)
```python
if TYPE_CHECKING:
    from leo.plugins.qt_text import QTextMixin
```

boltex:
@edreamleo after re-checking, `from typing import TYPE_CHECKING` nor `TYPE_CHECKING ` itself should be in the unit tests. If I remove that from test_leoAbbrev.py all the unit tests work.

edreamleo:
@boltex  Rev cd74514 uses your version of `expand_text`. Thanks for this work!

I made a few small changes:
- Compute `parent` once, for visual clarity.
- Use `c.p.moveToBack()` instead of `c.p.copy().moveToBack()`.
  `c.p` is a property which returns a copy of the underlying ivar. So copying it is always redundant.
- Use active voice in comments :-)

leo/unittests/commands/test_leoAbbrev.py
```diff
@@ -0,0 +1,284 @@
+# @+leo-ver=5-thin
+# @+node:ekr.20260512145309.1: * @file ../unittests/commands/test_leoAbbrev.py
+"""Tests of leoAbbrev.py"""
+# pylint: disable=no-member
+
+import re
+import time
+from typing import TYPE_CHECKING
```
boltex, line 9: TYPE_CHECKING not used in unit tests

leo/unittests/commands/test_leoAbbrev.py
```diff
@@ -0,0 +1,284 @@
...
+from leo.core import leoGlobals as g
+from leo.core.leoGui import LeoKeyEvent, StringFindTabManager
+from leo.core.leoTest2 import LeoUnitTest
+
+if TYPE_CHECKING:
```
boltex, line 14: No 'if TYPE_CHECKING' needed

edreamleo:
@boltex Imo, the declaration of `QTextMix` is correct and  it would be confusing/wrong to define `QTextMixin` any other way.

Within `test_leoAbbrev.py`, `QTextMixin` only appears in declarations, as you can easily verify.

It's a bit of mystery. What version of mypy are you using?

edreamleo:
@boltex Sure, the `TYPE_CHECKING` guard isn't needed, but it's bad style to omit it. Something weird is going on.

boltex:
I've got this version of mypy: 

image

When running the code TYPE_CHECKING  is true but when I do `py -m unittest` i guess its false. (because simply importing with `from leo.plugins.qt_text import QTextMixin` without the check for TYPE_CHECKING   works. - as for all the typing checks in the rest of the unittests. which all work fine)

edreamleo:
@boltex 
```console
>python -m mypy --version
mypy 1.20.1 (compiled: yes)
```

boltex (approved):
LGTM

PR #4676, opened 2026-05-10
PR: Fix obvious bugs in cursesGui.py

edreamleo:
This PR tries to revive `cursesGui.py`.  Imo, this file is useless and should be retired.

- [x] Replace 'init` with 'main'.
- [x] Use Leo's bridge to create `g.app`.
- [x] Fix various other problems.
- [x] Add some guards to the `LeoKeys` class.
- [x] Change unit test so it doesn't complain about the missing `init' method.


edreamleo:
I am going to close this PR. It will remain closed unless somebody can convince me that `cursesGui.py` has any value.

PR #4677, opened 2026-05-10
PR: retire cursesGui.py

edreamleo:
- [x] Move `cursesGui.py` to the attic.
- [x] Remove all mention of `cursesGui.py` from Leo's core.


tbpassin (approved):
There may be some documentation files that need to get updated.

edreamleo:
@tbpassin Thanks Thomas!

PR #4678, opened 2026-05-10
PR: Remove unnecessary 'type: ignore' statements, etc.

edreamleo:
Change only core files.

Many suppression could be remove due to more general annotations or improvements in mypy.

Many of the remaining `# type: ignore` comments apply to import statements.
I have found no way to workaround those suppressions.
Yes, they are a clear pattern, but I'm a bit surprised mypy seems to require them.

**Non-trivial changes**

- [x] `checkerCommands.py`: remove call to shlex in `run_pylint`.
- [x] `controlCommands.py: execute_subprocess`: Change where `g.toUnicode` is applied.
- [x] `spellCommands.py`:
    - Use `with` clauses to limit scope of `f` vars.
    - Fix a potential bug: move `process_word` method to the base class.
- [x] `qt_gui.py`: Remove small sections in `make_script_button` and `execute_script_callback` methods.
- [x] Create new variable names in the following core files:
   `abbrevCommands.py`, `leoNodes.py`.

### DIFF
--- a/leo/commands/abbrevCommands.py
+++ b/leo/commands/abbrevCommands.py
@@ -152,7 +152,7 @@ class AbbrevCommandsClass(BaseEditCommandsClass):
             )
             script = script.replace("\r\n", "\n")
             try:
-                exec(script, c.abbrev_subst_env, c.abbrev_subst_env)  # type:ignore
+                exec(script, c.abbrev_subst_env, c.abbrev_subst_env)
             except Exception:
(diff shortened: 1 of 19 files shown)


edreamleo:

I am pleased with this PR. Every improvement paves the way for others.

boltex (approved):
Looks all good to me :)

edreamleo:

PR #4680, opened 2026-05-13
Support myLeoSettings .leojs JSON format

boltex:
This PR adds a small quality of life feature that I should have added a few years ago when support for JSON .leojs file format was added 😄 

A user may have saved his personnal preferences as JSON .leojs file: This PR adds support for both instead of only XML .leo files. 

So instead of just checking for the usual places for myLeoSettings.leo, it also checks for myLeoSettings.leojs at those same places, prioritizing .leo if present.


boltex:
It also fixes a typo in commanderHelpCommands.py‎

edreamleo (approved):
Looks good.

PR #4693, opened 2026-05-20
PR: clean LeoFind class

edreamleo:
This PR removes some confusing code in the `LeoFind` class and related unit tests.

- [x] Add `LeoFind.ivars` ivar. Use it in `LeoFind.init_ivars_from_settings`.
- [x] Remove `p` key from result returned by `LeoFind.default_settings`.
- [x] Use common keys for all settings bunches: user settings plus `in_headline` and `reverse`.
- [x] `LeoFind.interactive_search_helper`: init `self.reverse`.

### Changes to unit tests

- [x] Remove tests on `settings.p`. These did nothing!
- [x] Remove references to `settings.wrapping`. `LeoFind.wrapping` no longer exists!

### What I *didn't* do

I wanted to make the annotations of settings more concrete, but that did not turn out well:

- I experimented with changing settings from `g.Bunch`'s to `dict`s, with the annotation `dict[str, str | bool]]`.
     Alas, mypy complains that ivars have the wrong type.
     Changing the annotation to `dict[str, Any]` is really no better than an annotation of `g.Bunch`.
- An alternative would be to define a new `FindSettings` *class*, but the benefits would be few.

### DIFF
--- a/leo/unittests/core/test_leoFind.py
+++ b/leo/unittests/core/test_leoFind.py
@@ -290,7 +290,6 @@ class TestFind(LeoUnitTest):
         # Test 3.
         init()
         settings.search_headline = False
-        settings.p.setVisited()
         x.do_find_all(settings)
         # Test 4.
         init()
@@ -440,8 +439,8 @@ class TestFind(LeoUnitTest):
         x.do_change_then_find(settings)
 
     def test_replace_then_find_in_headline(self):
-        settings, x = self.settings, self.x
-        p = settings.p
+        c, settings, x = self.c, self.settings, self.x
+        p = c.p
         settings.find_text = 'Node 1'
         settings.change_text = 'Node 1a'
         settings.in_headline = True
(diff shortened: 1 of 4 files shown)


edreamleo:

boltex (commented):
Looks good :)

leo/unittests/core/test_leoFind.py
```diff
@@ -440,8 +439,8 @@ def test_replace_then_find_regex(self):
         x.do_change_then_find(settings)
 
     def test_replace_then_find_in_headline(self):
-        settings, x = self.settings, self.x
-        p = settings.p
+        c, settings, x = self.c, self.settings, self.x
+        p = c.p
```
boltex, line 443: c and p are not required at all. 

p is redefined below before usage, and c is thus not used.

edreamleo:
@boltex Good catch. Rev 9aa02dcc6 fixes this quirp. It retains `c` and uses it instead of `self.c` later.

PR #4698, opened 2026-05-21
PR fixes #4685: Add up/down arrows to search text widget

### LINKED ISSUE #4685: The find and replace text boxes should support arrow up/down similarly to the command palette/terminal.
See PR #4698. Note: the PR only applies to minibuffer-based finds. That should be a good model for LeoJS.

Many editors and other software that have search box text input support pressing **up arrow** to get back the previous searches text in the input. (similar to the behavior of a terminal prompt or the command palette in Leo) 

I often try to press up to get back a previous search in Leo as a reflex from doing it in other software, hehe :smile:

That concept came to mind additionally much more often to me after testing out the latest abbreviations proposed changes in the #4675 branch which set the find text input programmatically, for which I often wanted to get back my previous search text by instinctively pressing the 'up' arrow.

So I thought I'd open an (low priority) issue to suggest this relatively straightforward feature be added.

edreamleo:
See #4685.

Here, **arrow keys** refer to the `Up` and `Down` keys only.

### Part 1: Support arrow keys when using the minibuffer for finds

That is, when `@bool minibuffer-find-mode` is `True`.

- [x] `k.doMode`: Call `k.getArg` for arrow keys.
- [x] `k.getArg`: Add special case for arrow keys.
- [x] Add `LeoFind.prevSearches` ivar.
- [x] Add `LeoFind. _remember_settings` method. Call it whenever calling `LeoFind.do_find_next`.
- [x] Add `LeoFind.do_arrow`: Handle arrow keys when getting search arguments.
- [x] Show all settings, *including* change text, after arrow keys.
- [x] Note: `tab` works properly (shows change text) after arrow keys
- [x] Make arrow work after tab shifts to replacement text.

### Part 2: Support arrow keys in the Find Tab/Dialog

That is, when `@bool minibuffer-find-mode` is `False`.

- [x] `EventWrapper.keyPress`: Add a hook to `LeoFind.do_arrow`.
- [x] Stay in Find Tab/Dialog after executing `toggle-find-x` commands.
- [x] Remove the "help" column. This reduces the necessary screen real estate!
- [x] The `show-find-options` *always* shows the Find Tab.  Hurray!
- [x] Remove the do-nothing `dw.create_help_row method`.

### Other changes

- [x] Improve traces in `LeoQtLog.__init__`. None of these traces are active unless `--trace=keys` is in effect.
- [x] Delete unused `QtSearchWidget` class.
- [x] Convert a comment to a docstring.
- [x] Remove several references to the dead `wrapping` option.
- [x] Delete the unused `LeoQtLog.finishCreateTab` and  `@bool auto-scroll-find-tab`.
- [x] Cleanups, including converting empty comment lines to blank lines.  An ongoing project.

### Testing

The following hand tests pass

- [x] Test legacy operation of Alt-Ctrl bindings to change search settings.
- [x] Test operation of up/down arrows after Ctrl-F.
- [x] Test that `tab` works properly (shows change text) after up/down arrows.
- [x] Test that arrow keys work when showing replacement text: they reset the entire search.

### DIFF
--- a/leo/config/leoSettings.leo
+++ b/leo/config/leoSettings.leo
@@ -88,6 +88,7 @@
 <v t="ekr.20190926110217.1"><vh>@string adoc-base-directory = None</vh></v>
 </v>
 <v t="ekr.20260419074741.1"><vh>Unused settings</vh>
+<v t="ekr.20141024165714.1"><vh>@bool auto-scroll-find-tab = True</vh></v>
 <v t="jlunz.20151119171553.1"><vh>@bool cursor-stay-on-paste = True</vh></v>
 <v t="tbrown.20151114110207.1"><vh>@string abbreviations-next-placeholder = ,,</vh></v>
 </v>
@@ -669,7 +670,6 @@
 </v>
 </v>
 <v t="ekr.20041119034357.20"><vh>Find/replace options</vh>
-<v t="ekr.20141024165714.1"><vh>@bool auto-scroll-find-tab = True</vh></v>
 <v t="ekr.20131119143342.20107"><vh>@bool minibuffer-find-mode = False</vh></v>
 <v t="tbrown.20151010094807.1"><vh>@bool show-find-result-in-status = True</vh></v>
 <v t="ekr.20150710065036.1"><vh>@bool preload-find-pattern = False</vh></v>
--- a/leo/core/leoFind.py
+++ b/leo/core/leoFind.py
@@ -3565,11 +3565,12 @@ class LeoFind:
                 setattr(self, key, val)
             if val and key in d:
                 options.append(d.get(key))
+        options.append(f"Change: {change_s}")
 
         # Update the gui.
         self.ftm.set_change_text(change_s)
         self.ftm.set_find_text(find_s)
-        c.k.setLabelBlue('Search: ')
+        c.k.setLabel('Search: ')
         c.k.extendLabel(find_s)
         c.frame.statusLine.put(f"Find: {' '.join(options)}")
 
--- a/leo/core/leoFind.py
+++ b/leo/core/leoFind.py
@@ -135,7 +135,6 @@ class LeoFind:
         self.iSearchStrokes: list[Stroke] = []
         self.findTextList: list = []
         self.changeTextList: list = []
-        self.prevSearches: list[g.Bunch] = []  # #4685
 
         # For find/change...
         self.find_text = ""
@@ -156,6 +155,8 @@ class LeoFind:
         self.in_headline = False
         self.match_obj: re.Match = None
         self.previous_settings: g.Bunch = None
+        self.prev_searches: list[g.Bunch] = []  # #4685
+        self.prev_searches_i = 0  # #4685
         self.reverse = False
         self.root: Position = None  # The start of the search. For suboutline-only.
 
@@ -2178,15 +2179,15 @@ class LeoFind:
             return all(b1.get(z) == b2.get(z) for z in b1.keys())
 
         # Remove any previous match.
-        for bunch in self.prevSearches:
+        for bunch in self.prev_searches:
             if equal(settings, bunch):
-                self.prevSearches.remove(bunch)
+                self.prev_searches.remove(bunch)
                 break
 
         # Insert the new setting at the start of the list.
-        self.prevSearches.insert(0, settings)
+        self.prev_searches.insert(0, settings)
 
-        g.trace(len(self.prevSearches))  ###
+        g.trace(len(self.prev_searches))  ###
 
     # @+node:ekr.20210117143611.1: *5* find.start_search1
     def start_search1(self, event: LeoKeyEvent = None) -> None:  # pragma: no cover
@@ -3035,6 +3036,7 @@ class LeoFind:
     ) -> QTextMixin:
         """Display the result of a successful find operation."""
         c = self.c
+        g.trace('*****')
         # Set state vars.
         # Ensure progress in backwards searches.
         insert = min(pos, newpos) if self.reverse else max(pos, newpos)
@@ -3532,7 +3534,34 @@ class LeoFind:
 
     # @+node:ekr.20260521170130.1: *5* find.do_arrow
     def do_arrow(self, char: str) -> None:
-        g.trace(f"{char=}")
+        prev = self.prev_searches
+        if not prev:
+            return
+        i = self.prev_searches_i
+        self.prev_searches_i = (
+            i - 1 if char == 'Up' and i - 1 >= 0 else
+            i + 1 if char == 'Down' and i + 1 < len(prev) else
+            i
+        )  # fmt: skip
+
+        # Show compact status, like find.compute_result_status.
+        bunch = prev[self.prev_searches_i]
+        status = []
+        d = {
+            'find_text':        bunch.find_text,
+            'ignore_case':      'Ignore Case',
+            'node_only':        '[Node Only]',
+            'pattern_match':    'Regex',
+            'search_body':      'Body',
+            'search_headline':  'Head',
+            'suboutline_only':  '[Outline Only]',
+            'whole_word':       'Word',
+        }  # fmt: skip
+        for key in bunch.keys():
+            val = bunch.get(key)
+            if val and key in d:
+                status.append(d.get(key))
+        g.trace(status)
 
     # @+node:ekr.20131117164142.17008: *4* find.updateChange/FindList
     def update_change_list(self, s: str) -> None:  # pragma: no cover (cmd)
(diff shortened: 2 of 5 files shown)


edreamleo:
@boltex This PR is ready for review! How did we ever live without this feature?

~~Imo, as separate PR should remove support for the search dialog and the search tab in log pane.~~
In the past, people have objected to minibuffer-based searches, but I think it's time.
What do you think?

On second thought, let's leave well enough alone.

boltex:
@edreamleo  Maybe its not ready yet?
 I've tried a few different searches, pressing enter (and or F3 to make sure I matched and found text) then changed that search text in the find input box, preeed enter and F3 a few times, repeated that with different text a few times...  then I placed my cursor in the find input  box (felt a small 'lag') and pressed up arrow, but noting happened.

edreamleo:
@boltex My apologies for the confusion. I've changed my goals after spending several hours reviewing the options. Imo, this PR should leave the Find Tab and Find Dialog completely unchanged.

Leo's find-related gui needs an overhaul. #4700 contains some preliminary ideas.
These ideas may be controversial, so I have assigned this issue to Leo 6.8.10.

I plan no further work on this PR. Please approve it unless there are issues with the way minibuffer-based finds work.

edreamleo:
@boltex Please test with `@bool minibuffer-find-mode = True`.

boltex:
@tbpassin I also agree that this issue was meant to apply only to the find / replace text input boxes of the find pane, not the nav pane. (Just to be clear about that find-pane / nav-pane distinction)

@edreamleo Oh! at first i didnt get what all that minibuffer stuff was all about, but after trying it with that **@bool minibuffer-find-mode = True** setting and pressing **ctrl+F** a few times, I now see what you mean by 'minibuffer search' : ctrl+f now brings me at the bottom of the screen instead of in the find panel! hoh!

Honestly, you kind-of have to know its going to happen otherwise you 'll miss it! 😄 took me a while to actually see that because it does not 'flip' to the find pane anymore, and the cursor at the very bottom of the screen was not apparent (nor expected).

I'm not used to using that bottom bar to search for something in leo hehe! Anyways, I gave it a shot and I tried playing a bit with it:

- I noticed that unmatched searches do not add to the search history. it should. (pressing enter or F3/F2 should add to search history if not already in list, not success matches per say)

- I also noticed that the up/down direction is reversed: I tried 'up' to get the last one(s) but I saw that I had to press 'down' instead. and vice-versa.

- Most importantly, I wondered why this could that not have been done by detecting the up/down arrow press in the textbox of the find panel directly instead? I feel somewhat that this PR addresses a completely different issue by completely revamping the effect of 'CTRL+F' and ditches the find panel usage entirely...

My original issue #4685 was about just having the up/down arrow in the find-pane text box replace its content with the last/previous text it had from a 'search history', instead of just having no effect. 

Just to clarify : This would add functionality to something which had no effect (pressing up in find-pane textbox), it would not disrupt the workflow of anywone for whichever way they were used-to, and would only add a nice-to-have effect when cursor is focused in those specific find/replace text-boxes of the find-pane.

But on the other hand, since this is specifically enabled by the setting called **minibuffer-find-mode = True** which means it enables the **find** operations to be in the minibuffer, it totally makes sense under that point of view. (i just would not make it the default) I therefore have no objection to this P.R.  - other than the 2 quirps noted above.

edreamleo:

- I'll make sure unmatched searches add to the search history.
- The meaning of the arrows could be reversed, but I think their operation is correct.
- Yes, Leo's Find pane/dialog should support up/down arrows.
  This PR should not change Leo's gui at all.
  Convincing Leonistas that they should use mini-buffer-based finds is likely a fool's errand ;-)

edreamleo:
@boltex @tbpassin This PR is ready for review.  I'm more than happy with the result: 

- The changes needed to work in Find Tab/Dialog were much smaller than I had feared.
- It supports up/down arrows in the Find Tab/Dialog smoothly.
  After lots of work, `LeoFind._remember_settings` is dead simple and it seems to work well with `LeoFind.do_arrow`.
- The `toggle-find-x` commands keep focus in the Find Tab.

tbpassin (commented):
I have looked through the code and tested the revised Find tab. Everything I tried worked, and the state of the boxes and radio buttons gets restored along with the search term.

I have two reservations, though:

1. The Replace command keystroke reminders are not present anymore. I don't remember which key strokes do which replace operations, and I think they should be restored to the Find tab in some form.

2. Even though the "Entire Outline" box is checked, the search only starts from the current outline position. When I make a change to the search phrase and the "Entire Outline" box is checked, I expect the entire outline to be checked but this does not happen. I believe this is the way the tab worked before but I think it should be changed to search the entire outline after a change is made to the search phrase.

edreamleo:

To remind yourself of key bindings, do `<alt-x>replace<tab>`. The Completion tab will then show:
```
Ctrl+=   replace
         replace-all
         replace-current-character
Ctrl+-   replace-then-find
```

The PR does not change the behavior of "entire outline" radio button which merely undoes the effect of "suboutline only" or "node only" radio buttons. The name might not be perfectly clear, but I do not wish to change the radio button's effect.

boltex:
@tbpassin @edreamleo I recomment we keep the current behavior: Searching from cursor position (_even_) when 'entire outline' is checked. 

Also after just changing the search text.  

 I often place my focus on the node after which I want to find the next match, because I already know about the ones near the top, or whichever else I want to avoid - not neccessarily a suboutline, but the whole 'rest of the entire outline' ... I then See if I got something past that point anywhere else in the outline, then refine my search, press f3 to check again. etc. 
 
 I suspect its the default most people expect. 
 
 (I'm not against making a boolean setting to alter that default)

boltex:
@edreamleo Wow its even in the real find pane now?! That's great! Thanks for this work! :)

 I've tested it - Here are the two things that I came across:

1- I've noticed that it randomly adds to the history, I do ctrl+f, type something, find some matches, repeat that with some other string, I do that 3 or 4 times and when I use the up/down arrows, about half the strings are missing. (I did not repeat strings - in case you dont re-add strings) so that felt like a bug.

Let me know if you can reproduce that behavior or not - I'll share my screen on zoom if you'd like no problem.

2- Also, sorry to come back with this, hehe, ...but when using the leo minibuffer, (alt+x and then 'up' once or twice ro redo some recent command), or, the terminal on any OS'es, pressing **up** always goes **back** in history.  So having to press 'down' instead, uninstinctively feels a bit wrong. I had to re-try often thinking it did not work at all. 

Many other software that have permanent /static text input fields in their U.I. also offer a 'quick history' access with the up-previous/down-next arrows. This convention seems well established not only in those more recent softwares, but also as I've stated above, in all command line terminals on lin*x/windows cmd, etc... and moreover, leo's minibuffer. 

The logic behind this is that when looking at a list, we 'read/process' going down, and 'up' is the past/history. So I think the majority of the users would instinctively press 'up' to see the last search term(s) they've used replace the content of the find box.

edreamleo:
@boltex Thanks for your comments!

- I'll reverse the meaning of the Up and Down keys. It's trivial to do.
- I'm surprised that some history items appear to be missing. I'll look into it.

edreamleo:
@boltex Rev 54ea509:

- Reverses the meaning of the Up and Down keys.
- Calls _remember_settings in the lower-level helpers, which should ensure it get called more often.

Please feel free to add your own changes to this PR!

edreamleo:
@boltex Rev 7d43d71 remembers settings where they are computed, in `ftm.get_settings()`.

This trick should ensure that Leo *always* remembers settings.

boltex:
@edreamleo 

Thanks for those changes! :) 

Your last modifications have now fixed the missing strings from my last tests. Great! They also fixed the direction (up to get past ones, down for opposite direction) 

The only thing that felt wrong is that 'up' only gave the last one, the rest of the find history was in the 'down', as if the items were inserted at end of list while the pointer stayed at 1. 

So I fixed directly in that branch as you suggested with a simple commit. 

I have now also tested and experimented with it to make sure it works as expected. 

To make sure I didnt break anything receltly done with the abbreviations which now use the find panel: I also tested in cunjunction with using abbreviations like **html;;** and cycling with F3 after replacing each placeholder: also works great!

That UI behavior now feels top-notch ! 😄 

Thanks again!

boltex:
**Note:** The subtle detail in comparison with undo beads, is that undo beads get added when a new command is done by the user, and that becomes automatically the last bead.  (and deletes anything that was after to clear the redo-history past that point)

While in contrast, Find-setting-beads get inserted in place and do not delete all beads after them like undo-beads do when they get added.

boltex (approved):
That feature is a super cool 'nice to have' for the user interface!

edreamleo:
@boltex Many thanks for your latest tweaks!

PR #4720, opened 2026-05-31
PR: Allow qt gui in bridge

edreamleo:
Per [this](https://groups.google.com/g/leo-editor/c/CeRgm9v0ESI/m/9xhgUuXrBAAJ) post.

The following external python test file works for me:
```python
import leo.core.leoBridge as leoBridge

path =  r'C:\Users\Dev\ekr.leo'
controller = leoBridge.controller(
    gui='qt',
    loadPlugins=False,  # True: attempt to load plugins.
    readSettings=True,  # True: read standard settings files.
    silent=True,        # True: don't print signon messages.
    verbose=True)       # True: print informational messages.
g = controller.globals()
c = controller.openLeoFile(path)
g.app.gui.runMainLoop()  # Does not return and does not throw exception.
```
The output is:
```console
bridge.initLeo: 0.77 sec.
bridge.open:    0.83 sec. ekr.leo
```

### DIFF
--- a/leo/core/leoBridge.py
+++ b/leo/core/leoBridge.py
@@ -249,12 +249,12 @@ class BridgeController:
     # @+node:ekr.20070227095743: *4* bridge.createGui
     def createGui(self) -> None:
         g = self.g
-        if self.guiName == 'nullGui':
+        if self.guiName in ('qt', 'nullGui'):
             g.app.gui = g.app.nullGui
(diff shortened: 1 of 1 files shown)


boltex (approved):
Could be useful - no harm if not :)

edreamleo:
@boltex Thanks for your approval. This PR will be part of Leo 6.8.10.

PR #4721, opened 2026-06-01
Authentication & Secure Websockets docs

boltex:
Added authentication / secure Websockets documentation to LeoDocs.leo (and fixed small comment typo in leoserver.py)


boltex (commented):
@edreamleo

 I've only changed the docs in the leoDocs file. not the exported .rst .txt and .html files which I guess get generated in the website build / export process.

edreamleo (approved):
@boltex Thanks for this. Yes, changing LeoDocs.leo is the best way to update the docs. In particular, it's the easy way for me :-)

PR #4732, opened 2026-06-06
PR fixes #4729: error messages

### LINKED ISSUE #4729: Provide better message in log pane for 'Orphan node' and 'Directory not found' error messages when writing external files fails.
See PR #4732.

Some new users may be startled by such messages which could be made much clearer by adding a single short line to those two common "_newbie mistakes_".

  _(maybe only show that additional line of text the very first time such errors occur per session?)_

Something in the likes of:

For **Orphan node**:
`An external file node has a parent node without section references nor @other directives for its children.`

For **Directory not found**:
`An external file node references a file with a path containing a folder that does not exist.`

Also that Orphan node error repeats the line that start with 'Orphan node' after already giving the orphan node itself, its parent but again with the parent listed as the Orphan node! (is that normal?) see screenshot below where i show the outline, and the errors upon saving: 

Image

That 3rd line "Orphan node: @Clean bonjour.txt" feels extraneous and somewhat wrong.

edreamleo:
See #4729.

- [x] Simplify `at.warnAboutOrphandAndIgnoredNodes` and improve warning.
- [x] Improve message in `at.precheck`.

### DIFF
--- a/leo/core/leoAtFile.py
+++ b/leo/core/leoAtFile.py
@@ -3240,25 +3240,19 @@ class AtFile:
     def warnAboutOrphandAndIgnoredNodes(self) -> None:  # pragma: no cover
         # Always warn, even when language=="cweb"
         at, root = self, self.root
+        message = 'The node is included neither by an @others directive nor a section reference'
         if at.errors:
             return  # No need to repeat this.
-        for p in root.self_and_subtree(copy=False):
+        if root.isAtAllNode():  # #4729.
+            return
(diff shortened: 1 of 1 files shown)


edreamleo:
@boltex Ready for review. The changes are non-trivial.

boltex (approved):
Also ported to leojs&leo-web

edreamleo:
@boltex Thanks for your quick work.

PR #4734, opened 2026-06-07
PR fixes small abbreviation quirps

### LINKED ISSUE #4724: Small abbreviation quirps
_See P.R. #4734 for this issue._

## init_place_holder_search

In init_place_holder_search, the file_only radio option is set to true along with suboutline_only. (and possible node_only) 

Those are radio buttons so only one shoud be set to true and the others to false. 

(Note that suboutline-only should be set to inverse of node_only : **not node_only**.)

---

## make_script_substitutions

replace_selection leaves the cursor at end of replaced text, which makes sense. but if scripting was enabled and it contained a script to be resolved, then the end placement of the cursor may be slighly off. To fix this use the solution I came up in leo-web (count insertion from the end of the string, and place back at same point from the end of the string)

Something similar to this:

```typescript
    public async make_script_substitutions(word: string): Promise<void> {

        if (!this.scripting_enabled) {
            return;
        }

        const c = this.c;
        const p = c.p;
        const w = this.w;

        if (!w) {
            return;
        }

        c.abbrev_subst_env['_abr'] = word;

        // Replace the contents only if they have changed!
        const ins = w.getInsertPoint() || 0;
        let new_ins = ins;
        if (this.in_head) {
            try {
                const contents = p.h;
                const end_text = contents.slice(ins);
                const new_contents = await this._substitution_helper(contents);
                if (new_contents !== contents) {
                    p.h = new_contents;
                    if (new_contents.endsWith(end_text)) {
                        new_ins = new_contents.length - end_text.length;
                    } else {
                        new_ins = Math.min(ins, p.b.length);
                    }
                }
            } finally {
                w.setInsertPoint(new_ins);
            }
        } else {
            // at this point, insertion point is at end of last 'replace_selection' call inserted,
            // so lets save all text past that point, to be able to compare after substitution 
            // and restore insertion point by looking from the end for that text.
            const contents = p.b;

            const end_text = contents.slice(ins);

            const new_contents = await this._substitution_helper(contents);
            if (new_contents !== contents) {
                p.b = new_contents;

                if (new_contents.endsWith(end_text)) {
                    const new_ins = new_contents.length - end_text.length;
                    p.setSelection(new_ins, new_contents.length);
                    w.setInsertPoint(new_ins);
                } else {
                    const new_ins = Math.min(ins, p.b.length);
                    p.setSelection(new_ins, p.b.length);
                    w.setInsertPoint(new_ins);
                }
            }
        }

    }
```

boltex:
Fixes #4724

### DIFF
--- a/leo/commands/abbrevCommands.py
+++ b/leo/commands/abbrevCommands.py
@@ -231,7 +231,7 @@ class AbbrevCommandsClass(BaseEditCommandsClass):
             in_headline     = False,
             find_text       = rf"({start_pat}.*?{end_pat})",
             change_text     = '',
-            file_only       = True,
+            file_only       = False,
             mark_changes    = False,
             mark_finds      = False,
             ignore_case     = True,
@@ -239,7 +239,7 @@ class AbbrevCommandsClass(BaseEditCommandsClass):
             pattern_match   = True,
             search_body     = True,
             search_headline = True,
-            suboutline_only = True,
+            suboutline_only = not node_only,
             whole_word      = False,
         )  # fmt: skip
         assert settings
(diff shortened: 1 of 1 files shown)


edreamleo (changes requested):
Heh. At line 259, there seems to be no need for the 'ivars' tuple. Isn't it the same as `self.ivars`?  I was responsible for this duplication.

So it should be possible to replace: `for ivar in ivars:` with `for ivar in self.ivars:`

What do you think?

boltex:
@edreamleo oh yes I see what you mean: the ivars tuple is meant to be the same as the leoFind ivars. so I now use it directly.

edreamleo (approved):

PR #4736, opened 2026-06-11
PR fixes #4735: no default for empty events

### LINKED ISSUE #4735: can't call (some?) c.editCommands.XXX() with None as event from user-defined commands
Seen in 6.8.10-devel.

I have cursor movement command macros in myLeoSettings.py which have previously been working (v6.8.9 etc). These have calls like

```
        c.editCommands.endOfLine(None)
        # ....
        c.editCommands.beginningOfLine(None)
        # ....
        c.editCommands.endOfBuffer(None)
        # ....
        c.editCommands.beginningOfBuffer(None)
```
The ```None``` parameter is 'event' in the actual calls - I was never sure what could be put there from my commands, and everything worked fine with ``None```

Because, I think, of new guard code, these calls now fail, due to eg. premature return from functions. eg 

```
# file leo/command/editcommands.py
class EditCommandsClass:
     def moveWithinLineHelper(self, event: LeoKeyEvent, spot: str, extend: bool) -> None:
        w = event.w if event else None
        if not g.isTextWrapper(w):
            return
```

I would like the calls in my commands to work, or there to be equivalents that work without ```event``` being known. I appreciate that the guard code is probably a good idea. There are probably other calls that I am not yet using which will now fail if called from user-defined commands, however.

Thanks a lot, Jon N

edreamleo:
See #4735.

- [x] Use self.c.frame.body.wrapper as a default throughout Leo.

### DIFF
--- a/leo/commands/editCommands.py
+++ b/leo/commands/editCommands.py
@@ -1030,8 +1030,7 @@ class EditCommandsClass(BaseEditCommandsClass):
     # @+node:ekr.20150514063305.222: *5* ec.findCharacterHelper
     def findCharacterHelper(self, event: LeoKeyEvent, backward: bool, extend: bool) -> None:
         """Put the cursor at the next occurrence of a character on a line."""
-        c = self.c
-        k = c.k
+        c, k = self.c, self.c.k
         self.w = event.w if event else c.frame.body.wrapper
         if not g.isTextWrapper(self.w):
             return
--- a/leo/commands/editCommands.py
+++ b/leo/commands/editCommands.py
@@ -496,7 +496,7 @@ class EditCommandsClass(BaseEditCommandsClass):
         cursor.
         """
         c, u = self.c, self.c.undoer
-        w = event.w if event else None
+        w = event.w if event else c.frame.body.wrapper
         if not g.isTextWrapper(w):
             return
 
@@ -643,7 +643,8 @@ class EditCommandsClass(BaseEditCommandsClass):
         self.tabifyHelper(event, which='untabify')
 
     def tabifyHelper(self, event: LeoKeyEvent, which: str) -> None:
-        w = event.w if event else None
+        c = self.c
+        w = event.w if event else c.frame.body.wrapper
         if not g.isTextWrapper(w):
             return
         self.beginCommand(w, undoType=which)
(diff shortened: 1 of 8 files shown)


edreamleo:
@jknGH Please test this PR.

jknGH:
Hi @edreamleo  - I have tested the changes and things look good with my commands - thank you!

I have eyeballed the diff, didn't see any issues. Quite widespread changes required, but I think the consistency is an improvement ;-)

Jon N

edreamleo:
@jknGH Thanks for your testing.

edreamleo:
@boltex This PR is ready for review. It will likely affect LeoJS.

leo/commands/editCommands.py
```diff
@@ -1025,8 +1030,9 @@ def findCharacterExtendSelection(self, event: LeoKeyEvent) -> None:
     # @+node:ekr.20150514063305.222: *5* ec.findCharacterHelper
     def findCharacterHelper(self, event: LeoKeyEvent, backward: bool, extend: bool) -> None:
         """Put the cursor at the next occurrence of a character on a line."""
-        k = self.c.k
-        self.w = event.w if event else None
+        c = self.c
+        k = c.k
```
jknGH: elsewhere you have coalesced similar lines into one - worth doing here for consistency?

```
     c, k = self.c, self.c.k
```

jknGH (commented):
one small comment - happy to approve with or without this being acted upon

leo/commands/editCommands.py
edreamleo: Fixed at rev 8a3c325.

edreamleo:
@jknGH Ready for review.

jknGH (approved):
all fine (apologies if I did something spurious just now)

edreamleo:
@jknGH No worries. Nothing spurious.

boltex:
@edreamleo Thanks, for the heads-up: I've just looked at this code - LeoJS and leo-web do not have events, (they use `const w = this.editWidget();` which defaults itself to c.frame.body.wrapper already, so the equivalent of this P.R. was already in place 😄

boltex (commented):
Looks good

edreamleo:
@boltex Thanks for your review. It's always appreciated.

boltex:
@jknGH @edreamleo I think the title is wrong: should this state it fixes #4735 ?

jknGH:
> @jknGH @edreamleo I think the title is wrong: should this state it fixes #4735 ?

Yes, I agree

[aside: Github markdown seems to use '#' shorthand for both Issues and Pull Requests - I haven't grokked this...]

edreamleo:
@boltex I've fixed the title of this PR, referenced #4735 in the first comment, and closed #4735.

PR #4737, opened 2026-06-15
Updated abbreviation documentation

boltex:
Mentions that abbreviations are enabled by default, and give a working example with date;; instead of ts;; (which is not present anymore) for the scripting-abbreviations section.


edreamleo (approved):
@boltex Thanks for this.

PR #4740, opened 2026-06-19
PR: Cleanup redrawing and recoloring code

edreamleo:
See #4743.

PR #4748 fixes a critical blunder in this PR!

This PR does not change Leo's API: All existing plugins and scripts should work as before.

- [x] Add `clear-last-colorizer-trace` command.
- [x] Improve traces produced by `--trace=coloring`.
- [x] Remove redundant call to `w.getAllText` in `selfInsertCommand`.
- [x] Remove `QTextMixin.onTextChanged` and  `w.changingText` lockout.
- [x] Suppress redraws during startup.
- [x] Remove `# pylint: disable=no-else-return` from `v.setBodyString` and `v.setHeadString`.
- [x] Remove `# type:ignore` by unifying signatures of `redraw`, `redraw_now` and `full_redraw` methods.
- [x] `c.redraw` calls `c.recolor`. Remove all calls to `c.recolor` when calling `c.redraw`.
- [x] Rename `tree.redraw` to `redraw_tree`, retaining `redraw` as an alias for compatibility.
- [x] Remove confusing delay with `--trace=drawing`.
- [x] Replace `c.redraw_after_select` with `c.redraw`.

### DIFF
--- a/leo/core/leoFrame.py
+++ b/leo/core/leoFrame.py
@@ -1056,17 +1056,16 @@ class LeoTree:
         c = self.c
         w = c.frame.body.wrapper
         s = p.v.b  # Guaranteed to be unicode.
+
         # Part 1: get the old text.
         old_s = w.getAllText()
         if p and p == old_p and s == old_s:
             return
+
         # Part 2: set the new text. This forces a recolor.
-        # Important: do this *before* setting text,
-        # so that the colorizer will have the proper c.p.
+        # Important: set c.p *before* setting text.
         c.setCurrentPosition(p)
         w.setAllText(s)
-        # This is now done after c.p has been changed.
-        # p.restoreCursorAndScroll()
 
     # @+node:ekr.20140829053801.18458: *5* 3. LeoTree.change_current_position
     def change_current_position(self, old_p: Position, p: Position) -> None:
--- a/leo/core/leoFrame.py
+++ b/leo/core/leoFrame.py
@@ -1114,7 +1114,7 @@ class LeoTree:
     def redraw_after_head_changed(self) -> None:
         self.c.redraw()
 
-    def redraw_after_select(self, p: Position = None) -> None:
+    def redraw_after_select(self, p: Position) -> None:
         self.c.redraw()
 
     # @+node:ekr.20040803072955.91: *4* LeoTree.onHeadChanged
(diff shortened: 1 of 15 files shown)


edreamleo:
@boltex @tbpassin @vivainio This PR is ready for review.

tbpassin (commented):
leoColorizer.py
---------------------

Line 1572  
`s2 = g.truncate(s[i:j], 40)  # Show only the colored string.`

I don't see why we are truncating here. Won't this produce a partially colored string?

leoCommands.py
------------------------
Line 4536 - Resolve "could", "should":

```
Leo 6.8.10: c.redraw calls this method, so could should call this
                    method only if the code *doesn't* call c.redraw.
```

edreamleo:
@tbpassin I want to truncate the output so that it fits in my console.

This truncation doesn't matter because it's the *state* of the colorizer that I really want to see.

leo/core/leoFrame.py
```diff
@@ -1724,15 +1711,18 @@ def printWidgets(self) -> None:
             w = d.get(key)
             g.pr('w', w, 'v.h:', key.headString, 's:', repr(w.s))
 
-    # @+node:ekr.20070228163350.1: *3* NullTree.redraw and scrollTo
-    def redraw(self, p: Position = None) -> None:
+    # @+node:ekr.20070228163350.1: *3* NullTree.redraw_tree and scrollTo
+    def redraw_tree(self, p: Position = None) -> Position:
```
boltex, line 1715: Does redraw_tree return a position or none ? or an optional position : this typing is not clear to the reader.

boltex (commented):
Only thing i found confusing was the typing of that line because it returns p which can be none... (is 'none' considered a position?) : 

``` python
def redraw_tree(self, p: Position = None) -> Position:
```

leo/core/leoFrame.py
tbpassin: `p` is set to `c.p` if it is None in qt_tree.py, so in practice it will always return a position. I agree this could be made more clear here.

tbpassin (approved):

leo/core/leoFrame.py
edreamleo: @boltex @tbpassin Good questions. A cff on `redraw_tree` shows this is a bit tricky:

- `redraw_tree` will return None if drawing is disabled.
- `redraw_tree` may *change* p if we are in a hoist.
- Only `c.redraw` calls `redraw_tree`, and `c.redraw` must be careful to set `p` if `redraw_tree` returns None.

Furthermore, with Leo's default mypy settings (optional None), methods may always return None instead of the stated type.

edreamleo:
@tbpassin I should have mentioned that the truncation happens in the `report` internal function, which creates traces. This function does not affect coloring in any way.

Issue #4738, opened 2026-06-16
xdg-open fails due to recent-ish changes in subprocess.Popen() shell parameters


jknGH:
I encountered a problem with opening ```file://`` URLs which I have tracked down to a recent change.

I have URLs in my leo file like:

```
file://${SEAFILE}/projects/my_project    # SEAFILE is an environment variable with a value starting '/home/jkn/'
file://home/jkn/project/my_project/myfile.txt
```
etc. Until recently I was able to CTRL-click on these and get a file explorer opening up (if the link is to a directory), or a text editor etc.

I am now seeing errors in the log pane like

```
xdg-open /home/jkn/Seafile/projects/my_project failed with exit code 1
```

I have used git bisect and tracked things down to this commit
```
jkn@tower:~/leo-editor$ git bisect bad
4bf4ed03f124306ff185196253220945562bb414 is the first bad commit
commit 4bf4ed03f124306ff185196253220945562bb414
Author: Edward K. Ream <edreamleo@gmail.com>
Date:   Wed May 27 12:30:46 2026 -0500

    Use shell=True everywhere

```

If I revert one line of this commit:

```
diff --git a/leo/core/leoGlobals.py b/leo/core/leoGlobals.py
index e0cde3c4e3..58e221f1a3 100644
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -7307,7 +7307,8 @@ def os_startfile(fname: str) -> None:
                 ree.close()
             return
         try:
-            subPopen = subprocess.Popen(['xdg-open', fname], stderr=wre, shell=True)
+            # subPopen = subprocess.Popen(['xdg-open', fname], stderr=wre, shell=True)
+            subPopen = subprocess.Popen(['xdg-open', fname], stderr=wre, shell=False)
         except Exception:
             g.es_print(f"error opening {fname!r}")
             g.es_exception()
```

Then I am able to get the expected (old) behaviour.

Note: I am running under linux - this seems to be platform-dependant code which will not be otherwise run.

I haven't particularly investigated why shell=True causes this issue, but it doesn't seem implausible...

jknGH:
PS: xdg-open seems to be a shell script ;-)

edreamleo:
Closed via PR #4741

PR #4741, opened 2026-06-19
one-line fix for issue #4738

jknGH:
see https://github.com/leo-editor/leo-editor/issues/4738: **xdg-open fails due to recent-ish changes in subprocess.Popen() shell parameters #4738**

This fix works for me under Linux


edreamleo (approved):
@jknGH Thanks for this fix!

Issue #4739, opened 2026-06-16
Allow '=' equals sign in abbreviation trigger strings (code to fix provided)


jknGH:
A bit perversely perhaps, I would like to use ```=``` equals characters as part of some abbreviation trigger strings. Currently this doesn't work, as the parsing of an abbreviation line in a ```@data abbbreviations``` node like

```
==;;=my_substitution_text
```

fails.

Below is a slight reworking of AbbrevCommandsClass.addAbbrevHelper() in commands/abbrevCommands.py to improve this

I am not terribly proud of this code; it works by substituting '\\=' characters with a Unicode character (FULLWIDTH EQUALS SIGN) before parsing the line in the @settings node, and substituting back when adding to the abbreviations dictionary. Nevertheless it does the job, even if there are some ways to make it fail. Improvements welcome.

I hope this is useful. I have put both the diff, and the new code, below. I could make a branch and PR if that would be [Edward's] preferred approach

with this change, a line in an ```@data abbreviations``` node such as

```
\=\=;;=my_substitution_text
```
works

Diff:
```
jkn@tower:~/leo-editor$ git diff
diff --git a/leo/commands/abbrevCommands.py b/leo/commands/abbrevCommands.py
index 0209ada3a2..cb51fb5b9b 100644
--- a/leo/commands/abbrevCommands.py
+++ b/leo/commands/abbrevCommands.py
@@ -449,8 +449,10 @@ class AbbrevCommandsClass(BaseEditCommandsClass):
             return
         try:
             d = self.abbrevs
-            data = s.split('=')
-            # Do *not* strip ws so the user can specify ws.
+            # crudely allow use of '=' in abbreviation definition -
+            #  substitute '\=' with Unicode FULLWIDTH EQUALS SIGN
+            data = re.sub("\\\\=", '\uff1d', s).split('=')
+            # Do *not* strip ws, so the user can specify ws. in replacement
             name = data[0].replace('\\t', '\t').replace('\\n', '\n')
             val = '='.join(data[1:])
             if val.endswith('\n'):
@@ -459,7 +461,8 @@ class AbbrevCommandsClass(BaseEditCommandsClass):
             old = d.get(name)
             if old and old != val and not g.unitTesting:
                 g.es_print(f"redefining abbreviation {name}\nfrom {old!r} to {val!r}")
-            d[name] = val
+            # add undecorated entry into dict
+            d[re.sub('\uff1d', '=', name)] = val
         except ValueError:
[...]

PR #4742, opened 2026-06-19
Allow '=' character in abbreviation trigger string

jknGH:
A fix for issue #4739.

~~This is not very beautiful - it works by temporarily substituting '\=' character sequence within an abbreviation line with an otherwise-not-expected-to-be-used character (Unicode FULLWIDTH EQUALS SIGN) for the purposes of parsing, and then changing it back when adding to the abbreviation dictionary. Better schemes are probably available, but this works and seems a reasonable improvement.~~


edreamleo:
@jknGH I'm not sure what I think about this.

@boltex @tbpassin What do you think?

jknGH:
@edreamleo I presume you mean you are not keen on my fix code?

In that case, can we consider other ways to accommodate my 'need' to allow '=' in abbreviation triggers?

If you are comfortable with allowing an equals sign to be escaped, I can probably think of a better actual implementation...

tbpassin:
I am against this change. It is a spot fix for one condition but there most likely a general problem here and it hasn't been identified clearly yet. Heck, the issue and PR don't even show the current/intended values explicitly.

It seems to me that the question is about a need to escape characters that are significant for parsing abbreviations. If that's right, then that's the level we should be thinking at.  Any change should not be specifically pointed at "=" unless that is the only possible case.

Currently the abbreviations code has virtually no documentation. There is a statement to see "apropos-abbreviations" for details but that string is not in leoPyRef.leo and I don't know where it is or if (assuming it exists) it is even current with the code. At a minimum I would like to see exactly what the expansion rules are supposed to be.

Let's not rush into this.

jknGH:
It is not a very general issue - the current format for an abbreviation settings line is implied to be

```
[trigger phrase][equals character][substitution text]
```
However the implementation (just looking for the first equals character) does not allow the use of this character in the trigger phrase, which I regard as a small limitation. Specifically, I would like to use '==;;' as a trigger phrase.

I would be happy enough to open out the discussion to "adopt a format for abbreviation setting lines that allows all characters to be used in a trigger phrase"; but I felt that the change was enough of a step in the right direction with no required changes elsewhere

jknGH:
@tbpassin : "Heck, the issue and PR don't even show the current/intended values explicitly."

I don't really know what this means. I would be happy to document the code to better capture the expected line format, if that is what you are pointing out.

There are no other expansions/replacements BTW.

tbpassin:
> Specifically, I would like to use '==;;' as a trigger phrase.

As I understand it, the current code does not require the ";;" characters. They are just a convention. some other distinctive string could be used. If we allow "==;;" as the trigger phrase, it might cause some abbreviations not to work because someone is using a different string besides ";;".  If we require "==;;", then all existing abbreviations will fail.

If we allow "==;;" as a trigger phrase, then what if someone wants to include "==", just as you now want to include "="?

If what you really want is to use an "=" as part of your trigger phrase, perhaps the "=" character could be escaped in the abbreviation definition:

"is=;;=equal" would be defined in an @data node as "is\=;;=equal".

tbpassin:
> If you are comfortable with allowing an equals sign to be escaped, I can probably think of a better actual implementation...

As I wrote above, I think escaping a non-significant equals sign in the abbreviation definition could be a reasonable way to go. I'm not necessarily opposed to replacing non-significant equals temporarily as your proposal does. I just want to make sure that the change(s) are being made at the right level of generality and that the new rules are very clear. I don't want to have to special-case one use case after another if it can be avoided.

A more general approach might be to percent-escape characters as they are in URLs. Any percent-escaped character would be treated as its non-significant bare character.

Escaping significant characters is a technique that is pretty familiar, from regexes to URLs to command lines. So that is in its favor.

jknGH:
@tbpassin I think you are misunderstanding.

My change just allows (any number of) ```=``` character to be used in the trigger string, by escaping them (you use ```\=``` instead). You are right that the double semicolon 'terminator' is a convention only. You have to type it in the definition line as well as using it, and it explicitly ends up in the abbreviations table. For instance, an entry like

```
# @settings/data abbreviations
blah;;=my replacement text
```

causes an entry in the abbreviations table

```
    "blah;;": "my replacement text"
```
all I am doing is allowing the otherwise-not -allowed character '=' to be specified, and present in the key of such an entry

```
# @settings/data abbreviations
# with new code extant
\=\=;;=my replacement text
```
causes an entry in the abbreviations table

```
    "==;;": "myreplacement text"
```

This was previously not possible. The use of any other combinations of equals characters, semicolons, etc, remain unchanged or improved.

The only limitation is that due to my crude implementation, using Unicode FULLWIDTH EQUALS SIGN in the setting line is problematic. One way to improve this would be to parse the line character by character. I couldn't find an idiomatic Python way to "find the first non-escaped equals character in a string", but suggestions welcome

boltex:
@tbpassin I also think you are misunderstanding, hehe :) 

This has nothing to do with ';;' . It's just about the equal sign beeing the sole and only separator between the intended **'trigger'** string (i.e. the 'abbreviation'), and its **'replacement'** string,  _when defined in `@data abbreviations` or `@data global-abbreviations` setting nodes_.

So as @jknGH rightfully pointed out, this precludes the equal sign to be part of a 'trigger' string. His usage of `==;;` as a trigger string is just an example that contains equal signs for the purpose of this demonstration.

I think his solution to allow for its escapement and his implementation is totally 'ok'. (he could have chosen another sequence of character other than the long equal symbol as the temporary placeholder, such as an actual string like "EQUALSIGNPLACEHOLDER"  - in case someone want to use the long equal symbol - But that's just an implementation detail.

The only drawback with his solution is that: what if someone wanted to end its trigger string with a backslash? then he would have a backslash right before the equal thus triggering this substitution workaround. 

So we would move the problem from "cannot have equal signs in the trigger string" to "cannot end the trigger string itself with an actual backslash" ... You could argue that this is objectively better I guess...

So I'm not against this proposition. But this PR should also modify the docs to mention that the equal sign can be escaped if needed as part of the "trigger string". (and that a backslash cannot end the trigger string without  triggering this) 

Trying to think outside the box here : What if that special rule that @jknGH proposes be only applicable if something else allows it? such that anyone with an existing trigger string ending with a backslash would not be disturbed...? Perhaps we could harness something to 'flag' that we want the equal signs to be escapable in abbreviations? 

The one obvious solution is to add a configurarion setting (normally False) available for that - and for the users to set it to True when they need equal signs in their abbreviations. This is not too bad because abbreviations are not a novice feature, and so, such a user would not be scared to change a setting to enable equal signs in his abbreviation trigger strings.

Another way to make this 'optional' that does not use a new configuration setting, would be to have the abbreviation definition actually start with an equal sign! (yes you've read that right!) because starting with an equal sign makes no sense, this condition can be used as a flag to enable the escapement feature. (We remove that starting equal sign completely, _as it was used as a flag_, and then do the implementation as @jknGH proposed for that specific case only where the definition started with an equal sign.)

**So those are my 3 propositions so far:**

1. allow @jknGH solution as-is (with perhaps a better placeholder string)
2. Use a config setting such as `@bool enable-abbrev-escape`
3. Use an special case 'starting with an equal' instead of a config setting.

@edreamleo @tbpassin @jknGH  What do you guys think ? 😄

edreamleo:
@jknGH Why do you want to allow an equal sign as part of the trigger string?

jknGH:
> @jknGH Why do you want to allow an equal sign as part of the trigger string?

- in this case, simply because I wanted an abbreviation to expand to something like "==========", as a 'dividing line'. '==;;' seems the most appropriate abbreviation for this.
- erm, why not? Having (accidentally) hit this limitation, I feel it is an itch worth scratching...

jknGH:
Brief comments to @boltex's comments, for now (I mostly agree with them)

- I would be happy to update the docs with more information on whatever we end up with
- I take the point about a possible problem with a backslash as the last character. But I guess I would see that as being an issue only if the current (unenforced) convention of using ';;' as a terminator is not followed.
- (would parsing the line with a tiny state machine be a way of getting round this?)
- I too mused about having some setting ... but I am not sure I like having some setting for such a special case as 'I am using equals signs in my abbreviations'. Better, I think, would be a setting to specify the 'splitter' character used in abbreviation definitions. But I need to read and think about this a bit more - tomorrow, perhaps.

tbpassin:
> @tbpassin I also think you are misunderstanding, hehe :)
> 
> This has nothing to do with ';;' . It's just about the equal sign beeing the sole and only separator between the intended **'trigger'** string (i.e. the 'abbreviation'), and its **'replacement'** string, _when defined in `@data abbreviations` or `@data global-abbreviations` setting nodes_.

Yes, I do understand this although at first I was not quite sure about the equals sign being the sole divider. Now I am clear on it.

This whole situation reminds me of how the statement terminator has to be changed for some database stored procedures, which may need to define SQL statements of their own including their terminator.  One starts the procedure definition with the SQL statement that changes the terminator. After the end of the definition, the terminator is changed back.  Here is an example, where the normal terminator, `;`, gets changed to `//` until until it gets changed back.:

```
DELIMITER //
CREATE FUNCTION wws_30.hasuser()
RETURNS TINYINT
DETERMINISTIC
READS SQL DATA

BEGIN
    DECLARE _hasuser TINYINT;
    SELECT COUNT(user) > 0 INTO _hasuser FROM mysql.user
        WHERE user = 'webapp' AND host = 'localhost';
    RETURN _hasuser;
END//
DELIMITER ;
```

A solution like this would work for every use case I can think of, and would have no effect on any existing abbreviation.

jknGH:
I think we are agreed that the most general question is:

How to specify a string which can be split definitively into two parts (abbreviation and expansion), whilst allowing all characters to be used in both abbreviation and expansion substrings?

Personally I think "continue to use an equals sign to split, and allow the use of equals signs elsewhere via backslash-escaping", is plenty good enough, for now at least

I do take @boltex's point that the implementation does not have to use a single 'special' character. It could use any "\_\_ESCAPE_EQUALS_SIGNIFIER_SHOULD_NEVER_OCCUR_OTHERWISE\_\_" string - this is only used at 'build abbreviation' time. Thanks for that reminder.

boltex:
@edreamleo @jknGH After thinking more about it and meditating on the question Edward asked : 

"Why do you want to allow an equal sign as part of the trigger string?" 

And at first also thinking like you "erm, well why not? an equal sign is a perfectly cromulent character!" 😉  here's an example and so on... 

I now realize the impled idea was : the specs mention that the 'first' equal sign encountered is the separator. (of course there can be anything afterwards in the expanded string) 

Thus, it came to me that the true fix to this whole issue is that **the documentation should have mentionned explicitely that equal signs cannot be part of the trigger string**. 

The point being that creating a very short trigger string _without equals_ is trivial, something like "--;;" can be used to expand into something like "===========" . But not "==;;" itself. 

So altough I applaud and commend the effort of @jknGH to creating the issue, making a P.R. and his whole contribution to Leo, I now also think that this should be considered a basic limitation of the abbreviation system that equal signs cannot be part of the trigger string.

jknGH:
Hi @boltex 
You are right that having improved documentation about the (current) limitation would have been useful; as mentioned I would be happy to improve that, whatever the outcome here.

I somewhat strongly disagree with your second stance here; IMO the limitation can and should be overcome.

- it is cognitively wrong to have to type '--;;' to get '=========='
- what if I want an abbreviation for '----------' ?

boltex:
@jknGH You are not wrong in theory, and I totally get your point. In fact I considered it when meditating about all this earlier today... 

But, it is also true that it is trivial to decide to use, let's say:
```
'-;;' (one dash) for long strings of '-' 

'--;;'  (two dashes) for long strings of '=' 
```
or some other trivially short trigger strings anyone could think of... 

(You could say it _cognitively_ works because an equal is two dashes on top of each other)  

And so, when considering everything, I felt it was better in the end to leave the behavior as-is, but to mention it explicitely in the documentation that equal signs cannot be part of the trigger string.

jknGH:
@boltex - of course something like you suggest could be used. But fixing the current limitation, which my PR mostly does (I am happy to (a) rework the decorate/undecorate string, and (b) improve the documentation) seems like a no-brainer to me.

The documentation would then not have to say "an equals sign cannot be used in the trigger string", but instead would say "an equals sign can be used in the trigger string by being escaped with a backslash".

I will augment the PR in these ways and see where that gets me ;-/

edreamleo:
@jknGH I would prefer to live with the present code. I dislike seldom-used special cases.

Renaming your abbreviation from `==;;`  to `sep;;` should suffice.

edreamleo:
@jknGH @boltex @tbpassin I've changed my mind.  Equal signs should be valid in definitions (prefix/trigger).

I suggest the following change in `addAbbrevHelper`:

Replace: `data = s.split('=')` by code that scans for the first equal sign that *isn't* preceded by a backslash.

Everything else can remain the same because `expandAbbrev` doesn't know what each prefix contains.

What do you all think?

jknGH:
Hi @edreamleo

well, I am relieved ;-). I was about to pen a comment of a different tone, following your earlier post.

I am happy to present a reworked PR that works as you suggest, including updating the documentation.

Regards, Jon N

edreamleo:
@jknGH Go for it.

tbpassin:
D'accord.

jknGH:
Just to be clear - IIUC the only difference between what @edreamleo has proposed, and my original PR code, is the implementation - a 'revised' parsing, instead of the 'decorate/split/undecorate' method I used.
Happy to go down this route, I just don't want to be surprised...

edreamleo:
@jknGH That's correct.

jknGH:
Question: after updating the documentation file ```leo/doc/_sources/tutorial-pim.html.txt.txt```, can anyone give me a quick pointer on regenerating the HTML documentation to check I have got the formatting correct? I am running under Linux ... I've found some older PRs and newsgroup postings on this, but if anyone has a one-or two-step process to pass on, that would be helpful...

edreamleo:
@jknGH 

- `git checkout gh-pages`
- open LeoDocs.leo
- click the `make-sphinx` button.
- Commit all files and push.

edreamleo:
@jknGH However, I prefer to be the only one to update Leo's docs. All you need to do is update LeoDocs.leo.

jknGH:
@edreamleo thanks. If I understand correctly your point about you maintaining the published docs, that is fine; I just want to check my work visualises as I expect...

edreamleo:
@jknGH A few new words in LeoDocs.leo would be fine, but to the first approximation nobody reads docs.
Putting the new words in LeoSettings.leo will be more effective. I suggest new comments in the following nodes:

- `@data abbreviations`
- `@data global-abbreviations`
- `@outline-data tree-abbreviations`

jknGH:
@edreamleo OK, I will bear that in mind. Since I rarely read the docs ;-) I have not followed their creation much.

I will probably tweak them a bit though, IMHO the abbreviations section could (also) be clarified a little. Thanks.

jknGH:
Hi @edreamleo - I have noticed a couple of things whilst looking at this PR:

1) I found that if (in testing) I added an ```@data abbreviations``` to my working file, then executed ```abbrev-list```, that my 'previous abbreviations' were lost. This is, I think, because in myLeoSettings.leo, my abbreviations are defined in an ```@data abbreviations``` node, not an ```@data global-abbreviations``` node.

1a) it is simple to change this in myLeoSettings.leo. But IIRC the reason that they are where they are is that ```LeoSettings.leo``` also has these as (commented-out) examples in an ```@data abbreviations``` node:

image

image

Should they be in an ```@data global-abbrevations``` node (still commented out) in the distributed LeoSettings.leo instead?

2) is the name of this node (in LeoSettings.leo) a little joke?

image

edreamleo (commented):
Why does this PR use the special unicode equal sign?

jknGH:
> Why does this PR use the special unicode equal sign?

This was my original, and mentioned as somewhat crude, choice of a 'should never happen' character sequence. Discussions were had in previous comments.

I have done some further work on this PR, not pushed yet, which amongst other things uses a different, 'should never occur' string. It needs a bit of testing before pushing - apologies for the delay, I just started a new job. This PR is not longer ready for review... I am not sure how to indicate that?

edreamleo:
@jknGH Thanks for your update. Near the bottom of this page is a "draft" button. I just clicked it ;-)

edreamleo:
@jknGH Please take your time with this PR. There is no rush.

PR #4745, opened 2026-06-21
PR fixes #4744: clean leoAPI.py

### LINKED ISSUE #4744: Remove leoAPI.py
See PR #4745.

I created leoAPI.py fairly recently, but the original ideas go back to ca. 2002.

However,  the `IconBarAPI`, `StatusLineAPI`, and `TreeAPI` no longer serve any constructive purpose:
 - These dummy classes pollute cff searches for the "real" methods.
 - mypy validates that definitions in subclasses match definitions in base classes.
 - Leo now only supports the Null and Qt guis. This isn't isn't likely to change.

Proposed changes:

- leoAPI.py: Retain the `StringTextWrapper` class.
  Moving this class elsewhere would be a substantial breaking change.
- leoAPI.py: Delete the `IconBarAPI`, `StatusLineAPI`, and `TreeAPI` classes.
- test_gui.py: remove the `TestAPIClasses` class.

edreamleo:
See #4744.

- [x] Remove all classes from `leoAPI.py` except the `StringTextWrapper` class.
- [x] Remove the `TestAPIClasses` class from `test_gui.py`.
- [x] Remove most 'must be defined' methods from various gui base classes.
  All of these methods raise NotImplementedError, which is a holdover from Leo's C++ days!
- [x] Move all deleted code to the attic.
- [x] Extra: narrow an try/except block that a previous PR introduced.


edreamleo:

This PR doesn't change Leo's API in any significant way.

tbpassin:
I'm not convinced that removing leoAPI is the best course. It seems to me that leoAPI was intended to be an abstract API that capture the essentials of what a concrete API needs to support. That is valuable knowledge that should not be scrapped. It can be hard to reverse engineer a concrete implementation and separate out the essentials from the implementation-specific details.

Most of the problems with spurious search hits can be handled by renaming the classes so they are prefixed by `abstract`. That's the approach I have used in other programs. 

There should be no need to have tests for the abstract API since it doesn't implement anything. The only concrete things in the abstract API can be constant declarations for those constants that are known to be needed by all concrete implementations, and possibly a few imports.

It may be that implementation details have crept into leoAPI over the years. If so, they can be removed since they should be overridden by concrete implementations anyway.

edreamleo:
@tbpassin Thanks for your comments. They are reasonable.

I'm happy to remove most of the classes from leoAPI.py for the following reasons:

- The Null classes in LeoFrame.py are, in effect, the minimal abstract base classes.
- The issue you raise is mostly moot: Leo is unlikely ever to have a new gui.
- flake8, mypy, and pylint will flag any API-related errors that any Leo dev (present and future) is likely to make.
- The classes this PR deletes just get in the way by polluting cffs.
- I never consult the deleted classes: the cffs are the ground truth.

PR #4748, opened 2026-06-22
PR: Fix blunder in PR #4740: disable drawing *only* during startup

edreamleo:
This PR is an urgent fix to PR #4740.  All drawing was *permanently* disable after opening a new window or file!

- [x] Disable drawing only during startup!


edreamleo:
@boltex @tbpassin I am going to merge this PR immediately: it is an urgent fix.

It should be completely safe: it deletes several assignments `g.app.disable_redraw = True`.
At worst, deleting these assignments could cause extra redraws, but tracing shows that all is well.

PR #4750, opened 2026-06-22
PR fixes #4727: honor the --select=UNL command-line arg

### LINKED ISSUE #4727: Honor the select=UNL command-line arg
See PR #4750.

Leo does not honor the --select command-line arg!

The present --help messages says that either headlines or gnxs are valid, but headlines containing spaces won't work.

Instead, the format of `--select` should be `--select=UNL`.

**Original report**

There are times I would like to have a program open a Leo outline and have a specific node selected. For example, a mind map open in Freeplane could have a node's GNX attached to a branch as a hyperlink. Clicking on the branch's hyperlink would open that Leo file in Leo with the corresponding node already selected.

If possible a running instance of Leo would be used to open the GNX rather than opening a new one.

edreamleo:
See #4727.

- [x] Honor the `--select` option. It was not functional.
- [x] Fix the `--help` message: only UNLs are valid.

### DIFF
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -2431,6 +2431,11 @@ class LoadManager:
             factory.setTabForCommander(c)
         except Exception:
             pass
+
+        # #4747: honor the --select option.
+        if unl := self.options.get('select'):
+            g.findAnyUnl(unl, c)
+
         g.app.initing = False  # "idle" hooks may now call g.app.forceShutdown.
         g.app.logInited = True
         g.app.initComplete = True
@@ -2790,7 +2795,7 @@ class LoadManager:
           --save-session        always save session data when Leo closes
           --script=PATH         execute a script and then exit
           --script-window       execute script using default gui
-          --select=ID           headline or gnx of node to select
+          --select=GNX          gnx of node to select
           --silent              disable all log messages
           --theme=NAME          use the named theme file
           --trace=LIST          add one or more strings to g.app.debug
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -2795,7 +2795,7 @@ class LoadManager:
           --save-session        always save session data when Leo closes
           --script=PATH         execute a script and then exit
           --script-window       execute script using default gui
-          --select=GNX          gnx of node to select
+          --select=UNL          unl of node to select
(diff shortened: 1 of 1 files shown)


edreamleo:

tbpassin (approved):

edreamleo:
@tbpassin Thanks for your suggestion.

boltex (commented):
Nice! :)

tbpassin:
I've already been able to open a Leo outline to the target node from a mind map being edited in Freemind. The link:

`py-leo-git.cmd --select=unl:gnx://zettel2.leo#tom.20260410002743.1  C:\tom\leo_projects\zettel_browser\zettel2.leo`

Next up, I hope: open the UNL in an already-running instance of Leo.

edreamleo:
@tbpassin 

> Next up, I hope: open the UNL in an already-running instance of Leo.

I have no idea how to do that. Do you?

tbpassin:
> I have no idea how to do that. Do you?

No, but many programs do it. In Windows, I have read that one has to ask Windows to find an existing window with the right name. Here's working code to do that:

```
def find_window_by_title(title):
    """Find or create a foreign program."""
    hwnd = ctypes.windll.user32.FindWindowW(None, title)
    if not hwnd:
        subprocess.Popen(CMDPATH, shell=False)
        # Wait while prog opens
        sleep(0.3)
        hwnd = ctypes.windll.user32.FindWindowW(None, title)

    return hwnd
```

Leo would run something like this during startup and if it found an existing instance, would send the command line to it instead of continuing to open.

How to get a message to the program that was found I don't know but I'm sure that Qt has a way. This sounds like a good subject to ask a chatbot about.

tbpassin:
According to duck.ai:

Use Qt’s local IPC: QLocalServer/QLocalSocket (works on Windows/Linux/macOS). Flow:

    First Leo instance creates a QLocalServer with a fixed app name.
    Later launches try to connect to that server:
        if connected: send their CLI args, then exit
        if not: start the server and run normally
    The running instance receives args and triggers the same “handle args” code path.

edreamleo:
@tbpassin Please create a new Leo issue for your suggestions about opening a Leo file externally.

Please include the code above in the issue.

PR #4751, opened 2026-06-23
PR: --trace=verbose enables traces for the dump-last-colorizer-trace command

edreamleo:
A recent performance optimization defines the `report` function inside `jEdit.setTag` only if `--trace-coloring` is in effect.
Alas, that optimization renders the `dump-last-colorizer-trace` command almost useless.

To have any chance of catching the intermittent docstring coloring bug, there must be a way to accumulate traces for the `dump-last-colorizer-trace` command *without* continually dumping coloring traces to the console.

This PR *silently* sets the `jedit.last_trace` ivar when `--trace=verbose` is in effect.
I have set this option in the script that launches my copy of `leoPy.leo`.


edreamleo:

boltex (commented):
Looks good

PR #4753, opened 2026-06-23
PR: remove mypy suppressions

edreamleo:
This PR is a milestone. It removes most `type:ignore` suppressions in Leo's core and core qt plugins.

**The highlight of this PR**

I have wanted to fix the following signatures for *years*:

- [x] Use `s: bytes | str` in the signatures of:
    - `g.toEncodedString` and `g.toUnicode`.
    - `v.setBodyString` and `v.setHeadString`.
    The old annotations were confusing and **wrong**, and may have confused mypy.
 
**Code changes** Marked with `#4753`.

- [x] Remove `c.import_error_nodes` ivar and related code.
- [x] Fix apparent bugs in `cpp.indent`. More testing needed.
- [x] Fix bug in `create_app` function in `leoTest2.py`.
- [x] Rewrite `fc.writeZipFile`. See https://docs.python.org/3/library/zipfile.html#zipfile-objects
- [x] Revise `g.import_module`.
- [x] Rewrite `NullLog.put`.
- [x] `viewrendered.py`: fix a potentially serious bug:'
    rename `vr.update` to `vr.update_vr` to retain base-class `update` method!

**Annotation changes**

- [x] Replace all `Optional` annotations with new `|` syntax.
- [x] Annotate all methods that might return None as `-> whatever | None:`.
    This is a precursor to setting strict-optional to True.
- [x] Replace obsolete `Tuple` annotation with `tuple`.
- [x] Fix several annotation bugs that were previously masked by `type:ignore`.

### DIFF
--- a/leo/commands/bufferCommands.py
+++ b/leo/commands/bufferCommands.py
@@ -6,7 +6,7 @@
 # @+node:ekr.20150514045750.1: ** << bufferCommands imports & annotations >>
 from __future__ import annotations
 from collections.abc import Callable
-from typing import Optional, TYPE_CHECKING
+from typing import TYPE_CHECKING
 from leo.core import leoGlobals as g
 from leo.commands.baseCommands import BaseEditCommandsClass
 
@@ -220,7 +220,7 @@ class BufferCommandsClass(BaseEditCommandsClass):
             self.names[h] = nameList
 
     # @+node:ekr.20150514045829.16: *4* findBuffer
-    def findBuffer(self, name: str) -> Optional[Position]:
+    def findBuffer(self, name: str) -> Position | None:
         v = self.vnodes.get(name)
         for p in self.c.all_unique_positions():
             if p.v == v:
--- a/leo/commands/checkerCommands.py
+++ b/leo/commands/checkerCommands.py
@@ -10,7 +10,7 @@ import pathlib
 import re
 import sys
 import time
-from typing import Optional, TYPE_CHECKING
+from typing import TYPE_CHECKING
 
 # Third-party imports.
 try:
@@ -114,7 +114,7 @@ def find_long_lines(event: LeoKeyEvent) -> None:
 
     # @+others # helper functions
     # @+node:ekr.20190609135639.1: *4* function: get_root
-    def get_root(p: Position) -> Optional[Position]:
+    def get_root(p: Position) -> Position | None:
         """Return True if p is any @<file> node."""
         for parent in p.self_and_parents():
             if parent.anyAtFileNodeName():
@@ -640,7 +640,7 @@ class PylintCommand:
 
     # @+others
     # @+node:ekr.20150514125218.11: *3* 1. pylint.run
-    def run(self, root: Position, *, last_path: str = None) -> Optional[tuple[str, Position]]:
+    def run(self, root: Position, *, last_path: str = None) -> tuple[str, Position] | None:
         """Run Pylint on all Python @<file> nodes in root's tree."""
         c = self.c
         if not lint:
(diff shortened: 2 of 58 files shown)


edreamleo:
@boltex Please review the change to `fc.writeZipFile`.

edreamleo:

As noted above, it is now feasible to consider setting mypy's strict-optional option to True.

tbpassin:
For the new `|` syntax, does the version of mypy or any other dependency need to be updated?

edreamleo:
@tbpassin Python 3.10 introduced the '|' syntax. Leo also requires Python 3.10, so all is well.

tbpassin (commented):
convertCommands.py
-----------------------------

Several places, including line 1408:
```
            def scan_d(self, kind: str) -> dict[str, list[str]]:
            def scan_d(self, kind: str) -> dict[str, str]:
```
These are not equivalent and I cannot tell by looking at the code here if the signature is supposed to be changed.

leoApp.py
---------------
Lines 150ff -
```
        self.trace_binding: Optional[str] = None  # The name of a binding to trace, or None.
        self.trace_setting: Optional[str] = None  # The name of a setting to trace, or None.
        self.trace_binding: str = None  # The name of a binding to trace, or None.
        self.trace_setting: str = None  # The name of a setting to trace, or None.
```
`:str = None` should be `:str | None = None` (two places)

Line 164 is similar:
` self.log_listener: Popen = None` should read `Popen | None = None`

Lines 212 and 214 should be corrected in a similar way.

leoChapters.py
-------------------

Line 43 `self.selectedChapter: Chapter = None` should be `self.selectedChapter: Chapter | None = None`



leoCommands.py
----------------------
Line 255 - `self.nodeConflictFileName: str = None` should be `self.nodeConflictFileName: str | none = None`

leoConfig.py
--------------------
Line s 1290 -
`self.enabledPluginsFileName: str = None` should be `self.enabledPluginsFileName: str | None = None`.

leoFileCommands.py
--------------------
Line 339 -
`s: str = self.resolveUa(key, val)` should be `s: str | None = self.resolveUa(key, val)`

leoGlobals.py
--------------------

Line 4824 -
`state: str = None,` should be `state: str | None = None,`.

Line 5818 -
`def toUnicode(s: bytes | str, encoding: str = None, reportErrors: bool = False) -> str:`

Should be 
`def toUnicode(s: bytes | str, encoding: str | None = None, reportErrors: bool = False) -> str:`

leoGui.py
------------
Line 56 -
` self.script: str = None` should be ` self.script: str | None = None`.

leoserver.py
-------------
Line 5082 - `param: dict | None = d.get('param', {})` should be `param: dict = d.get('param', {})` since the expression can never return None.

General Comments
----------------------------
1. Please make future wholesale changes like these in blocks that are smaller. They don't all need to be changed at once.
2. Please do not mix unrelated kinds of changes together. It makes reviewing much harder.

edreamleo:
@tbpassin Thanks for your careful reading of the code!

You are correct, the pattern `x: y = None` should be replaced by `x: y | None = None` everywhere.

I've fixed some of those patterns while working on PR #4755, but I'll fix them all for *this* PR.

edreamleo:
@tbpassin About your general comment. Imo, it would be impractical to split this PR into smaller pieces.

However, your point is well taken. I just corrected a serious bug in the next PR (#4755) that arose from a mistake in modifying code to avoid a mypy complaint. After a bit of thinking, I have decided to mark *any* code change that isn't just an annotation change with the PR's number. I'll do that with this PR and the next.

edreamleo:
@tbpassin Today's revs mark all significant changes with `# #4753`.

I shall merge this PR soon, so as to be able to carry on with PR #4755 with minimal conflicts.

PR #4755, opened 2026-06-24
PR: use the strict_optional mypy option on six core files

edreamleo:
This PR enforces mypy's strict_optional setting on Leo's core files.
Future PRs will do the same for leo/commands and Leo's Qt gui.

- [x] Remove g.execute_shell_command_with_options. It was a total mess.
- [x] Fix all mypy complaints in the following six core files:
    - [x] `leoApp.py`.
    - [x] `leoAtFile.py`.
    - [x] `leoCommands.py`.
    - [x] `leoGlobals.py`.
    - [x] `leoKeys.py`.
    - [x] `leoNodes.py`.
- [x] Fix all mypy complaints found with `--follow-imports`:
    - [x] 'leo/plugins/mod_scripting.py',
    - [x] 'leo/plugins/qt_commands.py',
    - [x] 'leo/plugins/qt_frame.py',
    - [x] 'leo/plugins/qt_gui.py',
    - [x] 'leo/plugins/qt_idle_time.py',
    - [x] 'leo/plugins/qt_layout.py',
    - [x]  'leo/plugins/qt_text.py',
    - [x] 'leo/plugins/qt_tree.py',
    - [x] 'leo/plugins/viewrendered.py',
    - [x] 'leo/plugins/viewrendered3.py',
- [x] Rev 5b17fa6 revises `c.currentPosition()` and `c.rootPosition` so they always return Positions.
       These methods may return a position `p` with `p.v == None`, but that's no worse than returning `None`!
- [x] Fix mypy complaints in other files as necessary, but without following imports.
- [x] Use the empty string instead of None as the default return for (almost) all methods in the files above.
    Exception: `s: str | None = None` is *essential* for some text-related methods!
    Rev ad72009 adds important comments.
- [x] Resolve all of Thomas's suggestions.
- [ ] Search for `x is None:` everywhere, changing them to `not x:` if necessary.
- [ ] Search for `x is not None:` everywhere, changing them to `x:` if necessary.
- [ ] Revert code that handles command history, tab completion, and find completion.

**Testing**

 - [x]  Check all callers of methods that now return only `str` (instead of `str | None`).
- [x] Fix Control-clicking on URLs.
- [x] Check that find-string completion works as before.
- [x] Rev f6a5745c4: Fix the find commands! The culprit: rev 45b17a3e39.
- [x] Rev 164af56: Fix an improbable bug  `p._parentVnode`. It must *always* return a `Position`!
- [x] Rev ad72009: Fix a devastating text bug! The culprit: rev 4e5d2306c.
- [ ] (For EKR): Review all diffs with git-diff-pr.

### DIFF
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -429,7 +429,9 @@ def refreshFromDisk(
     at.changed_roots = []
     c.fileCommands.handleNodeConflicts()
     c.setChanged()
-    c.redraw()
+    # We may be coming from idle_check_commander and not already have p selected,
+    # if the selected position was inside the expanded & refreshed outline, it may not be valid.
+    c.redraw(update_p)  # Select a valid update_p position or p
     c.undoer.clearAndWarn('refresh-from-disk')
 
 
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -960,7 +960,7 @@ def readFileIntoNode(self: Self, event: LeoKeyEvent = None) -> None:
     if isinstance(fileName, list):
         fileName = fileName[0]
     s, e = g.readFileIntoString(fileName)
-    if s is None:
+    if not s:
         return
     g.chdir(fileName)
     s = '@nocolor\n' + s
--- a/leo/commands/editFileCommands.py
+++ b/leo/commands/editFileCommands.py
@@ -518,10 +518,10 @@ class EditFileCommandsClass(BaseEditCommandsClass):
         if not fn2:
             return
         s1, e = g.readFileIntoString(fn)
-        if s1 is None:
+        if not s1:
             return
         s2, e = g.readFileIntoString(fn2)
-        if s2 is None:
+        if not s2:
             return
         lines1, lines2 = g.splitLines(s1), g.splitLines(s2)
         aList = difflib.ndiff(lines1, lines2)
--- a/leo/commands/editFileCommands.py
+++ b/leo/commands/editFileCommands.py
@@ -1462,7 +1462,7 @@ class GitDiffController:
         filename = c.fileName()
         if not filename:
             print('git-diff: outline has no name')
-            return str
+            return ''
         directory = os.path.dirname(filename)
         if directory and not os.path.isdir(directory):
             directory = os.path.dirname(directory)
--- a/leo/commands/editFileCommands.py
+++ b/leo/commands/editFileCommands.py
@@ -1411,7 +1411,7 @@ class GitDiffController:
         return None
 
     # @+node:ekr.20170806094321.3: *4* gdc.find_git_working_directory
-    def find_git_working_directory(self, directory: str) -> str | None:
+    def find_git_working_directory(self, directory: str) -> str:
         """Return the git working directory, starting at directory."""
         while directory:
             if g.os_path_exists(g.finalize_join(directory, '.git')):
@@ -1420,7 +1420,7 @@ class GitDiffController:
             if path2 == directory:
                 break
             directory = path2
-        return None
+        return ''
 
     # @+node:ekr.20170819132219.1: *4* gdc.find_gnx
     def find_gnx(self, c: Cmdr, gnx: str) -> Position | None:
@@ -1453,7 +1453,7 @@ class GitDiffController:
         c.treeWantsFocusNow()
 
     # @+node:ekr.20210819080657.1: *4* gdc.get_parent_of_git_directory
-    def get_parent_of_git_directory(self) -> str | None:
+    def get_parent_of_git_directory(self) -> str:
         """
         #2143.
         Resolve filename to the nearest directory containing a .git directory.
@@ -1462,18 +1462,18 @@ class GitDiffController:
         filename = c.fileName()
         if not filename:
             print('git-diff: outline has no name')
-            return None
+            return str
         directory = os.path.dirname(filename)
         if directory and not os.path.isdir(directory):
             directory = os.path.dirname(directory)
         if not directory:
             print(f"git-diff: outline has no directory. filename: {filename!r}")
-            return None
+            return ''
         # Does path/../ref exist?
         base_directory = g.gitHeadPath(directory)
         if not base_directory:
             print(f"git-diff: no .git directory: {directory!r} filename: {filename!r}")
-            return None
+            return ''
         # This should guarantee that the directory contains a .git directory.
         directory = g.finalize_join(base_directory, '..', '..')
         return directory
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -148,8 +148,8 @@ class LeoApp:
         self.start_fullscreen = False  # For qt_frame plugin.
         self.start_maximized = False  # For qt_frame plugin.
         self.start_minimized = False  # For qt_frame plugin.
-        self.trace_binding: str | None = None  # The name of a binding to trace, or None.
-        self.trace_setting: str | None = None  # The name of a setting to trace, or None.
+        self.trace_binding: str = ''  # The name of a binding to trace, or None.
+        self.trace_setting: str = ''  # The name of a setting to trace, or None.
         self.translateToUpperCase = False  # Never set to True.
         self.use_splash_screen = True  # True: put up a splash screen.
 
@@ -220,7 +220,7 @@ class LeoApp:
         self.sessionManager: SessionManager | None = None
 
         # Global status vars for the Commands class...
-        self.commandName: str | None = None  # The name of the command being executed.
+        self.commandName: str = ''  # The name of the command being executed.
         self.commandInterruptFlag = False  # True: command within a command.
         # @-<< LeoApp: global controller/manager objects >>
         # @+<< LeoApp: global importer/reader/writer data >>
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -292,7 +292,7 @@ class LeoApp:
         # @+<< LeoApp: plugins and event handlers >>
         # @+node:ekr.20161028040229.1: *5* << LeoApp: plugins and event handlers >>
         self.hookError = False  # True: suppress further calls to hooks.
-        self.hookFunction = None  # Application wide hook function.
+        self.hookFunction: Callable | None = None  # Application wide hook function.
         self.idle_time_hooks_enabled = True  # True: idle-time hooks are enabled.
         # @-<< LeoApp: plugins and event handlers >>
         # @+<< LeoApp: scripting ivars >>
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -1922,14 +1922,13 @@ class LoadManager:
         Return None if testing, or in batch mode, or if the containing
         directory does not exist.
         """
-        # lm = self
+
         # Never create a workbook during unit tests or in batch mode.
         if g.unitTesting or g.app.batchMode:
-            return None
+            return ''
         fn = g.app.config.getString(setting='default_leo_file') or '~/.leo/workbook.leo'
         fn = g.finalize(fn)
         directory = g.finalize(os.path.dirname(fn))
-
         return fn if os.path.exists(directory) else ''
 
     # @+node:ekr.20120219154958.10485: *4* LM.reportDirectories
@@ -2020,7 +2019,7 @@ class LoadManager:
         return None, None
 
     # @+node:ekr.20120223062418.10414: *4* LM.getPreviousSettings
-    def getPreviousSettings(self, fn: str) -> "PreviousSettings":
+    def getPreviousSettings(self, fn: str) -> PreviousSettings:
         """
         Return the settings in effect for fn. Typically, this involves
         pre-reading fn.
@@ -2028,6 +2027,7 @@ class LoadManager:
         lm = self
         settingsName = f"settings dict for {g.shortFileName(fn)}"
         shortcutsName = f"shortcuts dict for {g.shortFileName(fn)}"
+
         # A special case: settings in leoSettings.leo do *not* override
         # the global settings, that is, settings in myLeoSettings.leo.
         isLeoSettings = fn and g.shortFileName(fn).lower() == 'leosettings.leo'
@@ -2046,14 +2046,14 @@ class LoadManager:
             d1.setName(settingsName)
             d2.setName(shortcutsName)
             return PreviousSettings(d1, d2)
-        #
+
         # The file does not exist, or is not valid.
         # Get the settings from the globals settings dicts.
         if lm.globalSettingsDict and lm.globalBindingsDict:  # #1766.
             d1 = lm.globalSettingsDict.copy()
             d2 = lm.globalBindingsDict.copy()
         else:
-            d1 = d2 = None
+            d1 = d2 = None  # type:ignore # strict_optional
         return PreviousSettings(d1, d2)
 
     # @+node:ekr.20120214132927.10723: *4* LM.mergeShortcutsDicts & helpers
@@ -2121,9 +2121,8 @@ class LoadManager:
         Duplicates happen only if panes conflict.
         """
         # Fix bug 951921: check for duplicate shortcuts only in the new file.
-        for ks in sorted(list(d.keys())):
+        for ks, aList in sorted(list(d.items())):  # strict_optional
             duplicates, panes = [], ['all']
-            aList = d.get(ks)  # A list of bi objects.
             aList2 = [z for z in aList if not z.pane.startswith('mode')]
             if len(aList) > 1:
                 for bi in aList2:
@@ -2261,8 +2260,7 @@ class LoadManager:
     def traceSettingsDict(self, d: dict[str, str], verbose: bool = False) -> None:
         if verbose:
             print(d)
-            for key in sorted(list(d.keys())):
-                gs = d.get(key)
+            for key, gs in sorted(list(d.items())):  # strict_optional
                 print(f"{key:35} {g.shortFileName(gs.path):17} {gs.val}")
             if d:
                 print('')
@@ -3493,7 +3491,9 @@ class PreviousSettings:
 
     __slots__ = ('settingsDict', 'shortcutsDict')
 
-    def __init__(self, settingsDict: g.SettingsDict, shortcutsDict: g.SettingsDict) -> None:
+    def __init__(
+        self, settingsDict: g.SettingsDict | None, shortcutsDict: g.SettingsDict | None
+    ) -> None:
         if not shortcutsDict or not settingsDict:  # #1766: unit tests.
             lm = g.app.loadManager
             settingsDict, shortcutsDict = lm.createDefaultSettingsDicts()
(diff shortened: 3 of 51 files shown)


edreamleo:
I am closing this PR. It is too large. I'll create separate PRs for each of Leo's core files.

edreamleo:
I am reopening this PR because it will be the easiest way forward, despite the mistakes I made.

edreamleo:
@boltex @tbpassin This PR is ready for review!

edreamleo:
@boltex I'll wait for your approval before merging this PR!

leo/commands/editFileCommands.py
```diff
@@ -1462,18 +1462,18 @@ def get_parent_of_git_directory(self) -> str | None:
         filename = c.fileName()
         if not filename:
             print('git-diff: outline has no name')
-            return None
+            return str
```
boltex: that return is weird.

leo/core/leoChapters.py
```diff
@@ -294,7 +294,7 @@ def parseHeadline(self, p: Position) -> tuple[str, str]:
             if binding:
                 binding = binding.strip()
         else:
-            chapterName = binding = None
+            chapterName = binding = ''  # strict_optional
```
boltex: Why is that `# strict_optional` required here? Is binding possibly returned as None? (the typing of the return type says : tuple[str, str] )

So if binding is required to be at least empty sting and not None, then it should simply be initialized as "" before the ` if m := self.re_chapter.search(p.h):`

leo/core/leoConfig.py
```diff
@@ -1337,13 +1338,13 @@ def config_iter(self, c: Cmdr) -> Generator:
                 yield key2, val, c, letter
 
     # @+node:ekr.20041123070429: *3* gcm.canonicalizeSettingName (munge)
-    def canonicalizeSettingName(self, name: str) -> str | None:
-        if name is None:
-            return None
+    def canonicalizeSettingName(self, name: str) -> str:
+        if not name:  # strict_optional
```
boltex, line 1342: Here also I wonder why the `# strict_optional` is required? 

`name` seems to be a string all along, an if empty will be falsy in `if not name:` 

I'll wait for your comments on this and I,ll go read a bit about typing in python before continuing to review this P.R. because I need to have better grip of the python typing intricacies before being able to review this properly 😄

leo/core/leoImport.py
```diff
@@ -460,15 +460,15 @@ def outlineToWeb(self, fileName: str, webType: str) -> None:
...
         c = self.c
         self.encoding = c.getEncoding(c.p)
         for fileName in paths:
             g.setGlobalOpenDir(fileName)
             path, self.fileName = g.os_path_split(fileName)
             s, e = g.readFileIntoString(fileName, self.encoding)
-            if s is None:
-                return None
+            if not s:  # strict_optional
+                continue  # strict_optional: bug fix.
```
boltex, line 471: Is continuing the loop instead or returning intentional?

leo/core/leoImport.py
```diff
@@ -483,13 +483,12 @@ def removeSentinelsCommand(self, paths: list[str], toString: bool = False) -> st
             line = s[i:j]
             valid, junk, start_delim, end_delim, junk = at.parseLeoSentinel(line)
             if not valid:
-                if not toString:
-                    g.es("invalid @+leo sentinel in", fileName)
-                return None
+                g.es("invalid @+leo sentinel in", fileName)
+                continue  # strict_optional
```
boltex, line 487: Same here, Old code returned instead of continuing that loop intentional?

leo/core/leoNodes.py
```diff
@@ -1286,7 +1282,7 @@ def _linkAsRoot(self) -> Position:
         return p
 
     # @+node:ekr.20080416161551.212: *4* p._parentVnode
-    def _parentVnode(self) -> VNode | None:
+    def _parentVnode(self) -> VNode:
```
boltex, line 1285: This one can really return none, so why not type the return value as such?

boltex:
I'Ve left some comments and questions after reading through it once, but I'll review it more carefully in the next day or two to make sure its all ok and that i didnt miss anything.

edreamleo:
@boltex Thanks for your comments. The `# strict_optional` comments are *not* for mypy, they merely mark significant changes. As a result, all "real" changes marked with `# strict_optional` are indeed intentional. That doesn't mean they are correct!

I have just marked this PR as a draft:

- [x] Fix Leo's find commands!
- [x] Fix a real bug in `p._parentVnode`!
- [ ] Complete all the items in the testing list in the first comment of this PR.

leo/commands/editFileCommands.py
edreamleo: Ooooops!

edreamleo: Fixed at rev 50a90d0f5.

leo/core/leoChapters.py
edreamleo: As noted in a comment at the outer level, `# strict_optional` comments are only for human readers.

edreamleo: Good catch. Fixed at rev 170bd96a5.

leo/core/leoImport.py
edreamleo: @boltex Yes, the `continue` statement is intentional. It seems to fix a bug: why exit the entire group is just one file goes missing?

leo/core/leoNodes.py
edreamleo: Oops. I'm not sure why I wanted the simpler annotation. Telling the truth will have ripple effects, but I think you have uncovered a blunder on my part.

leo/core/leoConfig.py
edreamleo: @boltex My apologies for the confusion. The comment is for us devs, not mypy.

leo/core/leoImport.py
edreamleo: @boltex Yes, the continue was intentional. Again, why kill all processing if there is a problem with one file?

leo/core/leoNodes.py
edreamleo: I've added this fix to the testing list in the first comment of this PR.

edreamleo: @boltex Rev 164af56 fixes a real (highly improbable) bug in `p._parentVnode`!

The annotation `-> VNode` is now (for the first time) correct. No mypy suppression is required.

tbpassin:
In this fragment that @boltex found, why is the return value still specified to be `str | None`, when almost everywhere else is being changed to `str`?

```
In [leo/core/leoImport.py](https://github.com/leo-editor/leo-editor/pull/4755#discussion_r3491026839):

>          c = self.c
         self.encoding = c.getEncoding(c.p)
         for fileName in paths:
             g.setGlobalOpenDir(fileName)
             path, self.fileName = g.os_path_split(fileName)
             s, e = g.readFileIntoString(fileName, self.encoding)
-            if s is None:
-                return None
+            if not s:  # strict_optional
+                continue  # strict_optional: bug fix.
```

leo/commands/commanderFileCommands.py
```diff
@@ -962,7 +962,7 @@ def readFileIntoNode(self: Self, event: LeoKeyEvent = None) -> None:
     if isinstance(fileName, list):
         fileName = fileName[0]
     s, e = g.readFileIntoString(fileName)
-    if s is None:
+    if not s:
         return
```
tbpassin, line 966: I'd like to see an explicit `return None` here since that will match the type signature. I know that `None` is returned by default but better to be explicit.

leo/commands/editFileCommands.py
```diff
@@ -518,10 +518,10 @@ def diff(self, event: LeoKeyEvent = None) -> None:
         if not fn2:
             return
         s1, e = g.readFileIntoString(fn)
-        if s1 is None:
+        if not s1:
             return
```
tbpassin, line 522: Here and in several other places I'd like to see `return None` instead of the plain `return`.

leo/core/leoApp.py
```diff
@@ -3086,7 +3093,7 @@ def write(self, *args: list, **keys: dict) -> None:
                 }
                 # Handle keywords for g.pr and g.es_print.
                 d = g.doKeywordArgs(keys, d)
-                color: str = d.get('color')
+                color: str = d.get('color') or ''  # strict_optional
```
tbpassin: suggest `color: str = d.get('color', '')`

leo/commands/commanderFileCommands.py
edreamleo: @tbpassin @boltex Unless I am mistaken, the fragment you are discussing is `ic.removeSentinelsCommand`, whose signature is now:
```python
def removeSentinelsCommand(self, paths: list[str]) -> None:  # strict_optional
```
So this discussion seems moot.

leo/commands/editFileCommands.py
edreamleo: @tbpassin Returning an explicit `None` would contradict usual pythonic style, used in hundreds of places throughout Leo.

This style is perfectly safe: pylint will complain is if we try to return non-None from a method that is declared to return None.

leo/core/leoApp.py
edreamleo: @tbpassin Good catch. Fixed at rev 4a469048

leo/commands/editFileCommands.py
tbpassin: > This style is perfectly safe: pylint will complain is if we try to return non-None from a method that is declared to return None.

I only asked to use `return None` instead of `return`. I know they have the same result but `return None` makes it explicit so one can see directly that the method signature is followed.

leo/core/leoCommands.py
```diff
@@ -15,7 +15,8 @@
 import tempfile
 import time
 import tokenize
-from typing import Any, Generator, Iterable, Sequence, TYPE_CHECKING
+import typing
```
tbpassin, line 18: Can't we just:
`from typing import no_type_check`

And use the decorator

`@no_type_check`

This full  import is being made in several modules but so far I've only seen this one decorator method in use.

leo/core/leoCommands.py
```diff
@@ -891,7 +943,7 @@ def getExeKind(ext: str) -> str:
             If there is a language directive in effect, return it,
             otherwise use the file extension.
             """
-            return c.getLanguage(c.p) or LANGUAGE_EXTENSION_MAP.get(ext, None)
+            return c.getLanguage(c.p) or LANGUAGE_EXTENSION_MAP.get(ext, None) or ''
```
tbpassin: Change `get(ext, None) or ''` to `get(ext, '')`

leo/core/leoCommands.py
```diff
@@ -1137,11 +1189,11 @@ def runfile(fullpath: str, processor: str, terminal: str) -> None:
             # Check terminal from MAP_SETTING_NODE setting
             setting_terminal = terminal
             if setting_terminal:
-                terminal = which(terminal)
+                terminal = which(terminal) or ''
                 if not terminal:
                     g.es(f'Cannot find terminal specified in setting: {setting_terminal}')
                     g.es('Trying an alternative')
-                    terminal = getTerminal()
+                    terminal = getTerminal() or ''
```
tbpassin: `getTerminal()` already returns a string or empty string for all returns.

leo/core/leoCommands.py
```diff
@@ -1591,7 +1643,7 @@ def currentPosition(self) -> Position:
         c = self
         if getattr(c, '_currentPosition', None):
             # *Always* return a copy.
-            return c._currentPosition.copy()
+            return c._currentPosition.copy()  # type:ignore
```
tbpassin:
```
        if getattr(c, '_currentPosition', None):
            # *Always* return a copy.
            return c._currentPosition.copy()  # type:ignore
```
This can return `None` so the method signature should be 

`def currentPosition(self) -> Position | None:`

leo/core/leoCommands.py
```diff
@@ -1591,7 +1643,7 @@ def currentPosition(self) -> Position:
...
         # Returns a new copy of the root position or None
```
tbpassin: The comment says one thing but the line of code says something else:

```
# Returns a new copy of the root position or None
return c.rootPosition()
```

leo/core/leoCommands.py
```diff
@@ -2027,13 +2081,13 @@ def dumpPosition(self, p: Position) -> None:
     # @+node:ekr.20040803140033.2: *5* c.rootPosition
     _rootCount = 0
 
-    def rootPosition(self) -> Position | None:
+    def rootPosition(self) -> Position:
         """Return a new *copy* of the root position or None."""
```
tbpassin: `"""Return a new *copy* of the root position or None."""`

The signature says that `None` can not be returned. So does the final line in the method. Please correct this docstring.

leo/core/leoCommands.py
```diff
@@ -2360,7 +2414,7 @@ def new_gnx(v: VNode) -> None:
                 new_gnx(v)
                 g.es_print(f"empty v.fileIndex: {v} new: {p.v.gnx!r}", color='red')
         for gnx in sorted(vnode_d.keys()):
-            aList = list(vnode_d.get(gnx))
+            aList = list(vnode_d.get(gnx, []))
             ids = [id_ for (id_, v) in aList]
             if len(ids) > 1:
                 print('\nc.checkGnxs...')
```
tbpassin: This print command should say why it has been emitted, or it should not be there.

leo/core/leoCommands.py
```diff
@@ -3974,10 +4030,12 @@ def scan(fileName: str) -> None:
...
                     if fileName not in todo and fileName not in scanned:
-                        c2 = g.openWithFileName(fileName, gui=gui)
-                        todo.append(fileName)
-                        assert c2 not in result
-                        result.append(c2)
+                        if c2 := g.openWithFileName(
+                            fileName, gui=gui
+                        ):  # strict_optional # Bug fix!
+                            todo.append(fileName)
+                            assert c2 not in result
```
tbpassin: Why is this an `assert` instead of an `if`?

tbpassin (commented):
I'm about halfway through a review but I don't see any point in continuing since the code keeps changing while I'm doing the review. When it stabilizes and stops changing I'll see what I can do.

edreamleo:
@tbpassin You are quite right to hold off further reviews until the code stabilizes. Maybe tomorrow.

- Done: I'll comment on all all your reviews tomorrow. They all seem reasonable at first glance.

leo/core/leoCommands.py
edreamleo: @tbpassin `import typing` occurs in four files, and all four files use the `@typing.no_type_check`.

I'm pretty sure flake8 would complain if `typing` was imported but not used.

leo/commands/editFileCommands.py
edreamleo: I think it's safe not to worry about this matter.

leo/core/leoCommands.py
edreamleo: Good catch. Fixed at rev 9f029b7.

leo/core/leoCommands.py
edreamleo: @tbpassin Another good catch. Rev 5add823 simplifies `getTerminal`. What do you think?

leo/core/leoCommands.py
edreamleo: @tbpassin I agree that the signatures of `c.currentPosition` and `c.rootPosition` are incorrect.

However, I'd like to declare that fixing them is beyond the scope of this PR.

It's possible that throwing an exception (with an assert?) may be the proper fix, but let's leave such experiments for later.

leo/core/leoCommands.py
edreamleo: @tbpassin You aren't wrong. Rev db38873 improves the comment. That's all I want to do right now.

As with your related comment, changing either the code or the annotation is beyond the scope of this PR.

leo/core/leoCommands.py
edreamleo: Done at rev eddf05c.

leo/core/leoCommands.py
edreamleo: @tbpassin Good catch. Fixed at rev 2af809d.

leo/core/leoCommands.py
edreamleo: @tbpassin I'm not sure I understand your comment. Rev 6d7a63f seems like a good enough fix.

leo/core/leoCommands.py
edreamleo: I'll mark this comment as resolved. Please unresolve it if you disagree.

edreamleo:
@tbpassin @boltex I've changed my mind.

Rev 5b17fa6 revises `c.currentPosition()` and `c.rootPosition` so they always return Positions.
These methods may return a position `p` with `p.v == None`, but that's no worse than returning `None`!

tbpassin:
> Rev [5b17fa6](https://github.com/leo-editor/leo-editor/commit/5b17fa61cf75e351ce6dc514ff119e24f50f77e5) revises `c.currentPosition()` and `c.rootPosition` so they always return Positions.
> These methods may return a position `p` with `p.v == None`, but that's no worse than returning `None`!

I have thought for a while that there should be a NullPosition that evaluates to None for boolean purposes.  Probably there should be a NullVNode too.

edreamleo:
@tbpassin Interesting idea. I'll think about it.

edreamleo:
Closed, for reasons to be discussed in a new ENB. tl;dr: This PR changes Leo's API. The next PR won't.

edreamleo:
@tbpassin Your critiques were crucial in convincing me to close this PR. Many thanks.

tbpassin:
> @tbpassin Your critiques were crucial in convincing me to close this PR. Many thanks.

: ) ... what about them influenced you?

edreamleo:
@tbpassin I'm not sure exactly. Somehow it was the cumulative effect.

This morning my subconscious was screaming at me. Talking with Rebecca helped me see the way forward.

tbpassin:
Let's remember that Leo is working well as is. There is no pressing need to make these changes. We can take plenty of time to work them out. I realize that there are probably some bugs that the effort will uncover, but they don't seem to be hurting people too noticeably. 

Changes that constrain what type a method returns should be harmless. A method the currently returns `str | None` will rarely cause problems if only ever returns `str`, unless the calling methods do something unusual with the empty string, of course. That's where I would start, making sure that each caller of the change handles the empty string correctly, just little by little.  If we don't require strict mypy scrutiny of the merged result until an entire module has been upgraded, then the risk should be very small (and the review effort smaller).

edreamleo:
@tbpassin I agree. In particular, narrowing return values, for example `-> str` instead of `-> str | None` can create a virtuous cascade of simplifications *of the annotations*. By definitions, annotation changes are safe.

I'm particularly interested in narrowing returns from `-> Position | None` to `-> Position`. The new definitions (and truthful annotations) of `c.currentPosition` and `c.rootPosition` may enable many such improvements. We shall see.

PR #4756, opened 2026-06-25
Fix setAllAncestorAtFileNodesDirty 

boltex:
Update setAllAncestorAtFileNodesDirty to pass to_do_set parameter to findAllAncestorAtFileNodes to bring back exact behavior prior to creation of show-node-files command.


boltex (commented):
This is for getting back prior behavior of that line in leoExternalFiles.py:

```python
p.v.setAllAncestorAtFileNodesDirty(to_do_set=to_do_set)
```

edreamleo (approved):
Thanks. Good catch!

Issue #4758, opened 2026-06-26
Reading (refreshing) an @auto node fails refreshing its outline


boltex:
**Note:** See P.R. https://github.com/leo-editor/leo-editor/pull/4759 

Reading @auto nodes upon opening a Leo file is ok, but external files being changed (externally) are reloaded automatically and fail to render their refreshed outline with this printed in the console when it happens:

``` 
redraw Invalid position: <pos 2250523965952 childIndex: 1 lvl: 1 key: 2250523405552:1.2250523406080:1 asdfsecond titlesdf>
redraw on_idle,on_idle,idle_check_commander,refreshFromDisk
update: @auto test.md
redraw Invalid position: <pos 2248457946432 childIndex: 1 lvl: 1 key: 2250523405552:1.2250523406080:1 asdfsecond titlesdf>
redraw call_handler,on_idle,on_idle,idle_check_commander

```

Manually collapsing and expanding its root node with the mouse does show the updated outline content was read properly. But it seems that the outline refresh process happening right after reading an @auto node has a glitch.

boltex:
To reproduce, @auto has to be already expanded and maybe do change it "one way" in leo first, then change it the "other direction" in an outside editor for leo to refresh it.

boltex:
I've fixed it in leojs, I'll fix it here to.

boltex:
Also when reporting the modified node after detecting external file(s) changes(s), it reports updating external file nodes that were not really updated, they simply were already dirty. (and the reporting logic looks for dirty external file nodes instead of the nodes that were really updated by the idle_check of the external files)

All fixed in P.R. #4759

boltex:
Note: this bug is in devel branch only, main is still ok. 

The root cause is that now the redraw method simply exits if node to be selected is non-existant. (instead of forcing selection to  c.rootPosition() like it used to do) ... So the following UI refresh gave the error in the terminal about an invalid node.

That change in the redraw method felt intentional, so my fix in #4759 does not revert this, obviously. ...I fixed it by making the refreshFromDisk method select the proper node after finishing its duty :)

PR #4759, opened 2026-06-26
Boltex/issue4758

boltex:
Fixed possible external files outline refreshing errors when selected node was inside an @file or @auto outline. (Caused by leaving c.p to be a non-existent node after refreshing from disk)

Also fixed changed nodes reporting so that they do not print out headlines of external files that just hapened to be dirty when that refreshed occured (from a prior refresh or user editing, etc. )


boltex:
See notes about the cause and this fix in https://github.com/leo-editor/leo-editor/issues/4758

edreamleo (approved):
Many thanks for this excellent work!

PR #4760, opened 2026-06-27
PR: use strict optional for leoGlobals.py

edreamleo:
PR #4755 (closed) changes too many files. I shall break up that PR into smaller pieces.

This PR changes all annotations in `leoGlobals.py` so that mypy passes it with the `strict_optional` setting.
There are 1400+ mypy complaints when testing Leo's core with this setting!
Future PRs will reduce those complaints to zero.

- [x] `leo/scripts/mypy_leo.py`: test only leoGlobals.py.
- [x] `.mypy.ini`: Set `strict_optional` to `True`. 
- [x] `leoGlobals.py`: revise all annotations so that mypy is happy.
- [x] Mark all non-trivial code changes with `# strict_optional` comments.

**Highlights of this PR**

- Change `str | None` to `str` almost everywhere in leoGlobals.py.
  In other words, the default value for *all* strings is the empty string instead of `None`.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -11,7 +11,7 @@ cache_dir = mypy_cache
 show_error_codes = True
 check_untyped_defs = True
 # mypyc: strict_optional must be true: use--strict-optional
-strict_optional = False
+strict_optional = True
 disable_error_code=attr-defined
 
 # For v0.931, per https://github.com/python/mypy/issues/11936
@@ -38,11 +38,16 @@ exclude =
 [mypy-leo.core.*]
 disallow_untyped_defs = True
 disallow_incomplete_defs = True
+follow_imports  = skip
+strict_optional = False
+### strict_optional = True ### 1053 errors.
 
 # The default leo.commands: full annotation.
 [mypy-leo.commands.*]
 disallow_untyped_defs = True
 disallow_incomplete_defs = True
+follow_imports  = skip
+strict_optional = False
 
 # All importer and writer plugins: full annotation.
 [mypy-leo.plugins.importers.*,leo.plugins.writers.*,leo.unittests.plugins.test_importers]
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -39,8 +39,7 @@ exclude =
 disallow_untyped_defs = True
 disallow_incomplete_defs = True
 follow_imports  = skip
-strict_optional = False
-### strict_optional = True ### 1053 errors.
+strict_optional = True
 
 # The default leo.commands: full annotation.
 [mypy-leo.commands.*]
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -12,7 +12,7 @@ import string
 import sys
 import textwrap
 import time
-from typing import Any, TYPE_CHECKING
+from typing import cast, Any, Callable, TYPE_CHECKING
 import zipfile
 import platform
 from leo.core import leoGlobals as g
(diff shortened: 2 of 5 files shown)


edreamleo:
@boltex @tbpassin This PR is ready for review. The `# strict_optional` comments mark non-trivial code changes.

tbpassin (commented):
leoGlobals.py
-------------------

Lines 3631 - 3632 -
```    table: tuple = (
        c.last_dir if c else None,
```

Should `c.last_dir` be a string now?

Another good catch. The value in the table doesn't matter, but rev 600664fef inits c.last_dir to `''`.

edreamleo:
> ## leoGlobals.py
> lines 825ff -
> ```
>             val = shift_d.get(s.lower())
>             if isinstance(val, str):
>                 return val
> ```

> What is supposed to happen if `val` is not a `str`?

Good catch.  Rev 30c7fdc uses a cast to kill a warning without changing the executable code:
```
 if self.mods and s.lower() in shift_d:
     # Returning '' breaks existing code.
     return cast(str, shift_d.get(s.lower()))
```

edreamleo:
> ## leoGlobals.py

> Lines 1114ff -
> 
> ```
>         if s in d:
>             return d.get(s)  # type:ignore # Yes, s is in d.
> ```
> 
> Just return `d[s]` since we know `s in d`

Mypy doesn't know that. Happily the cast in rev 0e51cff7 now solves the problem.

tbpassin:
> > ## leoGlobals.py
> > lines 825ff -
> > ```
> >             val = shift_d.get(s.lower())
> >             if isinstance(val, str):
> >                 return val
> > ```
> 
> > What is supposed to happen if `val` is not a `str`?
> 
> Good catch. Rev [30c7fdc](https://github.com/leo-editor/leo-editor/commit/30c7fdc1dc0601005eb1af974912c52dff8153be) uses a cast to kill a warning without changing the executable code:
> 
> ```
>  if self.mods and s.lower() in shift_d:
>      # Returning '' breaks existing code.
>      return cast(str, shift_d.get(s.lower()))
> ```

I don't quite understand.  The original line, ` if isinstance(val, str):` was presumably there in case `val` is not a string. Using `cast()` won't change that at runtime. What type could the value be if not a string? If you are only concerned not to return an empty string or None, how about this variant, which is easier to grasp?

```
    val = shift_d.get(s.lower, None)
    if self.mods and val not in ('', None):
        # Returning '' breaks existing code. 
        return val
```

edreamleo:
> I don't quite understand. The original line, ` if isinstance(val, str):` was presumably there in case `val` is not a string. Using `cast()` won't change that at runtime. What type could the value be if not a string? If you are only concerned not to return an empty string or None, how about this variant, which is easier to grasp?
> 
> ```
>     val = shift_d.get(s.lower, None)
>     if self.mods and val not in ('', None):
>         # Returning '' breaks existing code. 
>         return val
> ```

I wasn't clear enough. The *real* original lines were:
```python
if self.mods and s.lower() in shift_d:
      # Returning '' breaks existing code.
      return shift_d.get(s.lower())
```
These lines are completely correct, because if `s` is in `shift_d` the `s` will be (really!) a string because all values in `shift_d` are strings. Alas, the annotation `shift_d: dict[str, str]` doesn't mollify mypy because mypy thinks that `shift_d.get(s.lower()) could be None.

mypy has ways to deal with these kinds of blind spots. See mypy's [type narrowing](https://mypy.readthedocs.io/en/stable/type_narrowing.html) page, including [user-defined type guards](https://mypy.readthedocs.io/en/stable/type_narrowing.html#user-defined-type-guards). I haven't played with the latter. For now, casts seem to be enough.

edreamleo:
@tbpassin @boltex Closed in favor of PR #4755. All bugs fixed in this branch should be fixed in the `ekr-strict` branch.

PR #4762, opened 2026-06-27
PR3: test leoCommands.py with strict_optional

edreamleo:
PR 3 of many.

**Note**: This PR is based on the ekr-strict-leoNodes branch (not devel), so it only shows the most recent diffs.

It will remain a draft until all the previous PRs have been merged.

**The highlights**

This PR:
- Imports all sub-commanders as before (to break circular imports), but it adds asserts to keep pyflakes and pylint happy.
- Removes the do-nothing definitions of the corresponding ivars from `__init__` to keep mypy happy.
  This is a happy development!
- As with other PRs, uses empty strings instead of `None`.
- Uses `@typing.no_type_check` to suppress all checks in several weird methods.
  Changing them would be too dangerous here. Maybe in another PR.

### DIFF
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -4030,7 +4030,9 @@ class Commands:
                 if p.isAtLeoNode():
                     fileName = p.atLeoNodeName()
                     if fileName not in todo and fileName not in scanned:
-                        if c2 := g.openWithFileName(fileName, gui=gui):  # #4755 # Bug fix!
+                        if c2 := g.openWithFileName(
+                            fileName, gui=gui
+                        ):  # strict_optional # Bug fix!
                             todo.append(fileName)
                             assert c2 not in result
                             result.append(c2)
@@ -5342,7 +5344,7 @@ class Commands:
             c.setChanged()
             c.contractAllHeadlines()
             root.expand()
-            return root  # #4755
+            return root  # strict_optional
         if failMsg:
             g.es(failMsg, color='red')
         return None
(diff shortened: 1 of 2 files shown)


edreamleo:
@boltex @tbpassin Ready for review.

leo/core/leoCommands.py
```diff
@@ -27,48 +28,129 @@
...
     from leo.core.leoApp import PreviousSettings
+    from leo.core.leoCommands import Commands as Cmdr  # pylint: disable=import-self
     from leo.core.leoGui import LeoKeyEvent
     from leo.core.leoConfig import LocalConfigManager
     from leo.core.leoNodes import Position, VNode
 
+    # The following imports are required, but flake8 won't know that without the asserts.
+
     # 11 subcommanders...
     from leo.core.leoAtFile import AtFile
```
tbpassin, line 41: duck.ai suggests:
`from leo.core.leoAtFile import AtFile  # noqa: F401`

and either replace the comment on the assert line with  `# noqa: F401` or remove the assert entirely.

edreamleo: @duck.ai is forgetting about pyflakes. The asserts are the only way to keep pyflakes happy.

tbpassin: > @duck.ai is forgetting about pyflakes. The asserts are the only way to keep pyflakes happy.

duck.ai didn't forget - the comment only mentioned flake8 so that is what I asked about. It also suggested assigning to a dummy variable so as not to be using `assert` for things it's not intended for. E.g.,

```
from leo.core.leoAtFile import AtFile
_ = AtFile
```

edreamleo:
@tbpassin @boltex Closed in favor of PR #4755. All bugs fixed in this branch should be fixed in the `ekr-strict` branch.

Issue #4766, opened 2026-07-01
Require mypy's strict_optional option in leo/core, leo/commands and Leo's most important plugins


edreamleo:
This project will be one of the defining features of Leo 6.8.10. Félix and I agree that this project is well worth doing, despite it's complexity.

[This post](https://groups.google.com/g/leo-editor/c/AvRRKAbdCtc/m/LtD_yLd-AAAJ) discusses lessons learned from previous attempts. 

The prime directive: Leo's API shall remain unchanged in all practical aspects. Only the following will be allowed:

- Improvements to mypy_leo.py and .mypy.ini.
- More accurate annotations.
- Bug fixes uncovered by improved annotations.
- Local code changes that aid mypy and have no global effects whatever.

edreamleo:
I am reopening this issue. It should remain open until all the subsidiary PRs have been merged into Leo's "devel" branch.

edreamleo:
Fixed via various PRs.

PR #4767, opened 2026-07-01
PR one for #4766: require strict_optional in leoNodes.py

edreamleo:
See #4766.

**Summary**

- [x] .mypy.ini: require strict_optional only in leoNodes.py.
- [x] Improve leo/scripts/mypy_leo.py.
- [x] Clean up all annotations. See below.
- [x] Add asserts as needed. See below.

**Significant code changes**

- [x] c.currentPosition and c.rootPosition always return a Position, *never* None.
  However, the returned Position may *empty* (p.v == None).
  This change *not* change any previously valid code because `if p:` and `if not p:` are the *only* valid tests on positions.
- [x] Raise ValueError in p._parentVnode and v.__init__.
    Doing so prevents data loss by aborting whatever was going on before it can cause damage.
   - p._parentVnode raises ValueError when `p` has no valid parent VNode, per [this ENB](https://groups.google.com/g/leo-editor/c/iH4Rh5YifqM/m/1BlSGbLvAAAJ).
   - v.__init__ raises `ValueError` if `g.app.nodeIndices` does not exist.
- [x] Add asserts, typically `assert p.v` to guard against empty Positions.
   These asserts help mypy without risk. They simply replace imminent AttributeErrors with AssertionErrors.
- [x] Modify imports and related code as needed. These have only local effects.
 
**Significant annotation changes**

- [x] Annotate the `p.v` *ivar* as `v: VNode | None`.  This implies many new `assert p.v` statements.
- [x] p.getVisBack and p.getVisNext are annotated to return `Position | None`.
- [x] For arguments only: change `x: whatever = None` to `x: whatever | None = None`.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -39,6 +39,12 @@ exclude =
 disallow_untyped_defs = True
 disallow_incomplete_defs = True
 
+# #4766: more specific descriptions override less specific.
+[mypy-leo.core.leoNodes]
+disallow_untyped_defs = True
+disallow_incomplete_defs = True
+strict_optional = True
+
 # The default leo.commands: full annotation.
 [mypy-leo.commands.*]
 disallow_untyped_defs = True
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -1589,7 +1589,6 @@ class Commands:
         As a result, c.p.copy() is never necessary.
         """
         c = self
-        # strict_optional
         return c._currentPosition.copy() if c._currentPosition else c.rootPosition()
 
     # For compatibility with old scripts...
@@ -2025,9 +2024,12 @@ class Commands:
     _rootCount = 0
 
     def rootPosition(self) -> Position:
-        """Return a new *copy* of the root position."""
+        """
+        Return a new *copy* of the root position.
+
+        This may return an empty Position (with p.v == None) during startup.
+        """
         c = self
-        # strict_optional
         v = c.hiddenRootNode.children[0] if c.hiddenRootNode.children else cast(VNode, None)
         return Position(v=v, childIndex=0, stack=None)
 
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -2055,7 +2055,7 @@ class Commands:
         return False
 
     # @+node:ekr.20070609122713: *5* c.visLimit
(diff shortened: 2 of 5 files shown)


edreamleo:
@boltex @tbpassin This PR is ready for review.  Please take your time. There is no rush.

edreamleo:
@boltex @tbpassin I'm going to merge this PR so that we can all test it.

All comments are welcome. I'll use another PR to fix any problems.

tbpassin:
What was the rush? We just said there isn't any need to rush on this program of changes.

boltex (commented):
Great! 

This is exactly what needs to be done for improving the typing of the codebase: not changing the code's behavior, but instead providing more precise typing of the present code behavior and returned values, while adding 'asserts' to tell the compiler/linter that 'wrong' types should not arise, and be throwing errors if they did. 

## Changed Return Types

Only thing I noted is that there were two instances that changed the code's behavior and return types:

###  c.rootPosition()

After changing my mind a few times about those two while verifying with the LeoJS codebase to make sure what you commented was actually right and no shenanigans could occur, I guess it's ok. In fact it may actually improve stability. 

I say that because I verified your claim that c.rootPosition was always used with checks such as 

`if p: `

or

`if not p:`

**but that was not the case!** In commanderFileCommads.py there is c.rootPosition() usage like this which uses it without checking first if not None, and could crash when calling 'next' on None : 

```python
        root = c.rootPosition()
        if not root.next() and root.isAtEditNode():
```

(.next is in fact alias for .getNext, which could crash if root was None instead of an actual position, etc... )

### p._parentVnode()

The other modification to a return value, **_parentVnode**, will now also throw instead of returning None, which is a good thing I guess because of other instances I found such as in the **demote** method where this is done:

```python
    parent_v = p._parentVnode()
    n = p.childIndex()
    followingSibs = parent_v.children[n + 1 :]
```

which could crash is parent_v was None.

So your proposed change would actually error out also, but in a more elegant and purposeful way...  so I guess it's ok. I'll port it to LeoJS/leo-web too ! 😄 

@tbpassin and @edreamleo  let me know if you agree also with the above statements and if I didnt miss other actual changes than those two cases above

edreamleo:
@boltex Many thanks for your comments. And your approval.

Imo, the new assert statements will never crash more often than the legacy code.

As you say, the asserts are a more purposeful way to crash. More importantly, they help mypy.

PR #4769, opened 2026-07-02
Fix macOS tabbed outline shortcut dispatch

exlee:
TL;DR; Changing tabs for macOS wouldn't update commander, so under some circumstances (usually after 1-2 minutes after starting) one outline was stuck. Even if the file was closed, commander would still dispatch to it.

## Summary

Fixes macOS shortcut dispatch when multiple outlines are open in Leo's tabbed UI.

Previously, `LeoQtEventFilter` could dispatch a key event through the commander captured when the event filter was created, even if another tab was currently active. This could make shortcuts such as `save-file` act on the wrong outline.

This PR resolves the current tab's commander at key-dispatch time on macOS and uses that commander for both `masterKeyHandler` and `LeoKeyEvent` creation.

## Notes

This is a Mac-specific bug/fix. It was developed and checked on macOS.

Change itself has more type annotation changes than the code itself, though I was advised to try and add them. 

LLM contribution: Codex helped investigate the dispatch path, draft the patch, add annotations, and run local verification. Contribution took: ~2h in total (includes patch grooming).

## Checks

All tests green. `py_compile`, `flake8`, `mypy`, `pylint` ran OK.

### DIFF
--- a/leo/plugins/qt_events.py
+++ b/leo/plugins/qt_events.py
@@ -2,6 +2,8 @@
 # @+node:ekr.20140907103315.18766: * @file ../plugins/qt_events.py
 """Leo's Qt event handling code."""
 
+from __future__ import annotations
+
 # @+<< about internal bindings >>
 # @+node:ekr.20110605121601.18538: ** << about internal bindings >>
 # @@language rest
@@ -35,10 +37,14 @@
 # rules.
 # @-<< about internal bindings >>
 import sys
+from typing import Any, TYPE_CHECKING
 from leo.core import leoGlobals as g
 from leo.core import leoGui
 from leo.core.leoQt import QtCore, QtGui, QtWidgets
 from leo.core.leoQt import Key, KeyboardModifier, Type
+
+if TYPE_CHECKING:
+    from leo.core.leoCommands import Commands as Cmdr
 
 
 # @+others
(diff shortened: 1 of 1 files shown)


leo/plugins/qt_events.py
```diff
@@ -156,6 +164,18 @@ def createKeyEvent(self, event, c, w, ch, binding):
...
         )
 
+    def currentCommander(self) -> 'Cmdr':
+        """Return the current tab's commander on macOS."""
+        c = self.c
+        if sys.platform != 'darwin':
+            return c
+        try:
+            master = getattr(c.frame.top, 'leo_master', None)
+            current = master and master.currentWidget()
```
tbpassin: This line can simply be 

`current = master.currentWidget()`

since we are already in a `try` block.

leo/plugins/qt_events.py
```diff
@@ -142,9 +150,9 @@ def eventFilter(self, obj, event):
         return True  # Whatever happens, suppress all other Qt key handling.
 
     # @+node:ekr.20110605195119.16937: *4* LeoQtEventFilter.createKeyEvent
-    def createKeyEvent(self, event, c, w, ch, binding):
+    def createKeyEvent(self, event: Any, c: 'Cmdr', w: Any, ch: str, binding: str) -> leoGui.LeoKeyEvent:
```
tbpassin: Recent PRs have moved away from the `c: 'Cmdr' style annotation in favor of `c: Cmdr`.

exlee:
I've updated PR, both feedback comments addressed :)

tbpassin:
Good! I marked the PR as a bug and set the next release as a milestone. Hopefully @edreamleo will spot it now so he can take a look.

edreamleo (approved):
Thanks!

PR #4770, opened 2026-07-02
Add macOS setting to preserve Qt Control/Command key roles

exlee:
## Summary

Adds a new macOS-only Qt setting:

`@bool qt-mac-dont-swap-ctrl-and-meta = False`

When enabled, Leo calls Qt's `AA_MacDontSwapCtrlAndMeta` application attribute before creating `QApplication`. This lets users opt out of Qt's default Command/Control role swapping while preserving the existing default behavior.

## Notes

I'm probably the only one who uses Caps Lock mapped to Control on macOS, and having to use `Cmd+...` binding made me annoyed pretty quick. This PR fixes it:

- Default remains `False`, so existing behavior is unchanged.
- The setting only has effect on macOS.
- The attribute is applied during Qt GUI startup, before `QApplication` is created.

LLM Disclosure
This PR was prepared with assistance from OpenAI Codex. Codex helped identify the Qt startup location, add the setting, and run verification checks. Contribution took ~1h in total.

### DIFF
--- a/leo/config/leoSettings.leo
+++ b/leo/config/leoSettings.leo
@@ -200,6 +200,7 @@
 <v t="ekr.20210914081311.1"><vh>@bool highlight-body-line = False</vh></v>
 <v t="ekr.20210914081428.1"><vh>@string line-highlight-color = None</vh></v>
 <v t="ekr.20240726065425.1"><vh>@string qt-layout-name = legacy</vh></v>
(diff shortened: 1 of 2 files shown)


tbpassin:
If I had a Mac I would definitely want to swap the keys so I'm glad to see this PR. On Windows/Linux I swap CTRL and CAPSLOCK, and getting them working to my satisfaction is very important.. I had no idea that Qt quietly messed with the key bindings.

edreamleo (approved):
@exlee Thanks for this PR!

PR #4771, opened 2026-07-02
Add macOS Leo launcher generator

exlee:
## Description

Adds a macOS helper script for building a `Leo.app` bundle from a Leo source checkout. The generated app bundles the Leo source tree read-only, writes a minimal `Info.plist`, creates an app icon from the bundled Leo icon asset, and launches Leo through a `uv`-managed virtual environment.

Also adds the new script to `LeoPyRef.leo` as an `@file` tree and keeps the script annotated with Leo thin sentinels.

## Notes

Installer requires `uv` in order to work.

This one is more of a "fun PR" - once again macOS specific, constrained to a single file runner (and `LeoPyRef.leo` hookup) and it's very shiny :)

It started as a need of having something better than a generic file-like icon for python3.XX and later became a need to encapsulate Leo while also adding additional features (like "read-only" source bundling to prevent accidental edition of Leo's files).

Icon was generated by LLM (GPT) but I edited it to make it more icon-like. As far as I know and believe, it's by default in public domain. 

LLM disclosure: Codex helped develop the application wrapper itself and then transform it into Python generation script. Contribution time: ~5h.


tbpassin:
I have set the label and milestone. I'm not sure about this proposal because it's Mac only, and uv isn't Leo's usual packaging. Yet it seems worthy to me.

exlee:
I was thinking about this PR and thought that maybe instead of merging it into devel branch the better idea would be to have a separate macOS Leo overlay (e.g. "Leo.app")

For myself I'm doing heavy customization:
- Deeply modifying Leo's keymap (to make it similar with other outliner I've seen)
- I made a theme with a macOS system available font as a first choice and custom icons

Having "Leo.app" as a separate overlay (and packing) over original Leo would:
- allow interested ones in checking packaged version
- would keep macOS-specific customizations outside of the main development
- prevent the risk of macOS-specific code to drag after main core
- in improbable case Leo gets widely popular on macOS Leo.app could be merged (I'd use MIT license for the overlay)

Idea behind the overlay would be changing first impression from this one:

image

...to this one:

image

edreamleo (approved):

PR #4773, opened 2026-07-04
PR3: require strict_optional in leoCommands.py

edreamleo:
See #4766.

- [x] Change .mypy.ini to require strict_optional in leoCommands.py.

**Discussion**

This PR uses a new pattern for improving annotations:

- For just *one additional* file (in this PR, it's leoCommands.py) change .mypy.ini to require strict_optional checking.
- Run mypy_leo.py with files = ['leo']. mypy will check all files in the leo/core directory
  However, mypy_leo.py tells mypy *not* to follow imports, thereby preventing mypy from checking plugins, etc.
- With these settings, mypy initially complained about many lines in leoCommands.py and leoGlobals.py.
  mypy also complained about a few lines each in a few other files.
- Some annotations fixes resulted in *new* complaints. I fixed complaints until none were left.

The following changes are relatively minor:

- [x] Clean c.__init__ and its helpers. Tricky, because almost no imports are allowed at the top level.
- [x] leoNodes.py: v.fileIndex is *always* a (possibly) empty string (instead of a string or None).
- [x] Add various asserts to suppress mypy complaints.
- [x] Found a likely bug in c.canMoveOutlineUp. This is a task for a separate PR.
- [x] Recheck with `files = ['leo'] in mypy_leo.py. That generates warnings about real annotation issues.
   - [x] Fix 30+ issues in leoCommands.py.
   - [x] Fix 30+ issues in leoGlobals.py.
   - [x] Fix minor related annotations complaints in leoserver.py and leoExternalFiles.py.
   - [x] Fix minor complaints in abbrevCommands.py, leoConfig.py, leoImport.py, leoNodes.py, and leoPlugins.py.

### DIFF
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -38,108 +38,108 @@ if TYPE_CHECKING:  # pragma: no cover
 
     from leo.core.leoAtFile import AtFile
 
-    assert AtFile
+    assert AtFile is not None
 
     from leo.core.leoChapters import ChapterController
 
-    assert ChapterController
+    assert ChapterController is not None
 
     from leo.core.leoFileCommands import FileCommands
 
-    assert FileCommands
+    assert FileCommands is not None
 
     from leo.core.leoFind import LeoFind
 
-    assert LeoFind
+    assert LeoFind is not None
 
     from leo.core.leoImport import LeoImportCommands
 
-    assert LeoImportCommands
+    assert LeoImportCommands is not None
 
     from leo.core.leoKeys import KeyHandlerClass
 
-    assert KeyHandlerClass  # type:ignore
+    assert KeyHandlerClass is not None
 
     from leo.core.leoHistory import NodeHistory
 
-    assert NodeHistory
+    assert NodeHistory is not None
 
     from leo.core.leoPersistence import PersistenceDataController
 
-    assert PersistenceDataController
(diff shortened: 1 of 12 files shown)


edreamleo:
@boltex @tbpassin Ready for review. The changes in this PR are straightforward.

leo/core/leoCommands.py
```diff
@@ -111,54 +189,25 @@ class Commands:
...
+        previousSettings: PreviousSettings | None = None,
+        relativeFileName: str = '',
     ) -> None:
         t1 = time.process_time()
         c = self
+
         # Official ivars.
-        self._currentPosition: Position | None = None
+        self._currentPosition = cast(Position, None)
         self._topPosition: Position | None = None
```
tbpassin, line 201: Why does this assignment use a different pattern the recently changed line 201?

edreamleo:
@tbpassin I don't remember the details, but I suspect mypy suggested it, directly or indirectly.

The more important question is why the following two neighboring declarations should be different:
```python
self._currentPosition = cast(Position, None)
self._topPosition: Position | None = None
```
The answer is that c._currentPosition really is *always* a Position, but c._topPosition may be a Position or None.

edreamleo:
@boltex @tbpassin I shall merge this PR now so that we can all test it.

PR #4775, opened 2026-07-05
Preserve duplicate shortcuts across repeated binding rebuilds

### LINKED ISSUE #4774: Duplicate shortcut bindings can disappear after startup because repeated `makeAllBindings()` mutates `shortcutsDict`
## Summary

On macOS[^note], duplicate `@shortcuts` entries in `myLeoSettings.leo` are left unbound, doesn't matter whether conflicting or not.

The problem is not just the duplicate warning itself. The issue is that Leo mutates `c.config.shortcutsDict` while constructing bindings, so a later `makeAllBindings()` pass rebuilds from already-trimmed shortcut data and loses user bindings.

[^note]: Issue most likely exists on other platforms, too

## Reproduction

Use active `@shortcuts` entries in `myLeoSettings.leo` like these:

```text
select-all = Meta-a
select-all = Meta-a

find-next = Meta-g
find-next = Meta-g

find-prev = Meta-Shift-g
find-prev = Meta-Shift-g
```

There was also a real collision in my file:
```
open-my-leo-settings = Meta-,
de-hoist = Meta-,
```

## Observed behavior

Startup reports warnings like:
```
duplicate, (not conflicting) key bindings in myLeoSettings.leo
all    Meta+a select-all
all    Meta+a select-all
```

After a fresh startup, all duplicate Meta bindings are missing from show-bindings and do not work.
Running `reload-settings` makes them appear/work again (note: this is my modified `reload-settings` that also reload bindings, not the stock one).

In my case all binds, i.e. `Meta+a`, `Meta+g`, and `Meta+Shift+g` and `Meta+,` were affected.

## Expected behavior

Duplicate shortcut definitions should not make a valid user binding disappear during startup.

At minimum, repeated binding rebuilds should be stable:

if a binding exists after one makeAllBindings() pass, it should still exist after a second pass using the same loaded settings.

exlee:
TL;DR; Fixes issue causing duplicating bindings in `myLeoSettings.leo` to be ignored.

## Summary

Fixes #4774 by making repeated key-binding rebuilds stable when duplicate shortcut definitions are present.

Leo was mutating `c.config.shortcutsDict` while constructing bindings. That meant an early `makeAllBindings()` pass could bind a duplicate user shortcut successfully, but also trim the underlying shortcut config. A later rebuild during startup could then recreate bindings from already-damaged shortcut data and lose user overrides.

This change preserves a raw snapshot of loaded shortcut settings and makes each `makeAllBindings()` pass rebuild from that original data.

## Changes

- Preserve the original loaded shortcut settings in `LocalConfigManager`
- Rebuild bindings from the raw shortcut snapshot on each `makeAllBindings()` pass
- Add a regression test covering duplicate override shortcuts across repeated rebuilds

## User-visible effect

Duplicate or overlapping shortcut definitions no longer are ignored after a later binding rebuild during startup (`myLeoSettings.leo`).

This was observed with duplicate Meta bindings, but the root cause is in Leo's generic binding rebuild logic and is not macOS-specific.

## Notes

Observed and validated on macOS while investigating duplicate Meta shortcut bindings during startup. Interesting tidbit: static code analysis pointed that code is correct, but it didn't match actual behavior. 

## LLM disclosure

I used Codex/LLM assistance to help trace the binding rebuild path, identify the mutation bug, draft the fix, and add the regression test. Contribution time: ~2h.

### DIFF
--- a/leo/core/leoConfig.py
+++ b/leo/core/leoConfig.py
@@ -1631,6 +1631,9 @@ class LocalConfigManager:
             self.shortcutsDict = d2 = lm.globalBindingsDict
             assert d1 is None or isinstance(d1, g.SettingsDict), d1.__class__.__name__
             assert d2 is None or isinstance(d2, g.SettingsDict), d2.__class__.__name__
+        # Preserve the raw shortcut settings so repeated binding rebuilds do not
+        # depend on mutations made while resolving duplicate strokes.
+        self.shortcutsDictRaw = self.shortcutsDict.copy() if self.shortcutsDict else None
         # Default encodings.
         self.default_at_auto_file_encoding = 'utf-8'
         self.default_derived_file_encoding = 'utf-8'
--- a/leo/core/leoKeys.py
+++ b/leo/core/leoKeys.py
@@ -2373,6 +2373,12 @@ class KeyHandlerClass:
     def makeAllBindings(self) -> None:
         """Make all key bindings in all of Leo's panes."""
         k = self
+        # Rebuild from the original shortcut settings each pass. The binding
+        # resolver mutates c.config.shortcutsDict while handling duplicate
+        # strokes, so repeated startup-wide rebuilds must not reuse that
+        # already-trimmed dict as input.
+        raw_shortcuts = k.c.config.shortcutsDictRaw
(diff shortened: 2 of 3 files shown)


edreamleo (approved):
Thanks for this PR!

PR #4777, opened 2026-07-06
Fix external editor launch on POSIX

### LINKED ISSUE #4776: External editor launch fails on POSIX when using subprocess.Popen
TL;DR; When trying to use `@string external-editor = ...` setting string for Popen is joined instead of being kept as an array. Supposedly this behavior doesn't exhibit on Windows and is POSIX-related (thus macOS is affected)

## Summary

Opening a node/file with an external editor can fail on macOS/Linux when Leo uses the `subprocess.Popen` open-with path. Leo builds a shell-style command string containing both the editor and target file, then passes that single string directly to `subprocess.Popen`.

On POSIX platforms, `Popen(string)` with `shell=False` treats the whole string as the executable path. It does not split quoted command text into argv.

## Observed behavior

With an external editor configured, Leo can raise:

```text
FileNotFoundError: [Errno 2] No such file or directory: '"/path/to/editor" /path/to/file.md'
```

The important detail is that the missing filename is the entire command string, including the quoted editor path and the target file path.

## Expected behavior

Leo should launch the configured editor and pass the selected temp/external file as an argument.

On POSIX, the launch should use an argv list, for example:

```python
subprocess.Popen(["/usr/bin/editor", "/path/to/file.md"])
```

instead of:

```python
subprocess.Popen('"/usr/bin/editor" /path/to/file.md')
```

## Notes

The issue is not specific to a particular editor. It comes from passing a combined command string to `subprocess.Popen` on POSIX. Windows has historically accepted command-line strings differently, so preserving the existing Windows behavior may be safest while fixing POSIX to pass a list.

exlee:
## Summary

Fixes #4776.

This PR fixes the `subprocess.Popen` open-with path on POSIX platforms. Leo previously built a shell-style command string by joining the configured editor and the target filename, then passed that single string directly to `subprocess.Popen`.

On POSIX, `Popen(string)` with `shell=False` treats the whole string as the executable path, so a quoted editor command such as `"/usr/bin/editor" /tmp/example.md` is interpreted as one nonexistent filename.

## Changes

- Preserve the existing combined-string behavior on Windows.
- On POSIX, pass `subprocess.Popen` an argv list instead of a shell-style command string.
- Strip matching quotes from configured open-with arguments before building the POSIX argv list.
- Add a regression test for a quoted editor path and a target filename containing spaces.

## Notes

The bug was observed on macOS, but the underlying behavior is POSIX `subprocess.Popen` argument handling rather than a macOS-specific API issue.

LLM disclosure: I used OpenAI Codex to help diagnose the issue, prepare the patch, and draft this PR text. Contribution time: ~1h.

### DIFF
--- a/leo/core/leoExternalFiles.py
+++ b/leo/core/leoExternalFiles.py
@@ -462,17 +462,14 @@ class ExternalFilesController:
         try:
             # All of these must be supported because they
             # could exist in @open-with nodes.
-            command = '<no command>'
             if kind in ('os.system', 'os.startfile'):
                 c_arg = self.join(arg, fn)
+                command = f"{kind} {c_arg}"
                 if not testing:
-                    try:
-                        subprocess.Popen(c_arg)
-                    except OSError:
-                        g.es_print('c_arg', repr(c_arg))
-                        g.es_exception()
+                    subprocess.Popen(c_arg)
             elif kind == 'exec':
-                g.es_print('open-with exec no longer valid.')
+                command = 'open-with exec no longer valid.'
+                g.es_print(command)
             elif kind == 'os.spawnl':
                 filename = g.os_path_basename(arg)
                 command = f"os.spawnl({arg},{filename},{fn})"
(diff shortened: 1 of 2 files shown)


edreamleo:
@exlee cf0cc213a reformats one file. No big deal.

However, mypy reports the following:
```console
mypy_leo.py
leo\core\leoExternalFiles.py:493: error: Incompatible types in assignment (expression has type "str | list[str]", variable has type "str")  [assignment]
Found 1 error in 1 file (checked 258 source files)
```
~~I'm not sure what to do about this complaint. Please fix it.~~
Rev e286264dd attempts a fix.

edreamleo:
@exlee Rev e286264dd fixes all mypy complaints and makes other improvements.

Please retest and report your results.

edreamleo (approved):

edreamleo:
@exlee I am going to merge this PR. Please file a new PR if I made a mistake. Thanks.

PR #4779, opened 2026-07-06
PR: Require strict_optional for leoApp.py and leoConfig.py

edreamleo:
See #4766.

This PR annotates the modules corresponding to g.app and g.app.config.

### Phase 1: Annotate leoApp.py

The acid test: does Leo crash during startup?

- [x] Change .mypy.ini to require strict_optional in leoApp.py, in addition to all previously strict modules.
- [x] Fix all mypy complaints, including follow-on complaints.
- [x] Annotate all ivars of the LeoApp class as well as possible.
    Some compromises were necessary. Better annotations would create circular imports.
- [x] Annotate all functional event args as `event: LeoKeyEvent`. They are not optional.
- [x] Annotate all unused event args as `event: Any = None  # Unused`.

### Phase 2: Annotate leoConfig.py

The acid test: does Leo honor its settings?

- [x] Change .mypy.ini to require strict_optional in leoConfig.py, in addition to all previously strict modules.
- [x] Fix all mypy complaints, including follow-on complaints.

### Phase 3: Generalize Leo's *code* while restricting the *annotations*

See the motivating example in #4780 for details and a discussion of why limiting the *annotations* of strings does not break existing code.

- [x] Rev 759b118: Using the new clone-diff-pr command, in the files this PR changes, do a regex search for:
```python
if x is None:
if x is not None:
```
   replacing them with the more general (and more pythonic) code:
```python
if not x:
if x:
```
Notice that rev 759b118 contains other self-explanatory changes.
    
### Bug fixes

- [x] LM.openWithFileName and g.openWithFileName now *always* return a commander. Hurray!
- [x] Fix the save-buffers-kill-leo command.

### Extras

- [x] Tweak the new clone-diff-pr command in leoEditFileCommands.py.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -41,7 +41,7 @@ disallow_incomplete_defs = True
 strict_optional = False
 
 # #4766: more specific descriptions override less specific.
-[mypy-leo.core.leoNodes,leo.core.leoGlobals,leo.core.leoCommands]
+[mypy-leo.core.leoApp,leo.core.leoNodes,leo.core.leoGlobals,leo.core.leoCommands]
 disallow_untyped_defs = True
 disallow_incomplete_defs = True
 strict_optional = True
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -38,6 +38,7 @@ exclude =
 [mypy-leo.core.*]
 disallow_untyped_defs = True
 disallow_incomplete_defs = True
+# strict_optional = True: 1100+ errors.
 strict_optional = False
 
 # #4766: more specific descriptions override less specific.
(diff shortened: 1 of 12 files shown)


edreamleo:
@boltex @tbpassin I shall do #4780 (git-diff-pr-clone-changed) first as part of the testing for this PR.

edreamleo:
@boltex @tbpassin This PR is ready for review. I'll sleep on it at least one more night before merging.

tbpassin:
> @boltex @tbpassin This PR is ready for review. I'll sleep on it at least one more night before merging.

I'm not going to look at it until it hasn't been changed for at least a day. There is no need to be in a rush about this series of PRs.

edreamleo:
@tbpassin I understand and agree with your attitude. I welcome your comments at any time.

I am going to merge this PR today so that I can move on to the next PR.

PR #4781, opened 2026-07-07
PR fixes #4780: clone-diff-pr

### LINKED ISSUE #4780: Add clone-diff-pr command
I sometimes want to distill the tree produced by Leo's git-diff-pr command as follows:

- Omit the diffs.
- Show only clones of the changed nodes.
- Don't show added or deleted nodes.

The result will be a tree of (clones of) all changed nodes, organized by file.

### Motivating example

When looking at a PR that changes (in signatures) many annotations of a string s of the form:
```python
s: str = None
s: str | None = None
```
to
```python
s: str = ''
```
These are non-trivial changes! I must search(in the changed nodes!) for:
```python
if s is None:
if s is not None:
```
and change them to the more pythonic:
```python
if not s:
if s:
```
This kind of search/replace can't be done easily with the result of the git-diff-pr command: the diffs get in the way.

### Limiting annotations of strings will *not* break existing code

The more pythonic tests shown above makes Leo's API *more* general than before because code generalized tests handle *all* values (including arguments) more robustly than in Leo's legacy code.

edreamleo:
See #4780.

### DIFF
--- a/leo/commands/editFileCommands.py
+++ b/leo/commands/editFileCommands.py
@@ -564,11 +564,11 @@ class EditFileCommandsClass(BaseEditCommandsClass):
         """
         GitDiffController(c=self.c).diff_pull_request()
 
-    # @+node:ekr.20260707054655.1: *3* efc.gitDiffPRClone (clone-diff-pr)
+    # @+node:ekr.20260707054655.1: *3* efc.cloneDiffPR (clone-diff-pr)
     @cmd('clone-diff-pr')
-    def gitDiffPRCloneChanged(self, event: LeoKeyEvent = None) -> None:
+    def cloneDiffPR(self, event: LeoKeyEvent = None) -> None:
         """
-        Produce a Leonine diff of pull request in the current branch.
+        Show the added, deleted and changed nodes (without diffs) of the given PR.
         """
         GitDiffController(c=self.c).clone_diff_pr()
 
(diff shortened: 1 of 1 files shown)


edreamleo:
This PR is straightforward. It copies methods rather than using kwargs.

edreamleo:
@boltex You might consider adding this to leoJS, leoserver.py, etc.

leo/commands/editFileCommands.py
```diff
@@ -1301,6 +1375,60 @@ def remove_cruft(self, diff_list: list[str]) -> str:
         return results3
 
     # @+node:ekr.20180510095801.1: *3* gdc.Utils
+    # @+node:ekr.20260707062134.1: *4* gdc.create_clone_diff_node
+    def create_clone_diff_node(
+        self,
+        branch: str,
+        c1: Cmdr,
+        c2: Cmdr,
+        d: dict[str, tuple[VNode, VNode]],
```
boltex, line 1384: @edreamleo 
Here in this **create_clone_diff_node** method, the dict '**d**' can sometimes be  dict[str, tuple[VNode, VNode]]  as shown here but depending on the '**kind**' parameter it can be expected to just be a dict of [str Vnode]

This should be explained in comments in the code.

boltex:
This was ported to LeoJS. 😄

leo/commands/editFileCommands.py
edreamleo: Fixed in rev cdb662073.

edreamleo:
@boltex Rev cdb662073 adds comments about the values in the `d` dict.

PR #4782, opened 2026-07-08
PR: Use full optional syntax for vars and kwargs

edreamleo:
See #4766.

This PR changes *only* annotation syntax. It does *not* change:

- Default *values* for any var, arg, or kwarg!
- .mypy.ini. In particular, mypy does not require strict_optional for any more modules!

A later PR will:

- Change default values for (most, all?) strings from None to the empty string.
- Use pythonic tests instead of tests against None.
- Require the event arg for all commands and in various other places.
  Annotate event vars and args as `event: LeoKeyEvent`, without a default.

### Use correct syntax for all optional values

Replace `x: y = None` with `x: y | None = None` in:

- [x] leo/core.
- [x] leo/plugins.
- [x] all Qt gui modules.

### Other changes

- [x] Use casts in leoVim.py.

### DIFF
--- a/leo/commands/abbrevCommands.py
+++ b/leo/commands/abbrevCommands.py
@@ -9,15 +9,15 @@ from collections.abc import Callable
 import functools
 import re
 import string
-from typing import TYPE_CHECKING
+from typing import cast, TYPE_CHECKING
 from leo.core import leoGlobals as g
 from leo.core import leoNodes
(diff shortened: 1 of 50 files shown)


edreamleo:
@tbpassin In retrospect, it would have been better to make this PR the first PR in this grand project.

Oh well, better late than never.

edreamleo:

This PR changes no executable code, so I'll merge it immediately.

PR #4783, opened 2026-07-08
PR: Improve annotations of event: LeoKeyEvent

edreamleo:
See #4766.

### Background

The `@cmd`, `@g.command` and similar decorators always supply an event arg,
so one would naively think that the correct annotations for commands would always be:
```python
event: LeoKeyEvent
```
However, scripts and unit tests may call the corresponding methods directly *without* an event arg,
so in almost all cases the correct annotations is:
```python
event: LeoKeyEvent | None = None
```
The only exceptions are "helper" functions that are always called with an event arg and should never be called from scripts or plugins.

### Tasks

- [x] Annotate all event args correctly, *defaulting to None*.
   Removing this default was a huge blunder.
   Restoring these default fixes #4784.
- [x] Review the diffs using Leo git-diff-pr command.

### Extras

- [x] Rev 6fa00db81: require explicit kwargs (using the `*,` syntax) in the following commands:
    file-save-to, refresh-from-disk, and unformat-paragraph.
    Taking a SyntaxError is far better than executing erroneous code.
- [x] Rev 9f3b5adb7: g.is_binary_string always returns False.
    A temporary workaround for #4778.

PR #4790, opened 2026-07-09
PR Add more defaults for LeoKeyEvent

edreamleo:
See #4766. This PR continues PR #4783. 

The following rules minimize semantic changes, ensuring that this PR changes as little as possible.

### Annotate commands

- [x] Use `event: LeoKeyEvent | None = None`  in all methods decorated with one of Leo's command decorators.

### Annotate helper methods

- [x] Use `event: LeoKeyEvent | None = None` where it would be syntactically correct to do so.
    - [x] Add guards as necessary to suppress mypy complaints.
- [x] Use `event: LeoKeyEvent | None` where a default would be syntactically incorrect.
    - Such annotations will never generate new mypy complaints about event being None!

### DIFF
--- a/leo/commands/abbrevCommands.py
+++ b/leo/commands/abbrevCommands.py
@@ -71,7 +71,7 @@ class AbbrevCommandsClass(BaseEditCommandsClass):
         self.w = cast(QTextMixin, None)
 
     # @+node:ekr.20150514043850.11: *3* abbrev.expandAbbrev & helpers (entry point)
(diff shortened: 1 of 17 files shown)


edreamleo:

The first comment explains why this PR makes the absolute minimum of semantic changes.

This is not to say that the legacy didn't have bugs, but imo there is little change this PR *adds* bugs.

boltex:
all good here :)

PR #4798, opened 2026-07-12
Fixed local config using multiple @menuat directives.

### LINKED ISSUE #4797: local@menuat only applies last one, also, @menuat cut/copy/clipboard operations only work in global (leoSettings/myLeosettings) configs
The @menuat cut/copy/clipboard operations only work in global (leoSettings/myLeosettings) configs. 

Meaning that they do not work in local file @settings node configurations such as this (which works fine in leoSettings/myLeoSettings as already stated). 

Image

I found the reason: its because when a subsequent @menuat node is found, it restarts with a copy of the menu in global config settings, instead of the local commander's menu config setting if available:

When using @menuat with clipboard operations, this implies that their will be more than one @menuat config setting... see screenshot above.

So with more than one @menuat config setting... local config should have itself apply the @menuat directives on its local config....  but the code below does 'restart' with a copy of the global g.app.config.menuList each pass!

```python
            is_local = 'leosettings' not in self.c.mFileName.lower()
            if is_local:
                mlist = g.app.config.menusList[:]  # <-- This restarts the menu customization from the global config instead of the local config if already started customizing menus !
                self.c.config.set(None, 'menus', 'menus', mlist)
            else:
                mlist = g.app.config.menusList
            # ans = self.patchMenuTree(g.app.config.menusList, targetPath)
            ans = self.patchMenuTree(mlist, targetPath)
```

Here's the solution which I've tried and seems to work: 

```python
            if is_local:
                mlist = self.c.config.getMenusList() # this will return the local menus if they exist, otherwise the global menus
                # if no local menus, start local menu with copy of the global menu.
                if mlist is g.app.config.menusList:  
                    mlist = g.app.config.menusList[:]
                self.c.config.set(None, 'menus', 'menus', mlist)
            else:
                mlist = g.app.config.menusList
```

I'll make a P.R.

boltex:
Fixes #4797

### DIFF
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -3230,7 +3230,11 @@ class LoadManager:
         # Official very late initialization.
         c.frame.tree.initAfterLoad()
         c.initAfterLoad()
-        lm.createMenu(c, c.fileName())
+        lm.createMenu(c, c.fileName())  # This will fetch c.config.getMenusList()
+
+        # Remove this outline's "doMenuat" settings so later outlines won't use them.
+        lm.globalSettingsDict['menus'] = None
+        c.config.settingsDict['menus'] = None
 
         # Common finishing code.
         g.app.unlockLog()
--- a/leo/core/leoConfig.py
+++ b/leo/core/leoConfig.py
@@ -424,7 +424,7 @@ class ParserBaseClass:
                 targetPath = '/' + targetPath
             is_local = 'leosettings' not in self.c.mFileName.lower()
             if is_local:
-                mlist = self.c.config.getMenusList() # returns local menu if it exist, otherwise global.
+                mlist = self.c.config.getMenusList(True) # returns local menu if it exist, otherwise global.
                 # if is global, (first @menuat found), start fresh with a copy of the global menu list.
                 if mlist is g.app.config.menusList:  
                     mlist = g.app.config.menusList[:]
@@ -1948,11 +1948,14 @@ class LocalConfigManager:
         return language
 
     # @+node:ekr.20120215072959.12534: *5* c.config.getMenusList
-    def getMenusList(self) -> list:
+    def getMenusList(self, skip_menulist_pass: bool = False) -> list:
         """Return the list of entries for the @menus tree."""
 
         # Typically empty, unless there is an @menuat setting.
         aList = self.get('menus', 'menus')
+        if skip_menulist_pass:
+            print(f"getMenusList: skip_menulist_pass")
+            return aList or g.app.config.menusList
 
         # Leo calls this method twice when loading an outline.
         if not hasattr(self.c, 'menulist_pass'):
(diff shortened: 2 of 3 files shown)


boltex:
I changed to draft to investigate that '**menulist_pass**' thing in the getMenusList method of localConfigManager. 

Seems a  bit weird. 

perhaps should be guarded by a flag while config is being read/applied ? or a dedicated optional parameter flag?

boltex:
@tbpassin  Edit: I overcame this elsewhere. - see comments below this one. (latest commits fix this)

Using print statements inserted in the codebase, i see that **menulist_pass** is only used to delete the menus from the local commander's config and such **after the plugins_menu.py plugin adds its menu with the method: add_menu_from_settings** 

In my opinion, (maybe wrong) If this was required, it should have been done there (in plugins_menu.py ) instead of as a side-effect of **getMenusList** in localConfigManager! 

but my experiments on windows seem to show this is not required.  (the actual menus are kept in frame.menu and those intances of menu in the config are not refered too anyways... 

Note: At first i tried with an added flag to keep the current behavior because my fix calls getMenusList, which would trigger that menulist_pass hack too early since it was relying (dangerously) on getMenusList getting called only twice as per comments.

In fact , even after opening leo with multiple documents, I see  the menus are correct (with only the added menus entries in the right commanders top menu) even if i open or create new documents before and after those with custom menus. Also ok if i open many at once from closing with many files opened.

~~Then I tested with that block commented out entirely and it still worked fine:~~
Edit: no,after experimentation,  your removal of the menuat settings was indeed required.

```python
    def getMenusList(self, skip_menulist_pass: bool = False) -> list:
        """Return the list of entries for the @menus tree."""

        # Typically empty, unless there is an @menuat setting.
        aList = self.get('menus', 'menus')
        if skip_menulist_pass:
            print(f"getMenusList: skip_menulist_pass")
            return aList or g.app.config.menusList

        # # Leo calls this method twice when loading an outline.
        # if not hasattr(self.c, 'menulist_pass'):
        #     self.c.menulist_pass = 0
        # self.c.menulist_pass += 1

        # # Remove this outline's "doMenuat" settings so later outlines won't use them.
        # if self.c.menulist_pass == 2:
        #     print("deleting this outline's doMenuat settings")
        #     lm = g.app.loadManager
        #     lm.globalSettingsDict['menus'] = None
        #     self.set(None, 'menus', 'menus', None)  # type:ignore
        #     self.c.menulist_pass = 0
        # else:
        #     print(f"getMenusList: menulist_pass: {self.c.menulist_pass}")

        return aList or g.app.config.menusList
```

~~Should we just remove that ?~~

~~Do you recall what side effects that was suposed to prevent? I.E. What was meant by "so later outlines won't use them." ?~~

boltex:
@edreamleo @tbpassin I've pushed latest changes to this PR with that added flag and all the print statements for you to easily experiment and see the trigger of menulist_pass is indeed after the plugin's menu is added to the menus.... (you can then try opening multiple leo files with custom (different) @menuat nodes in each to see that nothing spills over - and even comment out the whole menulist_pass block like I did above in my last comment to also assert that everything is still showing perfect menus after restarting leo and opening those same files. - and /or creating new or opening other files )

I'll cleanup those prints after we come to some conclusion about that menulist_pass counter 'hack' and its suposed intended effect.

boltex:
@tbpassin I removed the menulist_pass logic and  instead delete the config's accumulated menuat info after createing the frame's menu, as I did in Leo-Web. Seems to work fine!  😄

tbpassin:
I don't grasp the menu machinery well enough to make intelligent comments. I did get this branch and tried it on Windows, and my custom menus did get built and function correctly. This includes several menus that are defined only in a specific outline.

I will look through the code changes and see of anything jumps out at me.

tbpassin (approved):
Looks OK as far as I can tell.

edreamleo (approved):

edreamleo:
@boltex @tbpassin Thanks for this work. I don't know anything about the code :-)

Issue #4800, opened 2026-07-16
Document recent breaking changes to Leo's API


edreamleo:
Recent PRs have changed the values *returned* by several methods. 

See PR #4801 for a list of breaking changes to Leo's API. This PR uses truthy tests where required to accommodate the changed API.

edreamleo:
I am closing this issue. Recent PRs have updated LeoDocs.leo

PR #4801, opened 2026-07-16
PR: check all calls to routines whose API has changed

edreamleo:
See #4800.

- [x] Use truthy tests where required by the breaking changes listed below. Where valid, such tests generalize Leo's API.
    A *lot* of diffs and cffs were involved in discovering the places where truthy tests were needed.
    This PR is a big step forward, but only time will tell whether more work is needed.
- [x] LeoDocs.leo: document all breaking changes to Leo's API.
- [x] Use walrus operator in a few changed methods.
    A separate PR will use the walrus operator in many other places.
- [x] TestQtGui.test_bug_4626: always skip this tests: it can still hang.

### Breaking changes to Leo's API

Leo 6.8.10 introduces many breaking changes to Leo's API. Few scripts and plugins will likely be affected by these changes: most of the changed routines are for Leo's internal use.

<details></summary>
<br>

Leo 6.8.10 changes the values *returned* by several methods. As described below, in some cases these changes might break existing scripts or plugins.

### Changed string functions

The functions and methods listed below now return an empty string instead of None. Tests against None will no longer work as expected--the "truthy" tests below must be used instead::

    if s:
    if not s:

- AtFile.readFileToUnicode.
- g.findAtTabWidthDirectives.
- g.findFirstAtLanguageDirective.
- g.findLanguageDirectives.
- g.get_backup_directory.
- g.gitHeadPath.
- g.getUrlFromNode.
- g.guessExternalEditor.
- g.makeAllNonExistentDirectories.
- g.stripBOM.
- g.readFileIntoEncodedString.
- RecentFilesManager.sanitize.
- RecursiveImportController.resolve_dir_arg.

The following methods now return an empty string instead of None:

- SettingsDict.get_setting and SettingsDict.get_string_setting.

### Changed functions that return an optional commander

The functions below may return None instead of a commander:

- g.openUNLFile and g.openUrlHelper.
- g.handleUnl.

*Note* g.openWithFileName now *always* returns a valid commander. This is *not* a breaking change.

### Changed functions that return an optional position.

The functions below may return None instead of a position:

- g.findUnl, g.findGnx, and g.findAnyUnl.

### Deleted global functions and methods

- AtFile.chmod.

mypy revealed that the following functions  were impossible to use:

- g.execute_shell_commands_with_options.
- g.computeBaseDir.
- g.computeCommands.

### Other breaking changes

- LeoImportCommands.removeSentinelsCommand: Remove unused `toString` kwarg.

</details>

### DIFF
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -2952,7 +2952,7 @@ class Commands:
         else:
             format = c.config.getString("headline-time-format-string")
             gmt = c.config.getBool("headline-gmt-time")
-        if format is None:
+        if not format:
             format = default_format
         try:
             # import time
@@ -3226,7 +3226,7 @@ class Commands:
         for d in aList:
             # Look for @path directives.
             path = d.get('path')
-            if path is not None:  # retain empty paths for warnings.
+            if path:  # retain empty paths for warnings.
                 # Convert "path" or <path> to path.
                 path = g.stripPathCruft(path)
                 if path:
@@ -4059,7 +4059,7 @@ class Commands:
         t1 = time.process_time()
 
         # Using the Qt gui will likely overwhelm Leo!
-        if gui is None:
+        if not gui:
             gui = g.app.nullGui
 
         # The main loop.
@@ -4606,7 +4606,7 @@ class Commands:
         deltaTime = c.config.getFloat('outline-nav-extend-delay')
         if deltaTime in (None, 0.0):
             return False
-        if c.navTime is None:
+        if not c.navTime:
             return False  # mypy.
         return time.time() - c.navTime < deltaTime
 
--- a/leo/core/leoCommands.py
+++ b/leo/core/leoCommands.py
@@ -3225,11 +3225,9 @@ class Commands:
         paths = []
         for d in aList:
             # Look for @path directives.
-            path = d.get('path')
-            if path:  # retain empty paths for warnings.
+            if path := d.get('path'):
                 # Convert "path" or <path> to path.
-                path = g.stripPathCruft(path)
-                if path:
+                if path := g.stripPathCruft(path):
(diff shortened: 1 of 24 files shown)


leo/core/leoCommands.py
```diff
@@ -3225,11 +3225,9 @@ def scanAtPathDirectives(self, aList: list) -> str:
         paths = []
         for d in aList:
             # Look for @path directives.
-            path = d.get('path')
-            if path:  # retain empty paths for warnings.
+            if path := d.get('path'):
```
tbpassin, line 3228: `if path := d.get('path', ''):`

Always return string in case `g.stripPathCruft(path)` fails for `None`.

edreamleo: @tbpassin The default isn't necessary. path will be appended only if it is a non-empty string.

tbpassin: That's true but the signature of `g.stripPathCruft()` is

`def stripPathCruft(path: str) -> str`

Without the change I'm suggesting, the method could receive None, which would be a type violation for the `path` argument.

If you are not going to follow these type signatures, then why are you bothering with the immense effort of changing them?

edreamleo: @tbpassin Here is the code snippet in question:
```python
if path := d.get('path'):
    if path := g.stripPathCruft(path):
    ...
```
The first "if" statement will fail if d.get('path') returns None (or an empty string).
Therefore, g.stripPathCruft will never be called with 'None'.

I'm improving annotations in order to make this kind of analysis easier. The better the annotations, the more *likely* that mypy will complain about actual type violations. However, there are no guarantees.

edreamleo:

I will merge this PR soon, after updating LeoDocs.leo.

PR #4804, opened 2026-07-18
Add walrus operators to Leo's core files

edreamleo:
I used the following regex to discover possible changes. There were many false positives:
```console
find: ^(\s*)([\w]+) \= (.*?)\n\s*if\s*([\w.]+)\:\n
repl: \1if \2 := \3:\n
```

- [x] Add walrus op to all files in leo/core, leo/commands, all of Leo's Qt gui files, and Leo's most import plugins.

### Discussion

The more I use the walrus operator, the more I like it:
- All these tests are 50% smaller than before. For me, this is a big deal.
- Small loops are often 25% smaller.

After using the walrus operator, mypy, flake8, or pylint often suggested new changes. Each change is a slight improvement!  See the diffs for details.

In some cases, a var serves no purpose other than to be immediately tested. When using the walrus operator, flake8 complained that the var is unused. The solution was to omit the assignment operator entirely. However, in one place I've added a comment, which was the var's real purpose.

### Quirps

- c.doBatchOperations: The code appears to sort the wrong dict. I've left it alone.

### Substantive code changes

These are candidates for changes to LeoJS.

-  x.updatePublicAndPrivateFiles is now just:
```python
def updatePublicAndPrivateFiles(self, root: Position, fn: str, shadow_fn: str) -> None:
    """handle crucial @shadow read logic.

    This will be called only if the public and private files both exist."""
    x = self
    if x.isSignificantPublicFile(fn):
        # Update the private shadow file from the public file.
        x.propagate_changes(fn, shadow_fn)
```
All the reporting logic now resides in x.propagate_changes. This fixes a quirp: the message is not always given if x.propagate_changes returns True!


edreamleo:

I have carefully reviewed the changes using Leo's git-diff-pr command.

leo/commands/checkerCommands.py
```diff
@@ -668,8 +664,7 @@ def predicate(p: Position) -> bool:
 
         data: list[tuple[str, Position]] = []
         is_at_file = False
-        roots = g.findRootsWithPredicate(c, root, predicate=predicate)
-        if roots:
+        if roots := g.findRootsWithPredicate(c, root, predicate=predicate):
             roots = g.findRootsWithPredicate(c, root, predicate=predicate)
```
tbpassin: Delete this line. The walrus operator has already assigned it.

leo/commands/editCommands.py
```diff
@@ -269,35 +269,31 @@ def merge_node_with_prev_node(event: LeoKeyEvent | None = None) -> None:
 @g.command('next-or-end-of-line')
 def nextOrEndOfLine(event: LeoKeyEvent | None = None) -> None:
     # by Brian Theado.
-    c = event.get('c') if event else None
-    if c:
+    if c := event.get('c') if event else None:
         lineScrollHelper(c, 'next-', 'end-', '')
```
tbpassin, line 273: `c` can be None. can `lineScrollHelper()` handle that case?

leo/commands/editCommands.py
```diff
@@ -711,8 +701,7 @@ def capitalizeHelper(self, event: LeoKeyEvent | None, which: str, undoType: str)
             word2 = word.upper()
         else:
             g.trace(f"can not happen: which = {which}")
-        changed = word != word2
-        if changed:
+        if changed := word != word2:
```
tbpassin, line 704: use `if word != word2:`. the variable `changed` is not used here.

leo/commands/editCommands.py
```diff
@@ -2278,8 +2261,8 @@ def doSmartQuote(self, action: str, ch: str, oldSel: tuple[int, int], w: QTextMi
         # Pick the correct curly quote.
         s = w.getAllText() or ""
         i2 = g.skip_to_start_of_line(s, max(0, ins - 1))
-        open_curly = ins == i2 or ins > i2 and s[ins - 1] in ' \t'  # not s[ins-1].isalnum()
-        if open_curly:
+        if ins == i2 or ins > i2 and s[ins - 1] in ' \t':  # not s[ins-1].isalnum()
```
tbpassin, line 2264: Please revert this back to the previous `open_curly = ` form. Without this assignment the logic test is opaque. Adding the comment `# An open curly bracket` does not help the reader know what the test is supposed to do at the point of test. Also, you didn't even save a line because the added comment line made up for the line you saved by making the test unintelligible.

leo/commands/editFileCommands.py
```diff
@@ -146,18 +145,14 @@ def clone_and_move(parent: Position, section_p: Position) -> None:
                 clone.doDelete()
                 self.errors += 1
 
-        #
         # First, look in p's subtree.
-        section_p = self.find_section(p, section_name)
-        if section_p:
-            # g.trace('FOUND', section_name)
+        if section_p := self.find_section(p, section_name):
             # Already defined in a good place.
```
tbpassin: This comment is not understandable: what does a "good place" mean?  Please clarify the comment.

leo/commands/gotoCommands.py
```diff
@@ -374,8 +372,7 @@ def find_gnx2(self, gnx: str) -> Position | None:
 
         # Search the entire outline.
         positions: list[Position]
-        backwards = c.config.getBool('search-links-backwards', default=True)
-        if backwards:
+        if c.config.getBool('search-links-backwards', default=True):
```
tbpassin, line 375: Change this line and the next 3 lines to read:

```
positions = c.all_positions()
if c.config.getBool('search-links-backwards', default=True):
    positions = reversed(list(positions))
```

The consumer of `positions` can consume a generator as well as a list. The suggested version is shorter and more clear.

leo/commands/checkerCommands.py
edreamleo: @tbpassin Good catch.

leo/commands/editCommands.py
edreamleo: @tbpassin Maybe, but I'd like to preserve the consistent style.

leo/commands/editCommands.py
edreamleo: @tbpassin changed is used a few lines later. It's a bit hard to see.

Anyway, flake8 will complain if a var is unused.

leo/commands/editFileCommands.py
edreamleo: @tbpassin I think I'd rather just delete the comment :-)

leo/commands/gotoCommands.py
edreamleo: @tbpassin Apparently not.  This Leonine test:
```python
g.printObj(list(reversed(c.all_positions())))
```
generates the error:
```console
TypeError: 'generator' object is not reversible
```

leo/commands/editCommands.py
edreamleo: @tbpassin That's why I added a new comment at line 2265.

flake8 will complain if I use a walrus operator to assign to the open_curly var.

leo/commands/checkerCommands.py
edreamleo: @tbpassin Fixed at rev 1b547ec38.

leo/commands/editFileCommands.py
edreamleo: @tbpassin Rev 21f24ac19 deletes this unclear and unnecessary comment.

edreamleo:
@tbpassin Many thanks for your thorough review. I greatly appreciate it.

edreamleo:
@tbpassin Oops. Somehow one final check (my fft.cmd test) failed. pyflakes-leo reports:
```console
pyflakes_leo.py

:637:20: local variable 'provided' is assigned to but never used
1 errors in c:\Repos\leo-editor\leo\plugins\nested_splitter.py

:542:20: local variable 'out' is assigned to but never used
:3486:12: local variable 'state' is assigned to but never used
:3492:12: local variable 'state' is assigned to but never used
3 errors in c:\Repos\leo-editor\leo\plugins\qt_frame.py

:856:16: local variable 'is_dark_on_light' is assigned to but never used
:1065:24: local variable 'is_dark_on_light' is assigned to but never used
2 errors in c:\Repos\leo-editor\leo\plugins\qt_text.py
```
I'll fix these immediately.  Also, my normal (ft.cmd) check should always run pyflakes-leo!
*Note*: this is a private script: there is no corresponding file in the leo/scripts directory.

edreamleo:
@tbpassin Rev 4c3f7a624 removes the last-minute complaints.

*Now* the PR should be complete.

tbpassin:
> @tbpassin Rev [4c3f7a6](https://github.com/leo-editor/leo-editor/commit/4c3f7a624d87b044926c232f9e4219996302934a) removes the last-minute complaints.
> 
> _Now_ the PR should be complete.

I'm not even half-way through yet, so don't get hasty...

leo/core/leoApp.py
```diff
@@ -1778,13 +1774,7 @@ def computeLeoDir(self) -> str:
...
-            # __file__ is randomly upper or lower case!
-            # The made for an ugly recent files list.
-            path = g.__file__  # was leo.__file__
-            if path:
-                # Possible fix for bug 735938:
-                # Do the following only if path exists.
+            if path := g.__file__:
                 # @+<< resolve symlinks >>
                 # @+node:ekr.20120209051836.10257: *6* << resolve symlinks >>
                 if path.endswith('pyc'):
```
tbpassin, line 1780: Maybe check for `.pyo` too?

leo/commands/gotoCommands.py
tbpassin: > @tbpassin Apparently not. This Leonine test:
> 
> ```python
> g.printObj(list(reversed(c.all_positions())))
> ```
> 
> generates the error:
> 
> ```
> TypeError: 'generator' object is not reversible
> ```

I think I reversed `list` and `reversed`. `c.all_positions()` returns a generator. `reversed(list(` also returns a generator but any sequence-like type will work for the consumer.

`g.printObj(reversed(list(c.all_positions())))`

leo/commands/editCommands.py
tbpassin: > @tbpassin That's why I added a new comment at line 2265.
> 
> flake8 will complain if I use a walrus operator to assign to the open_curly var.

My point was that you can skip the walrus operator altogether, and have a more clear bit of code of the same length.

leo/core/leoAtFile.py
```diff
@@ -3137,8 +3115,7 @@ def replaceFile(
                 g.warning("correcting line endings in:", fileName)
 
         # Write a changed file.
-        ok = g.writeFile(contents, encoding, fileName)
-        if ok:
+        if ok := g.writeFile(contents, encoding, fileName):
```
tbpassin, line 3118: `ok` isn't used so far as I can see. Change line to 
`if g.writeFile(contents, encoding, fileName):`

leo/core/leoBeautify.py
```diff
@@ -107,8 +106,7 @@ def fstringify_files(event: LeoKeyEvent | None = None) -> None:
             print('')
             print(g.shortFileName(filename))
             changed = leoAst.Fstringify().fstringify_file(filename)
-            changed_s = 'changed' if changed else 'unchanged'
-            if changed:
+            if changed_s := 'changed' if changed else 'unchanged':
```
tbpassin, line 109: This line is very hard to scan and understand. Please clarify even if another line would be needed. e.g.,

```
chg_str, changed = 'changed', leoAst.Fstringify().fstringify_file(filename)
changed_s = chg_str if changed else 'unchanged'
```

There you go, same number of lines and much easier to understand.

leo/core/leoColorizer.py
```diff
@@ -1498,8 +1490,7 @@ def setTag(self, tag: str, s: str, i: int, j: int) -> None:
             return
         tag = tag.lower().strip()
         # A hack to allow continuation dots on any tag.
-        dots = tag.startswith('dots')
-        if dots:
+        if dots := tag.startswith('dots'):
             tag = tag[len('dots') :]
```
tbpassin, line 1494: The variable `dots` isn't used. Better to define a global constant

`DOTS  = 'dots'`

Then, with no walrus operator:

```
if tag.startswith(DOTS):
    tag = tag[len(DOTS):]
```

leo/core/leoCommands.py
```diff
@@ -3466,8 +3453,7 @@ def find_line(path: str, n: int) -> tuple[Position, int]:
             useSentinels=True,
         )
         # Create a temp file if root is not an @<file> node.
-        use_temp = not root.isAnyAtFileNode()
-        if use_temp:
+        if use_temp := not root.isAnyAtFileNode():
```
tbpassin, line 3456: The variable `use_temp` does not seem to be used. The test could simply be

`if not root.isAnyAtFileNode():`

leo/core/leoApp.py
edreamleo: @tbpassin Hmm. Let's leave this for another day.

leo/core/leoBeautify.py
edreamleo: @tbpassin Huh? This line is straightforward. Why? Because "changed_s" is a different var from "changed".
```python
 if changed_s := 'changed' if changed else 'unchanged':
```

leo/core/leoAtFile.py
edreamleo: @tbpassin Thomas, you can assume that flake8 will catch any unused vars.

There is a `return ok` line at the end of the method.

leo/core/leoConfig.py
```diff
@@ -1847,8 +1834,7 @@ def getData(
     ) -> list[str]:
         """Return a list of non-comment strings in the body text of @data setting."""
         # 904: Add local abbreviations to global settings.
-        append = setting == 'global-abbreviations'
-        if append:
+        if append := setting == 'global-abbreviations':
```
tbpassin, line 1837: The variable `append` does not seem to be used. The test could be simply

`if setting == 'global-abbreviations':`

leo/commands/gotoCommands.py
edreamleo: @tbpassin This is small beer. The original code can stay as it is.

Furthermore
```python
g.printObj(reversed(list(c.all_positions())))
```
prints:
```console
<list_reverseiterator object at 0x00000133E98E0610>
```

boltex:
There's no walrus operator in typescript hehe! :) (i'll remove the leojs tag)

edreamleo:
@boltex I've put the LeoJS tag back on because of the substantial change.

boltex:
@edreamleo Oh ok, thanks! I'll look for changes other than the walrus operator. no problem 😄

leo/core/leoBeautify.py
tbpassin: > @tbpassin Huh? This line is straightforward. Why? Because "changed_s" is a different var from "changed".
> 
> ```python
>  if changed_s := 'changed' if changed else 'unchanged':
> ```

You are too close to it. The line is confusing. Yes, I know `changed_s` is a different variable. That one line contains four similar variations on the word "changed", one assignment, and two logic tests. There is too much going on.

leo/plugins/qt_text.py
```diff
@@ -853,8 +853,7 @@ def calc_hl(palette: QtGui.QPalette) -> QColor:
             hsv_bg = bg.getHsv()
             v_fg = hsv_fg[2]
             v_bg = hsv_bg[2]
-            is_dark_on_light = v_fg < v_bg
```
tbpassin, line 856: Please restore the variable name `is_dark_on_light`. It has semantic value because it tells the reader the purpose of the test.

leo/plugins/qt_text.py
```diff
@@ -1063,8 +1062,7 @@ def set_cursor_width(width: int) -> None:
                     bg = palette.window().color()
                     hsv_fg = fg.getHsv()[2]
                     hsv_bg = bg.getHsv()[2]
-                    is_dark_on_light = hsv_fg < hsv_bg
```
tbpassin, line 1066: Same comment as line 856:

Please restore the variable name is_dark_on_light. It has semantic value because it tells the reader the purpose of the test.

tbpassin (commented):
i have finished my review and have nothing to add to the inline comments.

leo/core/leoColorizer.py
edreamleo: @tbpassin tag is an arg, so there's no way to define a global constant.

leo/core/leoConfig.py
edreamleo: Thomas, 'append' *is* used. To repeat, flake8 would tell me if weren't.

leo/plugins/qt_text.py
edreamleo: @tbpassin I disagree. The var is *less* clear than the code that immediately follows:
```python
if hsv_fg < hsv_bg:
    fg = fg.lighter()
else:
    fg = fg.darker()
```
Here, fg is a [QColor](https://doc.qt.io/qt-6/qcolor.html), and lighter and darker are QColor methods.
It is my strong opinion that the var adds nothing.

leo/plugins/qt_text.py
edreamleo: I disagree. See the next comment.

leo/core/leoCommands.py
edreamleo: To repeat, flake8 would complaint if use_temp is not used. In this case, it is used in the 'finally' clause, as a quick search would have verified.

edreamleo:
@tbpassin Many thanks for all your comments.

leo/core/leoBeautify.py
edreamleo: @tbpassin I often use the following pattern when defining complex f-strings:
```python
x_s = 'a-string' if x else 'another-string'
```
I'm comfy with it in this context.

PR #4806, opened 2026-07-19
Add missing support for .leojs file extension in specific places.

boltex:
A small & quick fix.


boltex (commented):
Also fixed a small 'db' vs '.db' string check

edreamleo (approved):
Looks good. Thanks for this fix.

PR #4808, opened 2026-07-20
PR fixes #4807: g.readFileIntoString should return None

### LINKED ISSUE #4807: g.readFileIntoString should return None if the file does not exist
See PR #4808.

@boltex reports:

If the user selects a file to be created as a node (@edit, or whatever commands uses those methods) the empty file will be intended to go along with the creation of the node, in effect with an empty body, and not merely aborted without creating a node. 
 
There are 18 calls to g.readFileIntoString:

- Most of these calls then use a truthy test. Those tests should be converted to tests on None.
- In some cases, assignments should be changed from `x = s` to `x = s or ''`.

edreamleo:
See #4807.

- [x] Change g.readFileIntoString. Its return type is now `tuple[str | None, str]`.
- [x] Convert truthy tests to tests on None.
- [x] Change assignments from `x = s` to `x = s or ''`.

### DIFF
--- a/leo/commands/commanderFileCommands.py
+++ b/leo/commands/commanderFileCommands.py
@@ -289,7 +289,7 @@ def import_txt_file(self: Self, fn: str) -> None:
     p = c.p.insertAfter()
     p.h = f"@edit {c.relativeDirectory(fn)}"
     s, e = g.readFileIntoString(fn, kind='@edit')
-    p.b = s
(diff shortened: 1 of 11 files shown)


edreamleo:
@boltex Ready for review.

boltex (approved):
Approved - looks good :)

PR #4812, opened 2026-07-21
PR: Require strict optional for leo/core files

edreamleo:
See #4812.

flake8, mypy, pyflakes, pylint guided all my changes. I avoided all "creative" changes.
The acid test: will Leo start up!

LeoDocs.leo *already* documents all breaking changes. 

- [x] .mypy.ini: Require strict optional in leo/core, leo/external, and most of Leo's qt-related files.
   For now, work around recent mypy bugs.
- [x] Fix all resulting mypy warnings.

### About event args

This PR greatly improves the annotations for "event" args.  The general rules are as follows.

For all functions/methods decorated with (@g.command, @cmd, etc),
the "event" arg *must* have the annotation `event: LeoKeyEvent | None = None`.
With strict-optional = True, mypy requires *runtime* checks that code works with an empty "event" kwarg.
You could say that these required checks are the main point of using strict-optional = True!

In most other cases, `event: LeoKeyEvent | None = None` is *still* the correct annotation.
However, there are important exceptions.

In particular, the event arg is *required* for k.masterKeyHandler and *all* of its helpers,
so the proper annotations are `event: LeoKeyEvent` for all those methods.
*Note*: k.masterKeyHandler calls k.checkKeyEvent to check the event.
This call is equivalent to the assertion: `assert event is not None`.

### Changes to Qt-related files.

mypy often suggested adding asserts. These asserts are excellent new additions to Leo.
If they fail, something is seriously wrong.

All through Leo, but *particularly* in Qt-related files, the comment `# PR #4812` marks a non-trivial change.

Leo's Qt event handlers now check their arguments much more thoroughly for edge cases.
These checks are significant: failures in such methods can cause a hard crash.

### Typical changes

- In some places, these checkers suggested that I replace a *definition* of an ivar with its *declaration*.
- In other places, using casts were necessary. Such casts often required changes to imports.
- Within function/method declarations, it was often possible to change `s: str | None = None` to `str = ''`.
  Such changes are typically *not* breaking changes to Leo's API.
  However, those changes did require careful examination of the function/method whose signature changed.

### Breaking changes

Changing the *returned* value of a function/method from `str | None` to `str` *is* a breaking change.
This PR handles such changes in Leo *itself*, but user plugins and scripts may break as a result.
Imo, the changes are unlikely to break actual scripts, but there are no guarantees.

The most concerning changes involve the global functions in leoGlobals.py.
LeoDocs.leo *already* documents all such changes.


edreamleo:
@boltex Please review changes to leoserver.py.

@boltex @tbpassin This PR is ready for review. I'm working on documentation.

boltex:
@edreamleo I revised the changes in leoserver.py and I tested it out also to make sure with the latest leoInteg version. 

All seemed to work fine.

edreamleo:
@boltex Thanks for your testing. I'm going to merge this PR now so that we can all test the code more thoroughly.

PR #4814, opened 2026-07-23
Update most themes, especially dark themes, to make radiobuttons work ...

### LINKED ISSUE #4813: FInd Dialog Radio Buttons Can Fail On Windows
Certain dark themes may cause the Find dialog radio buttons to not function on Windows. This is apparently caused by some obscure interaction between Qt and Windows for certain color combinations., for a recent Windows update.

I have been able to fix this issue by adding some CSS selectors for Radio Buttons. I am preparing a PR that updates most of the theme outlines that come standard with Leo.

At the same time I have added more selectors for checkboxes. Now checkbox outlines are visible for most themes where before many were invisible.

tbpassin:
... for new Windows updates. Also make check box indicator boxes more visible.

Fixes https://github.com/leo-editor/leo-editor/issues/4813

This PR changes only theme outlines.


edreamleo (approved):
Thanks, Thomas!

PR #4824, opened 2026-07-29
PR fixes #4757: better annotations for generators

### LINKED ISSUE #4757: Generators should be typed with generics
See PR #4824.

It would be nice to have the typing somewhat tighter with those nifty generators! 😄 

For example, here's the top of the generators copy-pasted directly from leoNodes.py and typed with generics in the generator type:

```python
    # @+node:ekr.20091001141621.6061: *4* c.Generators
    # @+node:ekr.20091001141621.6043: *5* c.all_nodes & all_unique_nodes
    def all_nodes(self) -> Generator[VNode, None, None]:
        """A generator returning all vnodes in the outline, in outline order."""
        c = self
        for p in c.all_positions():
            yield p.v

    def all_unique_nodes(self) -> Generator[VNode, None, None]:
        """A generator returning each vnode of the outline."""
        c = self
        for p in c.all_unique_positions(copy=False):
            yield p.v

    # Compatibility with old code...

    all_vnodes_iter = all_nodes
    all_unique_vnodes_iter = all_unique_nodes

    # @+node:ekr.20091001141621.6044: *5* c.all_positions
    def all_positions(self, copy: bool = True) -> Generator[Position, None, None]:
        """A generator return all positions of the outline, in outline order."""
        c = self
        p = c.rootPosition()
        while p:
            yield p.copy() if copy else p
            p.moveToThreadNext()

    # Compatibility with old code...

    all_positions_iter = all_positions
    allNodes_iter = all_positions
    safe_all_positions = all_positions
```

edreamleo:
See #4757.

- [x] Update all Generator annotations.
- [x] leoPersistence.py: Suppress one new mypy complaint.

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -5,7 +5,7 @@
 [mypy]
 python_version = 3.14
 cache_dir = mypy_cache
-incremental = False
+incremental = True
 
 # Defaults for all files.
 check_untyped_defs = True
--- a/leo/commands/commanderOutlineCommands.py
+++ b/leo/commands/commanderOutlineCommands.py
@@ -235,7 +235,7 @@ def pasteAsTemplate(self: Cmdr, event: LeoKeyEvent | None = None) -> None:
         return g.app.nodeIndices.computeNewIndex()
 
     # @+node:vitalije.20200529115141.1: *4* viter
-    def viter(parent_gnx: str, xv: Any) -> Generator[tuple, None, None]:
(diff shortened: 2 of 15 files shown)


edreamleo:

boltex (approved):
Thanks for typing the generators with specific & precise output types! This makes the typing of variables using their returned values at their point of usage be of the proper type! 😄

edreamleo:
@boltex You're welcome! Thanks for your approval.

PR #4826, opened 2026-07-29
PR: fix ruff lint errors on Linux

edreamleo:
These errors may have caused the ci.yml action to fail.


edreamleo:
Viktor reports all is well in his tests. That's good enough for me.

PR #4827, opened 2026-07-29
PR fixes #4821: remove ruff suppressions for useful tests

### LINKED ISSUE #4821: Remove recently-added ruff lint suppressions
PR #4820 added about 100 new suppressions to ruff.toml.

In general, the new checks are valuable and should not be suppressed!

edreamleo:
See #4821.

- [x] ruff.toml: Remove many suppressions, leaving other suppressions in place.
- [x] Fix all resulting ruff lint complaints.

### Discussion

Many suppressions would be hard to remove. These can wait indefinitely.

I used one or more commits for each removed suppression.


edreamleo:

@boltex Please review changes to leoserver.py.

leo/commands/editCommands.py
```diff
@@ -486,7 +486,6 @@ def dumpCaches(self, event: LeoKeyEvent | None = None) -> None:  # pragma: no co
     @cmd('do-nothing')
     def doNothing(self, event: LeoKeyEvent | None = None) -> None:
         """A placeholder command, useful for testing bindings."""
-        pass
```
tbpassin, line 489: This do-nothing function will fail without some statement here. An example:

```
>>> def nope():
...     
>>> nope()

Uncaught exception in Leo...
  File "<input>", line 1, in <module>
NameError: name 'nope' is not defined. Did you mean: 'None'?

>>> def nope1():
...         pass
...         
>>> nope1()
```

leo/commands/editFileCommands.py
```diff
@@ -652,7 +652,7 @@ def saveFile(self, event: LeoKeyEvent | None = None) -> None:
                 s = w.getAllText()
                 with open(fileName, 'w') as f:
                     f.write(s)
-            except IOError:
+            except OSError:
```
tbpassin, line 655: IOError is now an alias for OSError. So this code was right as it was. However, updating to OSError isn't wrong, it just wasn't needed. Or maybe ruff or someone complained?

leo/core/leoBeautify.py
```diff
@@ -205,7 +205,6 @@ def should_beautify(p: Position) -> bool:
                 return False
             # The ambiguous node has no effect.
```
tbpassin, line 206: Move these two comment lines up ahead of the `return False` statement, otherwise they don't seem to make much sense.

leo/core/leoCache.py
```diff
@@ -373,7 +374,6 @@ def do_block(cur: object) -> Value:
     # @+node:vitalije.20170716201700.23: *3* uncache (SqlitePickleShare)
     def uncache(self, *items: Args) -> None:
         """not used in SqlitePickleShare"""
-        pass
```
tbpassin, line 376: Some statement has to be in the body of a method def.

leo/core/leoColorizer.py
```diff
@@ -131,7 +131,6 @@ def __init__(self, c: Cmdr) -> None:
     # @+node:ekr.20190324045134.1: *4* BaseColorizer.init
     def init(self) -> None:
         """May be over-ridden in subclasses."""
-        pass
```
tbpassin, line 134: A method body has to have some statement.

leo/core/leoCommands.py
```diff
@@ -635,7 +635,6 @@ def idle_focus_helper(self, tag: str, keys: dict) -> None:
     # @+node:ekr.20081005065934.1: *4* c.initAfterLoad
     def initAfterLoad(self) -> None:
         """Provide an official hook for late inits of the commander."""
-        pass
```
tbpassin, line 638: A method body has to have a statement. If not `pass` it could be `...`.

Since there are a number of noops that have been changed in this PR, How about defining a g-level no-op function:

`g.noop = lambda *args, **kwargs: None`

leo/core/leoCommands.py
```diff
@@ -2772,8 +2771,10 @@ def checkOutlineXML(self, dump: bool = True) -> bool:
                 )
                 try:
                     with open(filename, 'bw') as f:
-                        for s in g.splitLines(translated_contents):
-                            f.write(g.toEncodedString(s, reportErrors=True))
+                        f.writelines(  # PR #4827
```
tbpassin: This change replaces two lines with four lines that do the same thing and are harder to read. Let's keep the original.

leo/core/leoCompare.py
```diff
@@ -478,8 +478,6 @@ class LeoCompare(BaseLeoCompare):
     These are not very useful comparisons.
     """
 
-    pass
```
tbpassin, line 481: A class declaration needs at least one statement.

leo/core/leoGlobals.py
```diff
@@ -6461,24 +6450,32 @@ def enl(tabName: str = 'Log') -> None:
 
 
 # @+node:ekr.20100914094836.5892: *3* g.error, g.note, g.warning, g.red, g.blue
-def blue(*args: Args, **kwargs: KWargs) -> None:
-    g.es_print(color='blue', *args, **kwargs)
+# PR #4827
+
+
+def blue(*args: Args, **kwargs: Any) -> None:
```
tbpassin, line 6456: kwargs is called like  dictionary. So `Any` isn't the correct type. Won't `dict` be correct?

leo/core/leoGlobals.py
```diff
@@ -6461,24 +6450,32 @@ def enl(tabName: str = 'Log') -> None:
...
+    kwargs['color'] = 'blue'
+    g.es_print(*args, **kwargs)
 
 
 def error(*args: Args, **kwargs: KWargs) -> None:
```
tbpassin, line 6461: just above, in ` def blue()`, the type in `**kwargs`` was changed to `Any`, which I suggested isn't right. It seems to me that both these functions should use the same signature for `**kwargs`.

There are some similar functions defined below which I won't comment individually on.

leo/core/leoUndo.py
```diff
@@ -1427,7 +1427,6 @@ def redo(self, event: LeoKeyEvent | None = None) -> None:
     # @+node:ekr.20191213085226.1: *4*  u.reloadHelper (do nothing)
```
tbpassin, line 1427: Must have a statement in the function body.

leo/core/signal_manager.py
```diff
@@ -38,8 +38,6 @@ class MsgSignalHandled:
     other listeners from being called
     """
```
tbpassin, line 40: Must have a statement in the class body.

leo/core/signal_manager.py
```diff
@@ -168,8 +166,6 @@ def check_work(self, num: object, animal: str = 'eels') -> None:
     class SomeProxy:
         """Like a public notice board"""
 
-        pass
```
tbpassin, line 171: Must have a statement in a class body.

leo/commands/editCommands.py
edreamleo: Correct, but adding a docstring will fix the problem.

leo/core/leoCommands.py
edreamleo: g.noop isn't necessary.

leo/core/leoCommands.py
edreamleo: I agree. But the "for" loop needs a # noqa suppression. This is what I'll do.

leo/core/leoGlobals.py
edreamleo: Iirc, annotating *args and **kwargs is more difficult than you might think.

edreamleo:

"pass" statements aren't necessary when a class or method contains a docstring.

About the use of f.write in a "for" loop. I agree! Rev a28bfd03b7 fixes this.

leo/core/leoBeautify.py
edreamleo: Good catch. Rev 1063364394 restores the "useless" pass statement, with a suppression.

This "pass" statement is the target of the comment!

leo/commands/editFileCommands.py
edreamleo: Thanks for this comment. It's true that OSError and IOError are synonyms, but as you can see, this PR replaces several seemingly unrelated errors with OSError. So there are two reasons to make the change overall:

- It appears to be the new preferred style. I think we can trust the ruff lint devs about this one.
- OSError appears to be more general. It can be used, as in this PR, for errors that don't pertain to IO.

Some examples:

- In geotag.py and mod_http: replace socket.error by OSError.

edreamleo:
@tbpassin Thanks again for your careful review. I shall merge this PR now, but please feel free to comment on this (or any other) PR at any time.

PR #4830, opened 2026-07-31
Fix ci.yml: suppress EXE001/EXE002 (WSL/Windows-invisible ruff rules)

vivainio:
## Summary

Every push to `ci.yml` on `devel` has failed, most recently: https://github.com/leo-editor/leo-editor/actions/runs/30537114482

The failure is `ruff check leo` reporting "Found 18 errors", all `EXE001`/`EXE002`:

```
EXE001 Shebang is present but file is not executable
 --> leo/core/leoBridge.py:1:1
EXE002 The file is executable but no shebang is present
 --> leo/core/leoCommands.py:1:1
...
```

18 files total: `leo/core/leoBridge.py`, `leo/core/leoCommands.py`, `leo/core/leoVersion.py`,
`leo/core/runLeo.py`, `leo/dist/leo-install.py`, `leo/modes/kivy.py`, `leo/plugins/bigdash.py`,
`leo/plugins/examples/slowOut.py`, `leo/plugins/examples/slowOutNoFlush.py`,
`leo/plugins/leo_pdf.py`, `leo/plugins/nodediff.py`, `leo/plugins/nodetags.py`,
`leo/plugins/nodewatch.py`, `leo/plugins/python_terminal.py`, `leo/plugins/qtGui.py`,
`leo/plugins/qt_quicksearch.py`, `leo/plugins/rss.py`, `leo/plugins/viewrendered.py`

**Root cause**: these two ruff rules compare the git-tracked executable bit against whether
a file has a `#!` shebang line — a permission-vs-content mismatch that's accumulated over the
years across editors/platforms. Per ruff's own docs: *"this rule is only available on
Unix-like systems, and is not enforced on Windows or WSL."* GitHub Actions runs on real
Ubuntu, where the rule fires; a Windows or WSL dev box (`ruff --version` confirmed on WSL
here) silently skips it. That's why this was unreproducible locally and kept getting
reverted (`5a8b03323`, `93bf528ed`), fixed once (#4826, which addressed two *different*,
genuinely cross-platform ruff findings), restored, and immediately failed again on the very
next run (`909ebbd90` → failure 6 minutes later) since #4826 never touched EXE001/EXE002 —
leading to it being dropped for good (`6b2da890f`).

## Changes
- `ruff.toml`: add `EXE001`/`EXE002` to `[lint].ignore`, with a comment explaining why —
  matches the existing pattern of pragmatic suppressions already in this file, and avoids
[...]

PR #4833, opened 2026-07-31
Add non-blocking ty (astral-sh) check job to CI

vivainio:
## Summary

Stacked on #4830 (which restores `ci.yml`) — this PR's diff includes that one's
two commits until #4830 merges; the new commits here are "Add non-blocking ty
check job to CI", "Remove type:ignore comments confirmed dead by both mypy
and ty", and "Merge ty check into the mypy job as a second step".

Adds a second, **non-blocking** (`continue-on-error: true`) check that runs
[`ty`](https://github.com/astral-sh/ty), Astral's Rust type checker, alongside
the existing mypy setup — purely to make its findings visible in PR logs while
we evaluate it, without gating merges on it.

### Why evaluate `ty`
- ~10x faster than mypy in local testing on this repo (0.6-0.7s vs ~7-10s,
  warm cache, same file scope).
- With real deps installed (PyQt6), mypy is currently 0-error clean; `ty`
  still surfaces real findings mypy's config doesn't check for — mainly
  legacy implicit-`Optional` parameters (`.mypy.ini` sets
  `strict_optional = False` by default) and Qt method overrides with
  incompatible signatures (mypy skips override-compatibility checks for
  unannotated parameters; `ty` infers the expected type from the PyQt6 stub
  and checks anyway).
- Still beta (`Development Status :: 4 - Beta`, versioning weekly) and
  noticeably stricter than mypy's current config in places — not something to
  gate merges on yet, hence `continue-on-error`.

## Changes
- `.github/workflows/ci.yml`: `ty` runs as a second step in the `mypy` job
  (not a separate job) — both need essentially the same environment (Python
  3.12 + PyQt6), so this reuses the `mypy` job's checkout/setup/install
  instead of paying for a second one, which was the dominant cost of either
  job. The `ty` step has its own `continue-on-error: true`, which works the
  same at step level as at job level, so it still can't fail the build.
- `ty.toml`: mirrors `.mypy.ini`'s `exclude` list for the same file scope, and
  suppresses `unresolved-attribute` (ty's `attr-defined` equivalent) to match
  the existing suppression for Leo's dynamically-assigned attributes across
  `leo.core`/`leo.commands`/`leo.plugins`/`leo.modes`/`leo.external`. Nothing
  else is suppressed — since the job can't fail the build, there's no need to
  hide findings from it yet.

## Cleanup: dead `type:ignore` comments
Running `ty` surfaced 180 `unused-type-ignore-comment` findings. Most of those
are only "unused" from `ty`'s point of view — `python -m mypy
--warn-unused-ignores leo` (that flag is off in `.mypy.ini`) independently
flags just 96 of the same lines, and only 74 lines appear in *both* lists:
comments neither checker finds a reason for anymore. Those 74 are removed
here. The other ~106 `ty`-only findings are left alone, since mypy's stricter
inference in those spots still relies on them — deleting them would silently
un-suppress a real mypy error the next time mypy tightens up.

## Verification
- `ty check --python <venv-with-PyQt6> leo` runs successfully (exit 1, as
  expected — findings exist but the CI job won't fail on it).
- `ci.yml` and `ty.toml` both parse cleanly (YAML/TOML).
- After removing the 74 confirmed-dead comments: `python -m mypy leo` still
  reports "Success: no issues found", and `ty check leo`'s diagnostic count
  drops from 733 to 659.

## Test plan
- [ ] Confirm the `ty` step runs and reports (but doesn't block) in the `mypy`
      job on this PR
- [ ] Confirm the `lint` (ruff) job still passes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,3 +21,20 @@ jobs:
           cache: pip
       - run: pip install ruff
       - run: ruff check leo
+
+  # Informational only: ty (https://github.com/astral-sh/ty) is an
+  # actively-developed, much faster alternative to mypy, still in beta.
+  # It's noticeably stricter than our current mypy config in some areas
+  # (see ty.toml), so this job is not allowed to fail the build -- it's
+  # here so its findings show up in PR logs while we evaluate it.
+  ty:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+          cache: pip
+      - run: pip install ty PyQt6 PyQt6-QScintilla
+      - run: ty check --python "$(python -c 'import sys; print(sys.prefix)')" leo
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,22 +33,14 @@ jobs:
       - run: pip install "mypy<2" mypy-extensions typing_extensions
           types-docutils types-Markdown types-paramiko types-PyYAML
           types-requests types-six
-          "PyQt6>=6.6" PyQt6-QScintilla PyQt6-WebEngine
+          "PyQt6>=6.6" PyQt6-QScintilla PyQt6-WebEngine ty
       - run: python -m mypy leo
-
-  # Informational only: ty (https://github.com/astral-sh/ty) is an
-  # actively-developed, much faster alternative to mypy, still in beta.
-  # It's noticeably stricter than our current mypy config in some areas
-  # (see ty.toml), so this job is not allowed to fail the build -- it's
-  # here so its findings show up in PR logs while we evaluate it.
-  ty:
-    runs-on: ubuntu-latest
-    continue-on-error: true
-    steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
-        with:
-          python-version: "3.12"
-          cache: pip
-      - run: pip install ty PyQt6 PyQt6-QScintilla
+      # Informational only: ty (https://github.com/astral-sh/ty) is an
+      # actively-developed, much faster alternative to mypy, still in beta.
+      # It's noticeably stricter than our current mypy config in some areas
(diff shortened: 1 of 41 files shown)


edreamleo:
@vivainio Thanks for this. Please resolve the merge conflict.

vivainio:
> @vivainio Thanks for this. Please resolve the merge conflict.

Done. I'm also happy to add more ty fixes in followup prs

edreamleo:
Here is my testing log in this branch:
```console
ft.cmd
beautify_all_leo.py
561 files left unchanged
run_test_leo.py
...................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................s.................................................................................s..................................................................................................................................
----------------------------------------------------------------------
Ran 872 tests in 8.417s

OK (skipped=2)
pyflakes_leo.py
flake8_leo.py
mypy_leo.py
leo\core\leoBeautify.py:17: error: Incompatible types in assignment (expression has type "None", variable has type Module)  [assignment]
leo\commands\checkerCommands.py:37: error: Incompatible types in assignment (expression has type "None", variable has type Module)  [assignment]
leo\plugins\viewrendered.py:255: error: Cannot assign to a type  [misc]
Found 3 errors in 3 files (checked 258 source files)
ruff_leo.py
All checks passed!
check_leo.py: files: 94 contexts: 6712 attrs: 47062 in 1.95 sec.
Done!
```
Please resolve the mypy complaints.

vivainio:
@edreamleo Fixed all 3 mypy complaints in a4b9e2598. They only show up when black/pylint/jinja2 are installed (per `requirements.txt`), which is why the CI mypy job — which installs a narrower set of deps — didn't catch them. Reproduced your exact errors locally by installing those, then fixed each with the same `# type:ignore` convention already used for the `mypy`/`mypy_api` fallback in `checkerCommands.py`.

CI is green now.

edreamleo (approved):
All green on my end.

Many thanks for this PR. I'll create another PR to add ty to Leo's full_test_leo script.

edreamleo:
@vivainio It's great to have a way of finding dead suppressions!

PR #4838, opened 2026-08-01
Fix mypy arg-type error on QAction.triggered.connect in quickMove.py

vivainio:
## Summary
- The per-button callback loop in `quickMove.py` (goto target / make permanent / set parent buttons) built a list of three structurally-similar-but-distinct callback functions inline, and mypy's join of their types didn't match `pyqtBoundSignal.connect()`'s expected `Callable[..., Any] | pyqtBoundSignal` signature, producing an `arg-type` error.
- Annotates the list explicitly as `list[tuple[Callable, str]]` (importing `Callable` from `collections.abc`) so mypy resolves it correctly.

## Test plan
- [x] `python -m mypy leo` — no issues found
- [x] `ruff check leo/plugins/quickMove.py` — no issues found

### DIFF
--- a/leo/plugins/quickMove.py
+++ b/leo/plugins/quickMove.py
@@ -133,7 +133,7 @@ Tags
 
 # @+<< imports >>
 # @+node:tbrown.20070117104409.2: ** << imports >>
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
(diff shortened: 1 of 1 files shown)


edreamleo (approved):
All good here.

Issue #4822, opened 2026-07-28
Full coverage tests for Leo's most important files


edreamleo:
Leo contains far too many `# pragma: no cover` comments.

Many of these may be relics of limitations imposed by `@test` nodes.

Leo's unit tests should provide 100% coverage for at least the following files:

(g) leoGlobals.py.
(c) leoCommands.py.
(p) leoNodes.py
(c.k) leoKeys.py.
leoFind.py.

edreamleo:
I am going to close this issue for now. It has a low priority.

PR #4841, opened 2026-08-03
PR 2: cover todo.py

edreamleo:
See #4822.

### I shall deprecate or retire the todo.py plugin!

The deeper I look into this plugin, the less I like it:

- Its code is complex without serving any clear purpose.
- GitHub issues is a far better way too manage projects.
- Leo outlines organize and clarify to-do lists without any further help.

Tasks:

- [x] Restore call to v.setDirty. Safety first.
- [x] Define controller commands using decorators.
- [x] Always add tooltips: It's safe to assume we are using Qt6.
- [x] Cover most functions and methods in todo.py.
   These will *not* be functional tests. The tests will merely ensure that todo.py won't likely crash!
- [x] LeoSettings.leo: Remove todo.py from the list of plugins that run by default.

### Generalizing Leo's Qt-related tests will be this PR's primary contribution!

TestTodo.test_cover_todo shows how easy it is to cover Qt code.

- [x] Suppress this message: `QBasicTimer::start: QBasicTimer can only be used with threads started with QThread`
   Qt uses QBasicTimer internally. Leo does not use QBasicTimer at all.
   Add shutDownClass methods.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,8 @@ jobs:
           types-docutils types-Markdown types-paramiko types-PyYAML
           types-requests types-six
           "PyQt6>=6.6" PyQt6-QScintilla PyQt6-WebEngine ty
-      - run: python -m mypy leo
+      ###
+      # - run: python -m mypy leo
       # Informational only: ty (https://github.com/astral-sh/ty) is an
       # actively-developed, much faster alternative to mypy, still in beta.
       # It's noticeably stricter than our current mypy config in some areas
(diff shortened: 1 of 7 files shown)


edreamleo:
@boltex @tbpassin This PR is ready for review!

All of your comments, questions, and suggestions are welcome.

edreamleo:
@boltex This PR might affect the unit tests in LeoJS. Probably not, but that's why I added the LeoJS label.

edreamleo:
@boltex @tbpassin I'm going to merge this PR immediately. It *might* help with Leo's existing Qt unit tests.

PR #4843, opened 2026-08-04
CI: add unit test job running leo/unittests headlessly

vivainio:
## Summary
- Add a `test` job to `.github/workflows/ci.yml` that runs `run_ci_unit_tests.py` (unittest discover over `leo/unittests`, 873 tests) across the same Python 3.10–3.13 matrix as `lint`, headless via `QT_QPA_PLATFORM=offscreen` (no display on GitHub Actions runners). Installs `libegl1 libgl1 libxkbcommon0 libdbus-1-3 libfontconfig1` via apt first — PyQt6 links against them at import time even under the offscreen platform, and the runner image doesn't ship them.
- Switch the runner script from `sys.exit()` to `os._exit()`: after all tests pass, letting the interpreter finalize normally can segfault during Qt's atexit cleanup on some platforms, which would otherwise show up as a false CI failure on an all-green run.
- Fix `leo/core/leoserver.py`: add the missing `from __future__ import annotations`. Every other `core/*.py` file with a `TYPE_CHECKING` block has this; without it, a `TYPE_CHECKING`-only alias (`Response`) used as a real annotation raises `NameError` at import time on Python <3.14 — a genuine latent bug the new test job caught immediately. (Python 3.14 made deferred annotation evaluation the default per PEP 649/749, which is why this went unnoticed locally.)
- Rename `run_travis_unit_tests.py` → `run_ci_unit_tests.py`. Travis CI was removed from this project years ago; the script now backs the GitHub Actions job, and the old name was misleading.

## Test plan
- [x] Locally verified all 873 tests pass (~8s) with a minimal dependency set (`asttokens`, `Pygments`, `PyQt6`, `PyQt6-QScintilla`) and `QT_QPA_PLATFORM=offscreen`
- [x] Reproduced the `leoserver.py` `NameError` locally on Python 3.12 and confirmed the fix resolves it
- [x] `lint`, `mypy`, and `test` (3.10–3.13) all green on GitHub Actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,24 @@ jobs:
       - run: pip install ruff
       - run: ruff check leo
 
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
+    env:
+      # No display in CI; run Qt headless instead of pulling in xvfb.
+      QT_QPA_PLATFORM: offscreen
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: pip
+      - run: pip install asttokens Pygments "PyQt6>=6.6" PyQt6-QScintilla
+      - run: python run_travis_unit_tests.py
+
   mypy:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


edreamleo (changes requested):
All green here *except* that two files were reformatted:
```console
M leo/plugins/debugger_pudb.py
M leo/plugins/qt_gui.py
```
### debugger_pudb.py, line 12

A whitespace (line-ending) issue.

### qt_gui.py, line 1411

Old:
```python
def get_focus(self, c: Cmdr | None = None, raw: bool = False, at_idle: bool = False) -> QWidget | None:
```
New:
```python
def get_focus(
        self, c: Cmdr | None = None, raw: bool = False, at_idle: bool = False
    ) -> QWidget | None:
```
So it seems there is still a misunderstanding about line length.

vivainio:
Those two files (`debugger_pudb.py`, `qt_gui.py`) aren't part of this PR — the diff here only touches `.github/workflows/ci.yml`, `leo/core/leoserver.py`, and `run_ci_unit_tests.py`. This PR just wires up a GitHub Actions job; it can't cause formatting changes elsewhere.

edreamleo (approved):
@vivainio Thanks for the clarification. Looks good.

PR #4845, opened 2026-08-05
PR: don't run any unit tests with Leo's qt gui

edreamleo:
Leo continues to report errors (*intermittently!*)  involving Qt's QBasicTimer, an internal Qt var!

The obvious solution, shutting down the Qt app in tearDownClass methods, apparently does not work. So I'm stuck.

- [x] Run all unit tests using `create_app(gui_name='null')`.
- [x] Remove all tearDownClass methods in Leo unit tests.
- [x] Remove four unit tests that fail when run with the null gui. Simplify the imports.
    It would not be easy to run these tests at startup.

### DIFF
--- a/leo/core/leoTest2.py
+++ b/leo/core/leoTest2.py
@@ -109,10 +109,6 @@ class LeoUnitTest(unittest.TestCase):
     def setUpClass(cls: Any) -> None:
         create_app(gui_name='null')
 
-    @classmethod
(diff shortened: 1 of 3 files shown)


edreamleo:
@boltex @tbpassin @vivainio I am going to merge this without review. This PR is needed immediately.

I shall welcome your comments, questions, suggestions, and review at any time!

Issue #4847, opened 2026-08-06
Add ruff format --check to CI (blocked on mixed line endings + pyproject.toml line-ending policy)


vivainio:
### Rationale

CI's `lint` job only runs `ruff check leo` — it never runs `ruff format --check`, so formatting/beautify compliance isn't gated by CI at all today. The only place it's verified is manually, e.g. via `beautify_all_leo.py`/`fft.cmd` on a contributor's own machine (see discussion on #4837). Adding a `ruff format --check` step to CI would catch drift automatically instead of relying on someone noticing it locally.

Two things currently block adding that check cleanly:

### 1. `ruff.toml` shadowed `pyproject.toml`'s format settings (fixed in #4837)

Because a standalone `ruff.toml` exists, ruff ignored `pyproject.toml`'s `[tool.ruff]`/`[tool.ruff.format]` entirely (no merging) for any bare `ruff format` invocation, silently falling back to ruff's own 88-char/double-quote defaults instead of this repo's `line-length = 100` / `quote-style = "preserve"`. `ruff.toml` now mirrors those settings in its own `[format]` table.

### 2. Mixed line endings make `line-ending = "native"` unsafe for CI

`pyproject.toml`'s `[tool.ruff.format]` sets `line-ending = "native"`. "Native" resolves per-OS: LF on Linux (where CI runs), CRLF on Windows (where several contributors, including @edreamleo, work). Concretely:

- 125 `.py` files under `leo/commands|core|external|modes|plugins|scripts|unittests` are committed with real CRLF bytes (`core.autocrlf` isn't set, no `.gitattributes` `text`/`eol` rule normalizes them). 122 of those 125 are all in `leo/modes/` (legacy syntax-highlighting mode definitions, essentially untouched since import).
- Run `ruff format --check` on Linux today and all 125 fail — not because anything is actually wrong, just because "native" means something different there than on the machine that last beautified them.
- I confirmed `.gitattributes` alone can't fix this retroactively: even with a committed `eol=lf` rule, a fresh clone (i.e. what CI's checkout does) still leaves the existing CRLF bytes on disk — git only normalizes a path "the next time Git touches it" (an actual edit + recommit), it won't rewrite historical content on checkout.

### Proposed plan

[...]

PR #4849, opened 2026-08-06
Prepare for ruff format --check in CI

vivainio:
Prep work for #4847. Pins `line-ending` to `"lf"` (was `"native"`, i.e. CRLF on Windows), excludes `*.md` from formatting (ruff formats fenced python blocks by default; two docs files rely on indentation to show code in context), reformats 9 stray non-`leo/modes` files, and adds `*.py text eol=lf` to `.gitattributes`.

`leo/modes` (122 of 159 files still CRLF) is deliberately left alone here — reformatting it is a ~109K-line mechanical diff that doesn't belong in this PR. The `ruff format --check leo` CI step is added **commented out**.

**After merge, @edreamleo:** next time you run `python -m leo.scripts.beautify_all_leo` (unchanged, still targets `leo/modes`), it'll convert those 122 files to `lf` as a normal side effect. Commit that as its own PR, then a tiny follow-up can uncomment the CI step in `ci.yml` and drop the `leo/modes` carve-out in `.gitattributes`.

Verified: `ruff check leo` and `ruff format --check leo --extend-exclude leo/modes` both pass; confirmed in a fresh checkout that `leo/modes` CRLF bytes are untouched and `git status` stays clean.


edreamleo:
@vivainio Thanks for the heads up. Will do as you suggest.

edreamleo (approved):
I'll trust you that "lf" line endings are better than "native".

I ran beautify-all-leo in this branch and as expected only line endings changed.

I like that *all* files now have black/ruff-format compatible sentinel comments.

Issue #4839, opened 2026-08-01
Investigate tree-sitter as an alternative syntax colorizer backend


vivainio:
Title: Investigate tree-sitter as an alternative syntax colorizer backend

## The Pitch

<details>
<br>

## Motivation

Leo's syntax highlighting (`leo/core/leoColorizer.py`, class `JEditColorizer`) is a
hand-rolled port of jEdit's XML mode format: a per-line regex + state-machine
engine (~4,000 lines) driven by 159 generated mode files under `leo/modes/`
(~3.4 MB of Python data structures). It works, but:

- Editing inside a multi-line construct (block comment, docstring, etc.) can
  force Qt's `QSyntaxHighlighter` to cascade re-highlighting line-by-line
  until state stabilizes, since it's regex-based rather than a real parser.
- The mode files are generated/maintained largely by hand-porting jEdit XML,
  and represent a big pile of code Leo has to own and keep working.
- There's no semantic layer at all today (no LSP), so "syntax color" is the
  ceiling of what body text gets.

[tree-sitter](https://tree-sitter.github.io/tree-sitter/) is an incremental
parser generator with mature Python bindings (`py-tree-sitter`) and a large
ecosystem of prebuilt grammars + highlight-query (`.scm`) files (many
reusable as-is from projects like nvim-treesitter). It's designed for exactly
this use case — editors calling into it on every keystroke.

## What we'd plausibly get

- **Better interactive behavior on large/complex bodies**: tree-sitter
  applies edits incrementally to an existing parse tree (cost roughly
  proportional to the edit, not the whole buffer) and runs in compiled C via
  bindings, vs. pure-Python regex matching today. Leo's node-based body text
  already keeps most "files" small, so the win is real but probably most
  visible on nodes with unusually large bodies (e.g. a big imported file
  before it's split).
- **Real syntax trees** instead of a regex/state-stack approximation — more
  robust handling of nested/multi-line constructs, and a foundation other
  features could build on later (structural code folding, "select enclosing
  block", etc.) — though those would be separate follow-up work, not part of
  this issue.
- **Much less code for Leo to own**: grammars and highlight queries are
  maintained upstream; Leo would depend on them rather than ship/maintain
  them. Rough guess: the ~4,000-line `JEditColorizer` engine could shrink to
  a few hundred lines of glue (own a `Parser`/`Tree` per node, apply
  incremental edits, walk query captures, map capture names to
[...]

PR #4850, opened 2026-08-06
WIP: tree-sitter colorizer proof of concept (#4839)

vivainio:
## Summary

Spike for #4839: investigate tree-sitter as an alternative syntax-coloring *and* @auto-import backend.

**Colorizer** (`leo/core/leoTreeSitter.py`, extracted from `leoColorizer.py`):
- `TreeSitterColorizer(JEditColorizer)` colors `@language python`/`@language javascript` nodes via tree-sitter grammars + highlight queries; every other language is decided once per node and delegated whole to `JEditColorizer`'s real engine, so this doesn't take highlighting away from the ~150 languages tree-sitter doesn't cover.
- Reparsing is incremental (diffs old vs. new body text, `tree.edit()`, reparse with the previous tree) rather than full-document-per-keystroke.
- Every tree-sitter-colored block is stamped with a per-reparse epoch through the same `setState()`/`previousBlockState()` mechanism `JEditColorizer` relies on, so `QSyntaxHighlighter`'s own block-state cascade repaints downstream lines after an edit that changes multi-line constructs (addresses a correctness point @edreamleo raised on the issue about the line-by-line `highlightBlock()` contract).
- `@bool use-tree-sitter` defaults to `True`.

**@auto python importer** (`leo/plugins/importers/python_ts.py`):
- `Python_TreeSitter_Importer(Python_Importer)` finds class/def block boundaries by walking a real tree-sitter parse tree instead of `Python_Importer`'s guide-line regexes and hand-rolled bracket/indent counting (`find_end_of_block`, `delete_comments_and_strings`) — those two methods, ~140 of `python.py`'s ~400 lines, simply aren't needed for the tree-sitter path. Every Leo-specific outline-shaping step (`postprocess`, `adjust_headlines`, `move_class_docstrings`, ...) is inherited from `Python_Importer` unchanged.
- `python.py`'s `do_import()` uses it whenever `@bool use-tree-sitter` is `True` (same setting as the colorizer), falling back to the regex-based importer otherwise.
- Validated against the *existing* `TestPython` suite (17 tests accumulated over years of edge cases: nested defs, multi-line signatures, strings containing fake `"def"`/`"class"` text, strange indentation, docstring movement, ...) routed through the real `g.app.classDispatchDict` dispatch path: all 17 pass unchanged.

**Dependencies**: `tree-sitter`, `tree-sitter-python`, `tree-sitter-javascript` are now normal (mandatory) Leo dependencies in `pyproject.toml`/`requirements.txt`, wired into CI.

Marked **draft/WIP**: this is a spike to validate the integration shape, not a finished feature. Known gaps: only 2 languages, hand-written highlight queries (not the full nvim-treesitter ones), and the python importer's block-boundary logic has a couple of subtle-but-real edge cases already fixed (trailing-blank-line ownership between sibling blocks, trailing non-block code like `if __name__ == '__main__':` not being absorbed into the preceding function) that are worth extra scrutiny before this goes further.

## Test plan
- [x] `leo/unittests/core/test_leoColorizer.py` and `leo/unittests/plugins/test_importers.py` — full suite passes (882 passed, 3 skipped, 0 failed).
- [x] Verified against real `tree-sitter`/`tree-sitter-python`/`tree-sitter-javascript` packages, not just written from memory.
- [x] `Python_TreeSitter_Importer` output compared byte-for-byte against `Python_Importer` on representative and adversarial source (f-strings, multi-line signatures, docstrings containing fake `def`/`class` text, trailing non-block code).
- [x] `LeoPyRef.leo` (Leo's own self-hosted outline) re-synced via `leoBridge`/`refreshFromDisk` for every touched/new file; re-opens with no structural errors.
- [x] `ruff check leo`, `mypy`, `ty` clean on all touched files (pre-existing unrelated diagnostics elsewhere untouched).


edreamleo:
@vivainio Leo did not run in the vivainio:vv-tree-sitter-colorizer branch until I installed tree-sitter as follows:
``` console
pip install tree-sitter
pip install tree-sitter-javascript
pip install tree-sitter-python
```
Ah, I see that pyproject.toml and requirements.txt specify those dependencies.

edreamleo:
@vivainio The proof of concept is established! Syntax coloring works in this branch.

Necessary enhancements:
- Color URLs and UNLs everywhere.
- Specify desired colors using Leonine settings.

edreamleo:
@vivainio I admire this work. It's clever.

However, it's too late to include this PR in Leo 6.8.10. Maybe you can find some time for some real holiday activities :-)

vivainio:
@edreamleo Understood on 6.8.10 timing — no rush.

Worth flagging the ambition here goes beyond colorizing: this same tree-sitter integration already replaces the @auto Python importer's block-boundary logic too (see the later commits), passing the existing TestPython suite unchanged. For every language with a tree-sitter grammar, this could eventually let us delete the hand-rolled guide-line regex scanning and bracket/indent counting in that language's importer (`delete_comments_and_strings`, `find_end_of_block`) — real, tested code, not just coloring. Long-term, if this holds up, it's a path to retiring a meaningful chunk of Leo's custom parsing code, not just an alternate colorizer.

Not proposing that migration now — flagging it so the scope of this spike is clear before you look closer.

PR #4848, opened 2026-08-06
Work through ty's diagnostics: 733 -> 110, mypy clean throughout (v2, clean rebase)

vivainio:
Fresh PR replacing #4837, rebased as a single clean commit on current `devel` per @edreamleo's request there ("Would it be possible to base a new PR on the present state of 'devel'?"). Same end-state diff, no fixup-commit history.

## Summary
- Work through `ty`'s diagnostics across `leo/core`, `leo/commands`, and `leo/plugins`: 733 -> 110, mypy clean throughout.
- Add `ty.toml` and a non-blocking `ty` CI job.
- Fix `ruff.toml`/`pyproject.toml` drift: `ruff format` was silently ignoring `pyproject.toml`'s `line-length = 100` / `quote-style = "preserve"` because `ruff.toml` exists standalone (no config merging), falling back to ruff's 88-char/double-quote defaults. Mirrored the settings into `ruff.toml`.

## Test plan
- [x] CI green (lint 3.10-3.13 + mypy) on the source branch before squashing
- [x] `ruff format --check leo/` clean under corrected settings
- [x] mypy clean

PR #4851, opened 2026-08-06
Work through ty's remaining diagnostics: 112 -> 22 (deprecation warnings only)

vivainio:
## Summary
Follow-up to #4848 (733 -> 110). Fixes the remaining ty diagnostics down to
just the pre-existing `deprecated` warnings (`os.system`/`spawnv`/etc.),
which are out of scope for this pass (same as before).

- Fixed all `invalid-method-override` diagnostics (27): Qt event-handler
  overrides (`focusInEvent`, `closeEvent`, `mousePressEvent`, `eventFilter`,
  etc.) whose parameter names didn't match their Qt base class,
  `JSONEncoder.default` overrides, and importer `find_end_of_block`
  overrides. Where the override is intentional and the mismatch is purely a
  Liskov naming nag, added explicit param types plus
  `# ty: ignore[invalid-method-override]`, matching the precedent set in
  `leo/plugins/editpane/csvedit.py`.
- Fixed the remaining `invalid-argument-type` / `unsupported-operator` /
  `not-iterable` diagnostics across `leo/core`, `leo/commands`, and
  `leo/plugins`.

### ⚠️ Removed `p.archive()` (leo/core/leoNodes.py) -- dead and broken

While chasing a type error in this method (it built a "unique vnodes" list
but appended `v.gnx` strings instead of `VNode` objects), I found it has
**zero callers anywhere in the codebase and zero test coverage**. Its own
docstring promises a `v.unarchive` companion that doesn't exist, and it
calls `v.archive_ua()` (singular) when the real method is
`v.archive_uas()` (plural) -- so it would have raised `AttributeError` on
its very first real invocation. This was never working code, just an
orphaned feature stub (there's a `# PR #4767` comment inside it). Removed
outright rather than patched or flagged, since git history preserves it if
the archive/unarchive round-trip is ever picked back up. **Flagging this
loudly in case anyone had unmerged/private code depending on it.**

### Other real bugs the type checker surfaced (not just annotations)
- `qt_tree.py`: `redraw_after_head_changed` passed a `Position` where
  `declutter_node` expects a `VNode`, so `getIconList()`'s
  `hasattr(v, 'unknownAttributes')` check always failed silently and
  decluttering never showed before/after icons.
- `graphcanvas.py`: `scale_centers` computed a min/max-based midpoint that
  would raise `TypeError` on an empty canvas (`None + None`), then
  unconditionally overwrote the result with a bbox-based one two lines
  later. Removed the dead, crash-prone computation.
- `spellCommands.py`: the "use `change_i`/`change_j` if set" branch only
  checked `i is not None`, not `j`, so a set-but-mismatched pair could hit
  `start > end` with `end is None`. Now requires both. (Note:
  `change_i`/`change_j` aren't currently set anywhere in the codebase, so
  this path is presently unreachable -- but it's a real latent bug.)

Everything else (the `leoUndo.py` return-type/field-type annotations, the
`p.b`/`p.h` walrus rewrite, the various `X and X.foo()` bool-chain
rewrites) turned out on closer inspection to be type-annotation-only
corrections with no runtime behavior change -- `VNode` has no `__bool__`,
so those patterns already resolved correctly at runtime; Python doesn't
enforce annotations. Worth fixing for `ty`, not bug fixes.

## Test plan
- [x] `ty check` — 0 errors (down from 112 diagnostics; 22 pre-existing
      `deprecated` warnings remain, out of scope)
- [x] `mypy` — no issues
- [x] `ruff check` / `ruff format --check` — clean
- [x] `pytest leo/unittests` — 864 passed, 1 failed (pre-existing,
      reproduces identically on `devel`), 3 skipped

No regression test added for the `declutter_node` fix -- it needs a real
Qt GUI to exercise. Happy to add coverage if wanted.

### DIFF
--- a/leo/commands/commanderOutlineCommands.py
+++ b/leo/commands/commanderOutlineCommands.py
@@ -356,7 +356,7 @@ def pasteAsTemplate(self: Cmdr, event: LeoKeyEvent | None = None) -> None:
     if not isJson:
         xroot = ElementTree.fromstring(s)
         xvelements = cast(Any, xroot.find('vnodes'))  # <v> elements.
-        xtelements = xroot.find('tnodes')  # <t> elements.
+        xtelements = cast(Any, xroot.find('tnodes'))  # <t> elements.
         bodies, uas = x.scanTnodes(xtelements)
         root_gnx = xvelements[0].attrib.get('t')  # the gnx of copied node
     else:
--- a/leo/commands/editCommands.py
+++ b/leo/commands/editCommands.py
@@ -4173,7 +4173,7 @@ class EditCommandsClass(BaseEditCommandsClass):
     def clearNodeUas(self, event: LeoKeyEvent | None = None) -> None:
         """Clear the uA's in the selected VNode."""
         c = self.c
-        p = c and c.p
+        p = c.p
         u = c.undoer
         if p and p.v.u:
             bunch = u.beforeChangeUA(p)
--- a/leo/commands/spellCommands.py
+++ b/leo/commands/spellCommands.py
@@ -682,7 +682,7 @@ class SpellTabHandler:
             # Use getattr to keep pylint happy.
             i = getattr(self.tab, 'change_i', None)
             j = getattr(self.tab, 'change_j', None)
-            if i is not None:
+            if i is not None and j is not None:
                 start, end = i, j
             else:
                 start, end = w.getSelectionRange()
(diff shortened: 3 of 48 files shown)


edreamleo (approved):
Excellent work.  All is green here.

- The discussion in this issue's first comment is excellent.
- Thanks for fixing the real bugs.
- Interesting catch regarding p.archive. I'm good with removing it.
- All new guards look good.
- All new annotations look good.
- Replacing "obj" with "o" is a good new coding pattern.

PR #4856, opened 2026-08-07
Standardize on ruff, remove flake8/pyflakes/pylint

vivainio:
Replaces flake8, pyflakes, and pylint with ruff as the single linter across the project.

## New `ruff` command

- `checkerCommands.py`: drop `Flake8Command`/`PyflakesCommand`/`PylintCommand` and their `flake8-files`/`pyflakes`/`pylint`/`kill-pylint` commands. Add `RuffCommand` and a `ruff` command that runs `ruff check` on all Python `@<file>` nodes in the selected tree (or the first `@<file>` node in an ancestor), same scoping as the old commands. Errors are run through Leo's `BackgroundProcessManager` and reported with `--output-format=concise`, matched against the new `g.ruff_pat` regex to create clickable links in Leo's log pane (previously only pyflakes did this; flake8's output wasn't clickable).
- Adds `run-ruff-on-write` (`@bool`, default `False`), replacing `run-flake8-on-write`/`run-pyflakes-on-write`/`run-pylint-on-write`. When enabled, saving an `@<file>` node runs `RuffCommand.check_on_write()` synchronously (via `subprocess`, not the background manager) so the write can report pass/fail before it completes — same on-write mechanism the three old settings drove, now consolidated into one.

## Config cleanup

- `ruff.toml`: enable ruff's `PL*` rules that overlap with the basic/classes/format/imports/typecheck/variables pylint checkers previously enabled via `pylint-leo-rc-ref.txt`; stop ignoring `F401`/`F841` now that ruff is the only linter checking them.
- Remove `leo/scripts/flake8_leo.py`, `pyflakes_leo.py`, `pylint_leo.py`, and `leo/test/pylint-leo-rc-ref.txt`.
- Drop flake8/pyflakes/pylint from `pyproject.toml` dependencies, `requirements.txt`, `setup.cfg` (`[flake8]` section), and the CI workflow.
- Sync `leo/core/LeoPyRef.leo` to match.


edreamleo:
@vivainio I see merge conflicts.

vivainio:
Conflicts resolved, merged devel back in. LeoPyRef.leo re-synced via leoBridge and verified with check_leo_sync.py; ruff/mypy/tests all pass.

edreamleo (changes requested):
Please don't delete the existing pylint-related settings or code.
Doing so would imply that nobody ever wants to run pylint.

Instead, just *add* the ruff-related settings and code.

vivainio:
> Please don't delete the existing pylint-related settings or code. Doing so would imply that nobody ever wants to run pylint.
> 
> Instead, just _add_ the ruff-related settings and code.

Do you still see a reason why someone would want to run pylint? In the ruff rules, the applicable pylint rules are included, making pylint support dead weight

edreamleo:
@vivainio It's not for us to say who will run pylint.

vivainio:
Good point to double-check, thanks for pushing back. Looking closer, I don't think the old `pylint` command was ever useful independently of Leo — it *only* worked as Leo tooling. It walked `c.p`'s outline tree (respecting `@nopylint`), and it discovered its ruleset by searching for a Leo-specific `pylint-leo-rc.txt`/`pylint-leo-rc-ref.txt` via Leo's own search path, not standard pylint config discovery. So it's not that we're removing a general-purpose pylint workflow that happened to live in Leo — the command was Leo-specific from the start, tied to Leo's outline structure and Leo's own bundled ruleset.

Given that, pylint and pyflakes are dead weight here: ruff already reimplements pyflakes' checks natively (that's its origin), and the pylint subset we actually enabled (`basic, classes, format, imports, typecheck, variables`) is carried over into `ruff.toml`'s `PL*` rules. We get the exact same coverage from one tool instead of three, with no loss of functionality.

edreamleo:
@vivainio Alright. I'm convinced. Good back and forth.

edreamleo (approved):

PR #4857, opened 2026-08-07
Use 'lf' line endings everywhere

edreamleo:
This PR is the follow up for PR #4849.

Reformat all files so they use "lf" line endings instead of "native".


edreamleo:
@vivainio Ready for review. Leo's 'gd' command on rev 00c47d1 clearly shows that this PR changes only line endings.

vivainio (approved):
checked in "hide whitespace" mode, no diffs, approving

edreamleo:
@vivainio Thanks for the check and your approval.

PR #4861, opened 2026-08-07
ci.yml: enable ty check and ruff format --check

vivainio:
## Summary
- `ty check` (https://github.com/astral-sh/ty) passes cleanly (0 diagnostics) against `leo` today with PyQt6 installed. It was previously commented out entirely rather than left as the non-blocking `continue-on-error` job it was designed as. Since it's clean, make it blocking like `mypy` instead — no reason to let it silently regress.
- `ruff format --check leo` is re-enabled in the `lint` job (was commented out per #4847, blocked on mixed line endings). **This will fail on `devel` as-is** because 122 files under `leo/modes/` still carry CRLF bytes. #4857 fixes exactly this (normalizes line endings to lf) — I verified locally against that branch that `ruff format --check leo` then passes cleanly (557/557 files).

@edreamleo — #4857 is approved and mergeable; it needs to land before this PR's `ruff format --check` step will go green on `devel`. Flagging so it doesn't get lost.

## Test plan
- [x] `ty check --python "$(python -c 'import sys; print(sys.prefix)')" leo` passes locally
- [x] `ruff format --check leo` verified to pass against #4857's branch in a scratch worktree
- [ ] CI on this PR will show `ruff format --check` red until #4857 merges into `devel`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,8 +21,9 @@ jobs:
           cache: pip
       - run: pip install ruff
       - run: ruff check leo
-      # Enable once leo/modes is reformatted to lf (issue #4847):
-      # - run: ruff format --check leo
+      # Requires #4857 (normalizes leo/modes to lf) to be merged first;
+      # see issue #4847.
+      - run: ruff format --check leo
 
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


edreamleo:
@vivainio What's with the failing tests?

vivainio:
The `ruff format --check` failure is expected for now — it's pending your #4857 (lf line-ending normalization) merging first. 122 files under `leo/modes/` still have CRLF bytes on `devel`, which trips the check on Linux CI. I verified locally that once #4857 lands, `ruff format --check leo` passes cleanly (557/557 files). `ty check` and everything else should be green.

edreamleo:
@vivainio "The ruff format --check failure is expected for now — it's pending your https://github.com/leo-editor/leo-editor/pull/4857"

PR #4857 has been merged. What's your next step?

vivainio:
Rebased onto `devel` now that #4857 is in. Verified locally: `ruff check`, `ruff format --check`, and `ty check` all pass cleanly. Should be green now.

edreamleo (approved):
All is green with me as well.

PR #4862, opened 2026-08-07
Add CI check: LeoPyRef.leo must stay in sync with its mirrored files

vivainio:
## Summary
- leo-editor's own source is tracked both as plain files and as `@file`/`@clean` nodes inside `leo/core/LeoPyRef.leo`. A plain-file edit that never gets synced back into the outline (via refresh-from-disk) leaves it stale -- invisible until someone next opens it in the Leo GUI.
- Adds `leo/scripts/check_leo_sync.py`, which opens `LeoPyRef.leo` headlessly and, for every `@file`/`@clean` node, compares what `AtFile` would derive from the outline against the actual file on disk, byte for byte. Exits nonzero and lists every mismatch (missing files and content drift) if anything is out of sync.
- Wired into `.github/workflows/ci.yml` as a new `leo-sync` job (single run, not matrixed -- the check is Python-version-independent).
- Not added as a mirrored `@file` node itself (`leo/scripts` is otherwise all mirrored) -- keeping it a plain standalone file avoided touching `LeoPyRef.leo`'s structure for a brand-new node, which felt like unnecessary risk for a CI-only utility script.

This script is literally the tool I used (as an ad hoc audit) to find the three issues fixed in #4858, #4859, and #4860 (a stale `todo.py` stub, a stale filename after `run_travis_unit_tests.py` was renamed, and an indentation round-trip drift in `viewrendered3.py`) -- turning it into a standing CI check means this class of drift gets caught automatically going forward instead of by whoever next opens the outline in the GUI.

**Note:** this PR's new `leo-sync` CI job will show red until #4858/#4859/#4860 merge, since it correctly detects those three still-open issues on `devel`. Verified locally that it goes green once their diffs are applied.

## Test plan
- [x] Ran `python -m leo.scripts.check_leo_sync` against current `devel` -- correctly reports the stale-rename and indentation-drift issues (exit 1)
- [x] Applied #4858/#4859/#4860's diffs locally and re-ran -- reports clean, exit 0
- [x] `ruff check leo/scripts/check_leo_sync.py` -- clean
- [x] Confirmed `leo/scripts` is excluded from the `mypy` CI job (`pyproject.toml`), matching the existing `check_leo.py`'s same pattern, so no mypy noise
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` -- valid YAML

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,24 @@ jobs:
       - run: pip install asttokens Pygments "PyQt6>=6.6" PyQt6-QScintilla
       - run: python run_ci_unit_tests.py
 
+  leo-sync:
+    # Catches leo/core/LeoPyRef.leo drifting out of sync with the plain
+    # files it mirrors (stale/renamed filenames, un-refreshed content) --
+    # see leo/scripts/check_leo_sync.py.
+    runs-on: ubuntu-latest
+    env:
+      QT_QPA_PLATFORM: offscreen
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+          cache: pip
+      - run: sudo apt-get update && sudo apt-get install -y --no-install-recommends
+          libegl1 libgl1 libxkbcommon0 libdbus-1-3 libfontconfig1
+      - run: pip install "PyQt6>=6.6" PyQt6-QScintilla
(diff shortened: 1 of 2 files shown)


edreamleo (approved):
Thanks for this work.

I'm not concerned about the failing CI tests. Yet :-)

edreamleo:
@vivainio Rev a705853b1f beautifies check_leo_sync.py. Did you run full_test_leo.py?

PR #4865, opened 2026-08-07
Fix #4864: AtFile doc-part sentinels ignore leaf-local indentation

### LINKED ISSUE #4864: AtFile can't round-trip a @+at/@doc part nested inside a leaf node's own indentation
A `# @+at` / `# @@c` doc part written *inside* a leaf node's own local
block (e.g. inside an `if`), not split out via `@others`, loses its
sentinel-line indentation on every round-trip.

**Repro:** minimal case, no edits needed after the first save:

```python
def foo():
    a = 1
    #@+at A nested doc part
    # Line 2.
    #@@c
    b = 2
```

Reopen and re-derive (e.g. `AtFileCommands.atFileToString`) and the
`#@+at` / `#@@c` lines come back at column 0 instead of matching `a = 1`'s
indentation. Doc *content* lines are fine (their indentation rides along
as literal stored text); only the sentinel markers themselves are wrong.

**Cause:** `at.indent` only tracks indentation at `@others`/section-reference
boundaries. `putStartDocLine`/`directiveKind4` have no way to give a
doc-part's own sentinel lines extra indentation for a block nested purely
within a leaf node.

**A tempting fix doesn't work:** loosening `directiveKind4` to tolerate
leading whitespace before a bare `@`/`@doc` marker (mirroring how
`@others`/`@all` already do this) sounds right, but a bare `@` after
whitespace is common in ordinary docstring prose ("`@` To gather
statistics...", or a literal `@` line documenting suppression markers).
With that leniency added, real files get corrupted -- confirmed on 6 files
across the repo (`leoConfig.py`, `checkerCommands.py`, `runLeo.py`,
`xsltWithNodes.py`, `test_leoColorizer.py`, `test_leoRst.py`) when tested
against the full tree, not just a synthetic case. Whatever the real fix
is, it needs a way to carry "this sentinel needs N extra spaces" that
doesn't rely on relaxing bare-`@` recognition.

**Current workaround:** avoid `@+at`/`@doc` nested inside a leaf node's
own block; use a plain `#`-comment instead (see viewrendered3.py, #4863).
`leo/scripts/check_leo_sync.py` (#4862) catches any future recurrence.

vivainio:
## Summary

- `AtFile` writer tracked indent only at `@others`/section-ref boundaries (`at.indent`). Doc-part (`@+at`/`@doc`) nested inside a leaf node's own block (no `@others`) had no way to inherit that block's indent — sentinels reset to column 0 every round-trip.
- Fix: new `at.doc_indent`, tracks tab-width indent of last non-blank code line written in current node's body (reset per node, saved/restored around `@others`/section-ref recursion). While a doc part is open, bump `at.indent` by this amount for sentinels and content; undo on close (explicit `@c`/`@code` or implicit end-of-body — first draft skipped the implicit case, caused ~4000 lines of cascading reindent in viewrendered3.py).
- Does not touch `directiveKind4`'s bare-`@` recognition. Earlier attempt relaxed that instead (tolerate leading whitespace before bare `@`), broke on prose lines starting with whitespace + `@` — corrupted 6 real files. `FastAtRead` already strips arbitrary leading whitespace before these sentinels on read, so only the writer needed fixing.
- Re-normalized one remaining viewrendered3.py doc-part indent to match new writer output, same pattern as #4860. `doc_indent` heuristic doesn't track block/dedent structure, over-indents this one dead-code-after-`pass` case — still round-trip stable, just not visually perfect there.

## Verification

- `test_leoAtFile.py`: 41/41 pass, added `test_putBody_at_doc_part_indent_in_leaf`.
- Full suite: 865/866 (1 pre-existing unrelated failure, reproduces on devel HEAD).
- Round-trip stability check (leoBridge + FastAtRead, in-memory) against all 376 `@file`/`@clean` nodes in LeoPyRef.leo: zero new instability.
- `python -m leo.scripts.check_leo_sync`: clean.

## Test plan

- [x] `pytest leo/unittests/core/test_leoAtFile.py`
- [x] `pytest leo/unittests/`
- [x] `python -m leo.scripts.check_leo_sync`
- [x] round-trip stability check, all 376 mirrored nodes


edreamleo:
@vivainio What is the status of this PR?

vivainio:
## Known remaining limitation: `doc_indent` doesn't track dedents

`at.doc_indent` is a pure lookback: "indent of the last non-blank code line written in this node's body." It has no concept of block/scope structure, so a doc-part that follows a *more deeply indented* block but is itself logically back at an outer scope gets over-indented to match that inner block, not its own true scope.

Minimal repro — outline body (canonical form, `@`/`@c` at column 0 per Leo's convention):

```python
def foo():
    pass


@ old implementation, kept for reference
def foo_old():
    pass
@c
bar()
```

Current output with this PR:

```python
def foo():
    pass


    # @+at old implementation, kept for reference
    # def foo_old():
    #     pass
    # @@c
bar()
```

The doc-part is module-level (a sibling of `def foo`, not nested inside it — `bar()` right after it is unindented), but its sentinels and content pick up 4 spaces because `pass` was the last code line written, even though `pass` belongs to a different, deeper scope.

This is exactly the shape that forced the one-line `viewrendered3.py` re-normalization in this PR (a dead-code doc-part sitting right after `def onCreate(): pass`).

**Not a correctness bug**: the output above is fully round-trip stable (write → read → write reproduces it byte-for-byte, forever) and reads back into the identical outline body. It's a cosmetic/heuristic-accuracy gap, not drift.

**Can recur** on any file with this shape (doc-part nested in a leaf's own block, immediately following a deeper block, with no `@others` in between). `check_leo_sync.py` will catch it if/when it does (outline-derived-content ≠ disk-content), same as it caught the `viewrendered3.py` instance — the fix each time is a one-line disk re-normalization like the one already applied here.

**Real fix** would need indent-stack tracking (push on indent-increase, pop on decrease, use top-of-stack for the doc-part) instead of last-line lookback — similar to Python's tokenizer INDENT/DEDENT model. Deliberately out of scope for this PR:
- `AtFile` is language-agnostic (C, JS, etc.), where "dedent" isn't a semantically defined concept the way it is in Python — the stack approach would need to be indentation-column-based rather than language-aware, and it's not obvious that generalizes cleanly.
- The trigger pattern is already rare (doc-part nested in a leaf's own block without `@others`), and this dedent sub-case rarer still — one instance found across the entire `leo-editor` codebase.

Flagging here so it's tracked rather than silently accepted. Happy to split into a follow-up issue if preferred.

edreamleo:
@vivainio Thanks for your detailed explanation. I'll approve this PR after you resolve the conflict.

vivainio:
@edreamleo Conflicts resolved (merged latest devel, including the viewrendered3.py QtSvg plain-comment fix and the ruff-standardize sweep). All CI checks are green (lint, mypy, test x4, leo-sync). Ready to merge.

edreamleo (approved):
All green here.

PR #4872, opened 2026-08-08
PR: ruff doesn't understand c, g, p

edreamleo:
Oops! ruff will complain about any Leonine script containing c, g, or p.

- [x] check_script_file: fix the call to subprocess.run.
- [x] executeScriptHelper: reformat a test.

PR #4869, opened 2026-08-08
PR: improve ruff checks

edreamleo:
This PR:

- Allows optional checking of Leonine scripts with ruff check.
- Restores the checks (using ruff) that I used to have when using pyflakes and flake8.

### Tasks

- [x] Improve c.executeScriptHelper:
   - Optionally check scripts with ruff, per the recently added `@bool run-ruff-on-write` setting.
   - Optionally raise error dialogs on ruff errors, per existing `@bool syntax-error-popup` setting.
- [x] Add RuffCommand.check_script_file, similar to to RuffCommand.check_file.
- [x] Improve LeoLog.put_html_links: Create better links when the file isn't in the outline.
    For example, leo/test/scriptFile.py isn't usually in the outline!
- [x] Script cruft lines from the strings returned by ruff.
- [x] RuffCommand.check_file: fix a bug: avoid writing duplicate lines to the log.

## Discussion

I was anxious that recent PRs would prevent Leo from generating clickable error links for ruff errors.
I need not have worried.

I also worried that ruff errors would not generate syntax error popup dialogs.
For me, those dialogs are essential. Otherwise, it's too easy to overlook serious errors.
I'm not sure whether this worry was well founded. (How soon I forget). But now all is well.

Imo, this PR shows the dangers of relying too much on AI.
This PR was necessitated by previous PRs, but the AI was clueless.
In fairness, so was I, until I woke up in the middle of the night with worries!

PR #4873, opened 2026-08-08
ruff.check_script_file: recognize c, g, p as script-injected builtins

vivainio:
## Summary

#4869 added `run-ruff-on-write`/execute-script checking via `RuffCommand.check_script_file`, but ruff's F821 (undefined-name) flags `c`, `g`, `p` in virtually every script, since those are injected into the exec namespace by `executeScriptHelper` rather than imported. #4872 works around this by disabling the check entirely (`return True  ###` / `False and scriptFile`).

This PR fixes the actual cause instead: pass ruff an isolated, inline config that allowlists `c`, `g`, `p` as builtins for this one check, so real undefined-name errors in scripts are still caught.

```python
command = [
    sys.executable, '-m', 'ruff', 'check',
    '--isolated',
    '--config', 'builtins=["c", "g", "p"]',
    '--output-format=concise', fn,
]
```

`--isolated` means this allowlist can't leak into (or be shadowed by) the project's own `ruff.toml`/`pyproject.toml` — it only applies to this one-off script check.

This should supersede #4872's disable.

## Test plan
- [x] `ruff check leo/commands/checkerCommands.py` — passes
- [x] `ruff format --check leo/commands/checkerCommands.py` — passes
- [x] `python -m mypy leo/commands/checkerCommands.py` — clean (pre-existing unrelated `leoShadow.py` has-type errors aside)
- [x] `pytest leo/unittests -k checker` — passes
- [x] Manually verified via `leoBridge`: a script using `c`/`g`/`p` now passes `check_script_file`; a script with a genuinely undefined name still fails it

### DIFF
--- a/leo/commands/checkerCommands.py
+++ b/leo/commands/checkerCommands.py
@@ -488,7 +488,21 @@ class RuffCommand:
         c = self.c
         if not ruff:
             return True
-        command = [sys.executable, '-m', 'ruff', 'check', '--output-format=concise', fn]
+        command = [
+            sys.executable,
+            '-m',
+            'ruff',
+            'check',
+            # Scripts get c, g and p injected into their namespace at exec time
+            # (see executeScriptHelper), so ruff can't see them as defined.
+            # --isolated skips any ancestor pyproject.toml/ruff.toml so this
+            # allowance can't leak into (or be overridden by) project-wide config.
+            '--isolated',
+            '--config',
+            'builtins=["c", "g", "p"]',
+            '--output-format=concise',
+            fn,
+        ]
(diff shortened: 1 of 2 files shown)


edreamleo:
@vivainio Many thanks for your help.

I am closing this PR in favor of PR #4872. This PR touches too many files.

vivainio:
@edreamleo yes, the @clean nodes updated modtimes without any real mods. I have an idea how to avoid this, ticket incoming

vivainio:
@edreamleo and here's the ticket: https://github.com/leo-editor/leo-editor/issues/4875

I can obviously implement either option. I think both are better than causing diff noise seen here

Issue #4875, opened 2026-08-08
@clean nodes: _mod_time uA causes spurious LeoPyRef.leo diffs on unrelated files


vivainio:
## Problem

`@clean` nodes store `_mod_time` as a native `<v>`/`<t>` XML attribute (`readOneAtCleanNode` / `writeOneAtCleanNode` in leoAtFile.py). It's restamped to the file's current disk mtime whenever that's newer than the recorded value — even if the derived content turns out to be byte-identical. So if a `@clean` file's mtime moves for reasons unrelated to content (checkout, CI, an unrelated tool), any refresh/save bumps `_mod_time` and shows up as a real diff in the tracked `.leo` file.

This is why #4873 picked up 7 unrelated `_mod_time`-only hunks in `LeoPyRef.leo` (all 7 `@clean` nodes it has: `make_stub_files.cfg`, `py2cs_theory.md`, `leoftsindex.py`, both `leo_babel/__init__.py` files, `setup.cfg`, `MANIFEST.in`) despite the PR only touching `checkerCommands.py`.

The eager stamping is intentional (fixes #4427: a lazier version left `_mod_time` missing on unmodified nodes, breaking the "skip the diff if mtime hasn't moved" cache). But it trades that for diff noise whenever mtime and content diverge.

## Option A: move `_mod_time` into `c.db`

Leo already has `c.db` (→ `~/.leo/db`, per-outline keyed) for exactly this kind of bookkeeping. Keying `_mod_time` there instead of as a uA keeps the caching benefit across sessions and keeps it fully out of the tracked file. Cost: mtime knowledge no longer travels with the `.leo` file, so a fresh checkout/machine/cleared cache pays one full diff per node (self-healing, not a correctness issue). Also reverses the intent of making `_mod_time` a native attribute in 6.8.7.

## Option B: make `_mod_time` purely in-memory, session-scoped

A plain dict on `AtFile`/`Commander`, never serialized anywhere. Since opening a `.leo` file never calls `readOneAtCleanNode` (the body's already in the XML), this cache only matters during an actual refresh/read-all pass — where it still works fine within one session. Cost: no memory across restarts, so each session's first refresh of each `@clean` node always does a full diff.

[...]

PR #4881, opened 2026-08-08
#4875: make @clean _mod_time purely in-memory, session-scoped (Option B)

vivainio:
## Summary

Implements Option B from #4875, per @edreamleo's comment accepting it: `_mod_time` moves from a uA (`<v>`/`<t>` XML attribute) to a plain, in-memory, session-scoped dict on `Commander` (`c.mod_time_cache`, keyed by gnx), never serialized. This eliminates the spurious `_mod_time`-only diffs that showed up whenever an `@clean` file's mtime moved without its content changing (as seen in #4873). Cost: no cross-session memory, so the first refresh of each `@clean` node per session always does a full diff -- negligible for this repo's small `@clean` files.

- `leoCommands.py`: new `c.mod_time_cache` ivar.
- `leoAtFile.py`: `readOneAtCleanNode`/`writeOneAtCleanNode` read/write the cache instead of `v.u['_mod_time']`.
- `leoNodes.py`, `leoImport.py`, `commanderFileCommands.py`: invalidate the cache entry instead of deleting the uA.
- `leoFileCommands.py`: drop `_mod_time` from `nativeVnodeAttributes` and the now-dead `createUaList` strip; `scanTnodes` discards any legacy `_mod_time` attribute on `<t>` elements so old files self-heal on next save.
- `LeoPyRef.leo`: resynced the touched files, which also cleaned up all 7 pre-existing legacy `_mod_time` attributes (the exact spurious diffs cited in the issue).

**Second commit**, follow-up cleanup discussed in the issue thread: 6 of those 7 `@clean` nodes turned out to be flat, childless single-blob nodes (no importer-driven substructure was actually in use) -- `MANIFEST.in`, `setup.cfg`, `leo/external/leoftsindex.py`, `leo/external/make_stub_files.cfg`, `leo/plugins/leo_babel/__init__.py`, `leo/plugins/leo_babel/tests/__init__.py`. Converted those to `@edit`, which gives the same sentinel-free external file with zero outline-side content duplication (and thus no `_mod_time`-style staleness risk at all), at no cost since there was no substructure to lose. `py2cs_theory.md` (the 7th) keeps `@clean` since it has 4 real child nodes (The problem / Design / Using TokenSync class / Summary).

## Test plan

- [x] `QT_QPA_PLATFORM=offscreen python -m pytest leo/unittests` -- 853 passed, 1 pre-existing unrelated flaky failure (`test_g_OptionsUtils`, a documented CLI-arg-parsing artifact of running under pytest, not caused by this change)
- [x] `ruff check leo` -- clean
- [x] `python -m mypy leo` -- clean
- [x] Standalone repro script confirming: writing an `@clean` file no longer puts `_mod_time` in the saved XML, and bumping a file's mtime without changing its content no longer dirties the node on refresh
- [x] `leo/scripts/check_leo_sync.py` -- only the one pre-existing, unrelated `viewrendered3.py` mismatch remains (confirmed present on unmodified `devel` too)
- [x] For the 6 `@edit` conversions: verified each node's write-path output (directive-stripped body) matches the external file on disk byte-for-byte


vivainio:
@edreamleo Heads up on the second commit specifically, since it goes a bit beyond the original #4875 scope: it converts 6 of the 7 `@clean` nodes touched by this fix to `@edit`, since they turned out to be flat, childless single-blob nodes with no importer-driven substructure actually in use.

Worth calling out as a general pattern, not just a one-off cleanup: for any `@clean` node that doesn't have children, `@edit` is strictly better -- same sentinel-free external file, but zero content duplicated into the outline XML, so there's nothing to go stale and no `_mod_time`-style caching needed at all. No downside either, since there was no substructure to lose. `@clean`'s real value is the importer-driven substructure (like `py2cs_theory.md`, which I left alone here since it has 4 real child nodes); for flat files it's just paying for caching machinery you don't need.

It also helps with git: since there's no duplicated content in the `.leo` XML, there's nothing there to conflict on. A concurrent edit to `setup.cfg` (say) only ever produces a normal, easy-to-resolve conflict on `setup.cfg` itself -- not an *additional* conflict inside `LeoPyRef.leo`'s XML/pickled uA blobs on top of that, which is a much worse place to have to resolve one.

edreamleo (approved):
Very clean code. All green here.

PR #4882, opened 2026-08-08
Fix ic.languageForExtension: empty-string default masked the extension_dict fallback

vivainio:
## Summary

Found this while double-checking PR #4881 (verifying `@language` directive behavior for `@edit` nodes converted from `@clean`).

`ic.languageForExtension` (`leo/core/leoImport.py`) is supposed to check `g.app.extra_extension_dict` first (a small 3-entry override table: `pod`/`unknown_language`/`w`), and fall back to the much larger `g.app.extension_dict` (which has the real entries -- `'py': 'python'`, `'cfg': 'config'`, etc.) when `extra_extension_dict` has nothing for that extension. The fallback check was broken:

```python
z = g.app.extra_extension_dict.get(ext, '')
if z not in (None, 'none', 'None'):
    language = z
else:
    language = g.app.extension_dict.get(ext, '')
```

`extra_extension_dict.get(ext, '')` returns `''` for every extension *not* in that 3-entry table -- i.e. nearly all of them. `''` passes the `z not in (None, 'none', 'None')` check, so `language` gets set to `''` right there, and the `extension_dict` fallback never runs. In practice, `languageForExtension('.py')` returned `''` instead of `'python'`, and likewise for basically every extension.

This silently breaks `@language`-directive detection everywhere `languageForExtension` is used:
- `at.readOneAtEditNode` (`leoAtFile.py`): every `@edit` node falls back to `@nocolor` on read instead of the correct `@language` line, regardless of its actual file type.
- `ic.scanUnknownFileType` (`leoImport.py`): the `@language` line is silently omitted when importing an unknown file type.

I confirmed this isn't specific to any file I'm working with -- it reproduces on pre-existing `@edit` nodes already in `LeoPyRef.leo` before any of my changes (`launchLeo.py`, `README.md`, `CONTRIBUTING.md`), so it's a real, previously-unnoticed bug, not something new.

**Fix**: use `.get(ext)` (defaulting to `None`) instead of `.get(ext, '')`, and check `z is not None` explicitly so the "value is `''`" vs "key is missing" distinction isn't lost.

## Test plan

- [x] Added `TestLeoImport.test_languageForExtension` -- confirmed it fails against the old code (`AssertionError: '' != 'python'`) and passes with the fix
- [x] `QT_QPA_PLATFORM=offscreen python -m pytest leo/unittests` -- 854 passed, 1 pre-existing unrelated flaky failure (`test_g_OptionsUtils`, a documented CLI-arg-parsing artifact of running under pytest)
- [x] `ruff check leo` -- clean
- [x] `python -m mypy leo` -- clean (hit the documented occasional `leoShadow.py` `has-type` flake on a couple of runs, unrelated to this change -- cleared `.mypy_cache` and reran until stable; also confirmed it doesn't reproduce at all on unmodified `devel`, then reproduced and cleared 3x on this branch, consistent with known mypy inference flakiness rather than a real issue)
- [x] `leo/scripts/check_leo_sync.py` -- outline stays in sync after resyncing the two touched files

### DIFF
--- a/leo/core/LeoPyRef.leo
+++ b/leo/core/LeoPyRef.leo
@@ -971,7 +971,7 @@ c.backup_helper(sub_dir='leoPy')
 <t tx="ekr.20150514035207.1"></t>
 <t tx="ekr.20160122104332.1">@language python
 </t>
-<t tx="ekr.20160123142722.1" _mod_time="4741da9b69061c67b92e"># An example configuration file for make_stub_files.py.
+<t tx="ekr.20160123142722.1" _mod_time="4741da9ddfe1f8ea3d2e"># An example configuration file for make_stub_files.py.
 # By default, this is ~/stubs/make_stub_files.cfg.
 # Can be changed using the --config=path command-line option.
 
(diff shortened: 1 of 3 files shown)


edreamleo:
@vivainio Please resolve the conflicts.

edreamleo (approved):
All good.

It's a pleasure to have you back with us.

PR #4885, opened 2026-08-09
Bump version to 6.8.10.dev0 to test publish.yml

vivainio:
## Summary
- Bumps `pyproject.toml` version to `6.8.10.dev0` so we can cut a pre-release and exercise the full `publish.yml` pipeline (build → GitHub release asset upload → PyPI trusted-publish) end to end, without touching the real `6.8.10` release.

Related to #4884.

## Test plan
- [ ] Merge this PR into `devel`
- [ ] Create a GitHub pre-release targeting `devel` with tag `v6.8.10.dev0` (this creates the tag and the release together), which fires `publish.yml`
- [ ] Confirm dist assets attach to the GitHub release
- [ ] Confirm `leo==6.8.10.dev0` appears on PyPI
- [ ] Once confirmed working, bump version to the real `6.8.10` for the actual release

Issue #4884, opened 2026-08-09
Test the publish.yml action


edreamleo:
I am creating this issue because I'm not sure the publish.yml will work as intended.

Is there way of testing this action without actually releasing Leo 6.8.10?

vivainio:
@edreamleo Good question — as written, `publish.yml` doesn't have a safe dry-run mode: `workflow_dispatch` still runs the `github-release` and `pypi` jobs unconditionally, so testing it against the real `6.8.10` tag would actually publish it for real.

Plan to test it safely instead: bump the version to a PEP 440 dev release (`6.8.10.dev0`) via #4885, then once merged, cut a GitHub pre-release for a new tag pointing at `devel`'s tip. That fires the same `release: published` event and exercises the whole pipeline (build, GitHub release asset upload, PyPI trusted-publish) for real — but as a disposable dev version that won't collide with or block the actual `6.8.10` release later.

Once #4885 is merged, the pre-release (this creates the tag and the release together, no separate tagging step needed) can be created either way:

- Via the `gh` CLI:
  ```
  gh release create v6.8.10.dev0 --target devel --prerelease \
    --title "6.8.10.dev0 (test)" --notes "Testing publish.yml — not a real release, see #4884"
  ```
- Or via the GitHub web UI: **Releases → Draft a new release**, type the new tag `v6.8.10.dev0`, set target to `devel`, tick **"Set as a pre-release"**, then **Publish release**. Same effect — no CLI required.

Side question: do you use the `gh` CLI at all, or would you rather do this from the web UI?

Will report back here once we've run it.

edreamleo:
@vivainio Thanks for your comments.

Are you willing to be the release manager for Leo 6.8.10?

I'll soon update the distribution files in the 6.8.10 branch per the distribution checklist in leoDist.leo.

vivainio:
@edreamleo  that's fine. I will proceed with the dev0 test publish. Do NOT upload any files, ci should do it

edreamleo:
@vivainio Thank you!

vivainio:
Ran the `6.8.10.dev0` test. Two findings:

[...]

PR #4886, opened 2026-08-09
Fix publish.yml duplicate-run trigger; bump to 6.8.10.dev1

vivainio:
## Summary
- The `6.8.10.dev0` test release (#4885, see #4884) showed `publish.yml` runs **twice** per release: GitHub fires both `release.created` and `release.published` webhooks simultaneously for a non-draft release, and the workflow listened for both. The two concurrent runs raced on the PyPI upload — one succeeded, the other failed with `400 File already exists`. This also explains why both runs of the real `v6.8.9` release show as failed in the Actions history.
- Fix: restrict the trigger to `types: [published]` only, matching the pattern used in `~/zaira`'s own `publish.yml`.
- Bumps version to `6.8.10.dev1` to re-test the fix with a fresh pre-release.

Compared the `6.8.10.dev0` build artifacts (wheel + sdist) against the real `v6.8.9` release assets: same packaging structure and `dist-info` layout; the only content differences (new icons, removed `npyscreen`/`cursesGui2.py`, pruned doc sources) are real changes from the intervening two months of `devel` history, not build regressions.

Related to #4884.

## Test plan
- [ ] Merge this PR into `devel`
- [ ] Create a GitHub pre-release targeting `devel` with tag `v6.8.10.dev1`
- [ ] Confirm `publish.yml` runs exactly once
- [ ] Confirm dist assets attach to the GitHub release and `leo==6.8.10.dev1` appears on PyPI
- [ ] Once confirmed working, bump version to the real `6.8.10` for the actual release

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,7 +2,7 @@ name: Publish
 
 on:
   release:
-    types: [published, created]
+    types: [published]
   workflow_dispatch:
     inputs:
       tag:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -234,7 +234,7 @@ packages = [
 
 [project]
 requires-python = ">= 3.10"
-version = "6.8.10.dev0"  # No space.
+version = "6.8.10.dev1"  # No space.
 name = "leo"
 
 #@+<< developer info >>


edreamleo (commented):
Cool!

Issue #4892, opened 2026-08-09
Retire gh-pages branch and delete leo-editor/docs folder


vivainio:
See PR #4900.

## Context

The current docs-publishing flow rebuilds `leo/doc/LeoDocs.leo` into HTML by hand and commits the result into the `docs/` folder on the `gh-pages` branch (GitHub Pages source: "Deploy from a branch", `gh-pages` / `/docs`, `build_type: legacy`).

This adds a large generated-content diff to git history on every publish (hundreds of HTML files, tens of thousands of changed lines per rebuild), permanently, since git can't diff/compress that kind of content well. `gh-pages` history is essentially write-only bloat at this point.

## Proposal

Switch the repo's Pages source (Settings > Pages > Build and deployment > Source) from "Deploy from a branch" to **"GitHub Actions"**, and publish via `actions/upload-pages-artifact` + `actions/deploy-pages` instead of committing generated HTML anywhere.

A `.github/workflows/publish-docs.yml` workflow (workflow_dispatch) already exists that builds the docs from `LeoDocs.leo` (via `leo/scripts/build_docs.py`) and:
- always uploads the build as a plain, reviewable workflow artifact, and
- when triggered with its `publish` input checked, additionally deploys straight to GitHub Pages via `actions/deploy-pages` -- but only once the Pages source setting above has actually been switched.

## Follow-up once this is adopted

- Retire the `gh-pages` branch's `docs/` folder as a publish target (the branch can eventually be deleted once Pages no longer reads from it).
- Consider whether `docs/` on `devel` (a separate, apparently stale copy) should also be removed, since it isn't what Pages actually serves.

vivainio:
Related PR: #4893 (adds the CI docs-build check and the manual publish-docs.yml workflow; this issue tracks the follow-up of dropping the gh-pages branch commits once that's adopted).

vivainio:
Note for whoever deletes the \`gh-pages\` branch once this is adopted:

**Quantified the actual reclaimable bloat**: ~49.5 MB (674 blobs, mostly generated HTML snapshots across 122 commits) exists *only* in \`gh-pages\` history -- not reachable from \`devel\`, any other branch, or any tag. Confirmed by diffing objects reachable from \`gh-pages\` against every other ref in the repo.

A few mechanics worth knowing before/when deleting it:

[...]

PR #4893, opened 2026-08-09
Add CI check and manual publish workflow for docs

vivainio:
## Summary

- Adds a `docs` job to CI (`leo/scripts/build_docs.py`) that runs `rst3` over `LeoDocs.leo` and then `sphinx-build`, so a broken documentation build fails CI.
- Adds `.github/workflows/publish-docs.yml`, a manual (`workflow_dispatch`) workflow that builds the documentation and uploads it as a reviewable artifact. When its `publish` input is enabled, it deploys the same build through `actions/deploy-pages`.
- See #4892 for the follow-up migration away from committing generated HTML to the `gh-pages` branch, which currently adds generated output to git history on every manual rebuild.

## Post-merge activation

1. In **Settings > Pages**, change **Build and deployment > Source** from the current branch-based source to **GitHub Actions**.
2. In **Actions > Publish docs**, run the workflow with **Deploy to GitHub Pages** enabled.

The deployment continues to use the repository's existing GitHub Pages URL and custom-domain configuration.

## Bugs fixed along the way

Running `rst3` headlessly in a fresh process surfaced two pre-existing bugs:

- `leoRst.py`: `RstCommands.root` was a bare type annotation and was never initialized. The first `@rst` node processed in a fresh session crashed with `AttributeError`. This was hidden in the GUI because a long-running session eventually sets it once and retains it.
- `LeoDocs.leo`: 17 `@rst` headlines used Windows-style backslashes (for example, `@rst html\FAQ.html`) instead of forward slashes. On Linux and macOS this wrote malformed filenames rather than files under `html/`, leaving pages missing from the built site.

## Test plan

- [x] `python -m leo.scripts.build_docs` succeeds from a clean checkout with zero Sphinx warnings and all documentation sources built.
- [x] `ruff check`, `ruff format --check`, `mypy leo`, `ty check`, and `leo/unittests/core/test_leoRst.py` pass.
- [x] `leo.scripts.check_leo_sync` confirms `LeoPyRef.leo` is synchronized after the `leoRst.py` edit.
- [x] GitHub CI passes after merging the latest `devel` into this branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,23 @@ jobs:
       - run: pip install "PyQt6>=6.6" PyQt6-QScintilla
       - run: python -m leo.scripts.check_leo_sync
 
+  docs:
+    # Catch LeoDocs.leo failing to build into Sphinx HTML.
+    # See leo/scripts/build_docs.py.
+    runs-on: ubuntu-latest
+    env:
+      QT_QPA_PLATFORM: offscreen
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+          cache: pip
+      - run: sudo apt-get update && sudo apt-get install -y --no-install-recommends
+          libegl1 libgl1 libxkbcommon0 libdbus-1-3 libfontconfig1
+      - run: pip install "PyQt6>=6.6" PyQt6-QScintilla docutils sphinx
+      - run: python -m leo.scripts.build_docs
+
   mypy:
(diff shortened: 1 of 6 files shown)


vivainio:
@edreamleo This is ready to go: it is out of draft, up to date with devel, and all CI checks pass. I think we should merge it. Afterward, we will switch Settings > Pages > Source to GitHub Actions and run the Publish docs workflow with the publish option enabled to verify the first deployment at the existing Pages URL.

edreamleo (approved):
All green here. Again with the _mod_time, but we can deal with that later.

PR #4895, opened 2026-08-10
Improve menu entries for abbreviations

### LINKED ISSUE #4803: Improve menu entries for abbreviations
See PR #4895 

The abbreviations system is pretty cool 😄 but users looking for its related settings may be looking to simply turn them on on or off.

So I'd suggest two things: first, that in the menu, we rename the **settings -> List/Kill Abbreviations** menu entry to simply **Abbreviations**.

Second, that we add the **toggle-abbrev-mode** command to the two that are already there. (list and kill all)

Image

boltex:
Fixes #4803


edreamleo (approved):
All green here.

PR #4896, opened 2026-08-10
Add support for additional uppercase file extensions in LeoApp.

### LINKED ISSUE #4894: Uppercase file extensions are not recognized for @<file> nodes when determining language
See PR #4896 

# Preambule

If there is no @langage directives in effect for an @\<file\> node, a language will default to the recognized file extension, if present. 

e.g.: a node such as `@file myFile.c` will default to the C programming language if no @language directives are in effect at the top of this node.

As any leonista knows, this affects an @file (and other @\<file\> directives) sentinels which start with the 'comment character of the language of the file. Also, the color-syntax applied will be in effect. etc.

# The issue

I was working on some legacy Leo outline today from the early 2000s, and an @file node failed to load properly. (The sentinels were not recognized.) 

Turns out, it was an @file node with no specified @language directive, for which the extension was in uppercase: It failed to auto-detect the language. from the extension. 

Since I remember specific that outline working fine back in the days, I guess that when Leo was trying to check the @language in effect from a possible file extension it was simply not case-sensitive.

## Could that be brought back?

I don't see any drawbacks or undesired side effects in recognizing such filenames as myFile.ASM, myFile.X86, myFile.PAS, myFile.BAS myFile.CBL  etc... to be effectively assembly, pascal, basic and cobol. 

Moreover, if present, an @language directive will override and have precedence anyways. So, to me it feels safe to allow for such uppercase extensions to be recognized. 

I suspect that this would be trivial to change back to being case insensitive. But I wonder if anyone sees any reason to keep it case sensitive... 😄

boltex:
Fixes #4894

### DIFF
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -539,10 +539,6 @@ class LeoApp:
 
             "JSON":     "json",
             "MD":       "md",
-            "INI":      "ini",
-            "TXT":      "plain",
-            "SQL":      "plsql",
-            "TEX":      "latex",
             "YAML":     "yaml",
             "VIM":      "vim",
             "SH":       "shell",
--- a/leo/core/leoApp.py
+++ b/leo/core/leoApp.py
@@ -352,6 +352,7 @@ class LeoApp:
             "c":        "c",
             "c++":      "cplusplus",
             "cbl":      "cobol",  # Only one extension is valid: .cob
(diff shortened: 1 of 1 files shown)


edreamleo (approved):

edreamleo:
@boltex All green here.

PR #4899, opened 2026-08-10
Include home page in published docs

vivainio:
## Summary

- Copy the existing hand-maintained `docs/index.html` homepage into the Sphinx output directory after a successful build.
- Fail the build if the deployment root still has no `index.html`.

The previous branch-based Pages deployment published the whole `docs/` directory, so this homepage was included even though Sphinx generates `leo_toc.html` as its root document. The new Actions workflow publishes only the Sphinx output, which omitted the homepage and left the Pages root without an index document.

## Test plan

- [x] `python -m leo.scripts.build_docs --out /tmp/leo-docs-index-fix`
- [x] Generated `index.html` is byte-for-byte identical to `docs/index.html`
- [x] `ruff check leo/scripts/build_docs.py`
- [x] `ruff format --check leo/scripts/build_docs.py`
- [x] `python -m pytest leo/unittests/core/test_leoRst.py`
- [x] `python -m leo.scripts.check_leo_sync`

### DIFF
--- a/leo/scripts/build_docs.py
+++ b/leo/scripts/build_docs.py
@@ -19,6 +19,7 @@ Usage:
 import argparse
 import glob
 import os
+import shutil
 import subprocess
 import sys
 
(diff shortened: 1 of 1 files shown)


edreamleo (commented):
Looks good.

PR #4904, opened 2026-08-11
PR: Require ty

edreamleo:
I see no reason not to require ty. Adding ty to the two requirements files should be less confusing for Leo's devs.


edreamleo:
@vivainio Ready for your review and comments.

vivainio (approved):

edreamleo:
@vivainio Thanks for your review and approval.

PR #4917, opened 2026-08-12
Clarify @asis, @edit, and @nosent directives

vivainio:
Clarify the practical differences between `@asis`, `@edit`, and `@nosent` in the external-file directives reference.

The new text explains:
- whether each directive can contain a tree
- whether Leo performs structural expansions when writing
- whether changes to the external file can be read back

Also fixes a nearby grammatical typo.

Validation:
- `git diff --check`
- parsed `leo/doc/LeoDocs.leo` with `xml.etree.ElementTree`


edreamleo (approved):
Excellent.

PR #4925, opened 2026-08-13
PR: tracing_utils.py:to_unicode: fall back to reporting original string

edreamleo:
There is no point in reporting a made-up empty string.


edreamleo:
@vivainio Ready for review.

vivainio (approved):

PR #4951, opened 2026-08-23
PR: Set unused-ignore-comment = ignore

edreamleo:
Otherwise minor differences between EKR's setup and Leo's official requirements will cause problems.

### DIFF
--- a/leo/core/leoImport.py
+++ b/leo/core/leoImport.py
@@ -23,12 +23,12 @@ try:
     assert docutils.core
 except ImportError:
     # print('leoImport.py: can not import docutils')
-    docutils = None  # type:ignore
+    docutils = None  # ty: ignore
 try:
     import lxml
     import lxml.html
 except ImportError:
-    lxml = None  # type:ignore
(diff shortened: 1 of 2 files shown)


edreamleo:
@vivainio Any comments?

vivainio:
I don't think we should do it just yet. So far this was fixed by adding deps in ci, which is cheap. The first time we see an unfixable error, let's do it then.

Getting errors for these also allows to improve typing as we get newer versions of Ty, and are instructed to delete the annotations

edreamleo:
@vivainio The line `lxml = None  # ty: ignore` seems necessary on my machine, so I would say this is an unfixable error.

My plan: replace `unused-ignore-comment = "ignore"` with `unused-ignore-comment = "warn"` from time to time on my machine.

What I'd really like is to replace:
`lxml = None  # ty: ignore`
with something like:
`lxml = None  # ty: ignore[list-of-possible-errors]`
but that doesn't seem to work.

edreamleo:
Closed in favor of PR #4954.

Issue #4952, opened 2026-08-23
Script 'ty_leo' is reporting 35 dignostics


ranvik14:
For the context of this issue please see [Recent PRs replace pylint and mypy by ty](https://groups.google.com/g/leo-editor/c/bmwvQqMF7VQ).

Here's the log from the re-run:

```
    📦[user@leo-on-debian ~]$ 
    📦[user@leo-on-debian ~]$ cat /etc/os-release
    PRETTY_NAME="Debian GNU/Linux 13 (trixie)"
    NAME="Debian GNU/Linux"
    VERSION_ID="13"
    VERSION="13 (trixie)"
    VERSION_CODENAME=trixie
    DEBIAN_VERSION_FULL=13.6
    ID=debian
    HOME_URL="https://www.debian.org/"
    SUPPORT_URL="https://www.debian.org/support"
    BUG_REPORT_URL="https://bugs.debian.org/"
    📦[user@leo-on-debian ~]$ 

    ###

    📦[user@leo-on-debian ~]$ 
    📦[user@leo-on-debian ~]$ cd leo-editor/
    📦[user@leo-on-debian leo-editor]$ 
    📦[user@leo-on-debian leo-editor]$ python3 -m leo.core.runLeo --version
    Leo 6.8.11.dev1, main branch, build 0b05e16860
    2026-08-23 05:10:50 -0500
    Python 3.14.5
    linux
    📦[user@leo-on-debian leo-editor]$ 
    📦[user@leo-on-debian leo-editor]$ python3 -m leo.scripts.full_test_leo
    full_test_leo.py
    beautify_all_leo.py
    557 files left unchanged
    run_test_leo.py
    ..........................................................................................................................................................................................................s........................................................................................................................................................................................................................................................................................................................................................................................................................................................................s...............................................................................s................................................................................................................................
    ----------------------------------------------------------------------
    Ran 868 tests in 9.517s

    OK (skipped=3)
    ruff_leo.py
    warning: Selection `PLE0304` has no effect because preview is not enabled.
    All checks passed!
    setting leoID from os.getenv('USER'): 'user'
    check_leo.py: files: 93 contexts: 6631 attrs: 46332 in 2.67 sec.
    ty_leo.py
    warning[unused-type-ignore-comment]: Unused blanket `type: ignore` directive
      --> leo/commands/checkerCommands.py:20:18
       |
    20 |     mypy = None  # type:ignore
       |                  ^^^^^^^^^^^^^
[...]

PR #4965, opened 2026-08-26
Check mypy availability without importing it

vivainio:
## Summary
- `leo/commands/checkerCommands.py`'s mypy fallback-import (`try: import mypy / except: mypy = None  # type:ignore`) needed the `# type:ignore` because mypy ships full type stubs, so `ty` can actually verify the reassignment when mypy is installed -- and flags the ignore as "unused" when it isn't (see #4952).
- But `mypy`/`mypy_api` were only ever used as truthy availability checks -- the actual `mypy` command shells out via subprocess (`python -m mypy ...`) -- so the import itself was never needed.
- Replaced it with `MYPY_AVAILABLE = importlib.util.find_spec('mypy') is not None`, which needs no ignore comment in any environment, and updated the 4 call sites.
- Dropped `mypy` from `ty_leo.py`'s `check_optional_deps` list, since the spurious "unused ignore" warning it guarded against is now impossible.
- Checked the codebase's other optional-dependency fallback imports (docutils, lxml, pygments, nbformat, websockets, enchant, yaml) -- none of them qualify for the same fix, since they all use the real module API at runtime (unlike mypy, which was only ever presence-checked). So `ty_leo.py`'s dependency check remains necessary for the rest of that list.

## Test plan
- [x] `ty check leo/commands/checkerCommands.py` is clean both with and without mypy installed (verified in a throwaway venv without mypy).
- [x] `ruff check leo`, `ruff format --check leo`, `python -m leo.scripts.ty_leo`, and `pytest leo/unittests/commands/test_checkerCommands.py` all pass.
- [x] `cub sync --dry-run` confirms both edited files (both `@file` derived nodes) match `leo/core/LeoPyRef.leo`, so no outline resync is needed.

### DIFF
--- a/leo/commands/checkerCommands.py
+++ b/leo/commands/checkerCommands.py
@@ -5,6 +5,7 @@
 # @+<< checkerCommands imports >>
 # @+node:ekr.20161021092038.1: ** << checkerCommands imports >>
 from __future__ import annotations
+import importlib.util
 import os
 import re
 import subprocess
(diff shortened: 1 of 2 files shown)


edreamleo:
@vivainio Ah, *this* is the "other" PR.

edreamleo (approved):
All green here.

Issue #4961, opened 2026-08-25
Script 'ty_leo' is reporting a missing dependency


ranvik14:
For the context of this issue please see [Six new PRs now in Leo's "main" branch](https://groups.google.com/g/leo-editor/c/iPJMWcsXHqk/m/PkNgRWxwAgAJ).

Here's the log from the re-run on Fedora:

```
    Container Setup Complete!
    📦[user@leo-on-fedora ~]$ 
    📦[user@leo-on-fedora ~]$ cat /etc/os-release
    NAME="Fedora Linux"
    VERSION="44 (Toolbx Container Image)"
    RELEASE_TYPE=stable
    ID=fedora
    VERSION_ID=44
    VERSION_CODENAME=""
    PRETTY_NAME="Fedora Linux 44 (Toolbx Container Image)"
    ANSI_COLOR="0;38;2;60;110;180"
    LOGO=fedora-logo-icon
    CPE_NAME="cpe:/o:fedoraproject:fedora:44"
    DEFAULT_HOSTNAME="fedora"
    HOME_URL="https://fedoraproject.org/"
    DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f44/"
    SUPPORT_URL="https://ask.fedoraproject.org/"
    BUG_REPORT_URL="https://bugzilla.redhat.com/"
    REDHAT_BUGZILLA_PRODUCT="Fedora"
    REDHAT_BUGZILLA_PRODUCT_VERSION=44
    REDHAT_SUPPORT_PRODUCT="Fedora"
    REDHAT_SUPPORT_PRODUCT_VERSION=44
    SUPPORT_END=2027-05-19
    VARIANT="Toolbx Container Image"
    VARIANT_ID=toolbx
    📦[user@leo-on-fedora ~]$ 

    ###

    📦[user@leo-on-fedora ~]$ 
    📦[user@leo-on-fedora ~]$ cd leo-editor/
    📦[user@leo-on-fedora leo-editor]$ 
    📦[user@leo-on-fedora leo-editor]$ python -m leo.core.runLeo --version
    Leo 6.8.11.dev1, main branch, build 7f14f70450
    2026-08-24 10:23:09 -0500
    Python 3.14.7
    linux
    📦[user@leo-on-fedora leo-editor]$ 
    📦[user@leo-on-fedora leo-editor]$ python -m leo.scripts.full_test_leo
    full_test_leo.py
    beautify_all_leo.py
[...]

PR #4966, opened 2026-08-27
Check ty_leo.py's optional deps by resolvability, not import

vivainio:
Fixes the root cause reported in #4961.

`check_optional_deps()` used `importlib.import_module()`, which actually executes a module's import machinery. `ty` is a static type checker: it resolves types from a package's on-disk source and never runs that machinery itself, so this check was stricter than what `ty` actually needs.

This bit `pyenchant` on Linux: `pip install pyenchant` installs the Python bindings, but *importing* them also loads the native `libenchant` C library via ctypes — which isn't bundled by the PyPI wheel and may not be present as a system package. That runtime import failure got reported as "missing package" even though pyenchant was fully pip-installed and perfectly resolvable for typing purposes (confirmed: `ty` passes fine with pyenchant installed but `libenchant` unavailable).

Switched to `importlib.util.find_spec()`, which locates a module without running its import-time code — matching what `ty` itself actually relies on.

## Test plan
- [x] `python3 leo/scripts/ty_leo.py` passes the optional-deps check and runs `ty` successfully
- [x] `ruff check` / `ruff format --check` pass on the changed file
- [x] Verified `importlib.util.find_spec` resolves all 9 packages in the list without triggering their import-time side effects

### DIFF
--- a/leo/scripts/ty_leo.py
+++ b/leo/scripts/ty_leo.py
@@ -4,7 +4,7 @@
 ty_leo.py: Run ty on all of Leo.
 """
 
-import importlib
+import importlib.util
 import os
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


edreamleo (approved):
@vivainio Thanks again for your work.

PR #4967, opened 2026-08-28
PR: enable VR commands for c only if the VR plugin is enabled for c

edreamleo:
## Highlights

- [x] Disable all VR functions when viewrendered.py isn't enabled in the relevant `@enabled-plugins` node.
- [x] Import optional modules as late as possible. Kinda like Python 3.15 [lazy imports](https://docs.python.org/3.15/whatsnew/3.15.html#whatsnew315-lazy-imports).
  However, import security is more important than efficiency here.
  Exception: initVR imports docutils so it can warn about warning without docutils.
- [x] Add initVR. It injects the c.vr ivar.
   - Only DynamicWindow.create_layout should call this function.
   - This is the *only* function that creates instances of the ViewRenderedController class.
- [x] Allocate widgets at most once per outline. They are *never* deallocated! Add vr.hide_all.
- [x] Send messages to the Help tab if neither VR nor VR3 is enabled.

## Details

- [x] Rewrite getVR: it just returns c.vr.
- [x] Revise show_scrolled_message. It uses c.vr.
- [x] Use walrus operator when testing c.vr.
- [x] Revise LeoFigureManagerQt.__init__.

## Deleted methods

- [x] Delete top-level on_create function.
- [x] Hurray! Delete vr.destroy_widgets.
- [x] Remove the unused(!) LeoQtGui.runScrolledMessageDialog method.

## Testing

Besides the typical unit tests, I exercised the VR plugin using tests in leo/test/test.leo.

- [x] Switch between graphics (QWebEngineView) and text QTextBrowser widgets.
- [x] Switch between outlines.

## Discussion

- vr.hide_all eliminates all possibilities of underlying Qt race conditions. It does not cause significant flash.
- This PR may be faster than the legacy code, but the rending of graphics widgets when changing nodes can take time.


tbpassin:
VR is used to display system help messages such as help for commands. If neither VR nor VR3 are enabled, how will these messages be shown?

edreamleo:
@tbpassin Good question. I'll look into it.

leo/plugins/viewrendered.py
```diff
@@ -1146,7 +1109,7 @@ def update_katex(self, s: str, keywords: Any) -> None:
 
         # Create a new QWebEngineView.
         w = self.create_web_engineview()
-        if not has_webengineview:
+        if not w:
             g.print_unique_message('katex rendering requires PyQt6-WebEngine')
             w.setHtml(s)
```
vivainio: `w` can be `None` here (when PyQt6-WebEngine isn't available), but this branch still calls `w.setHtml(s)` unconditionally, raising `AttributeError: 'NoneType' object has no attribute 'setHtml'` instead of the intended graceful fallback.

The same pattern repeats in `update_latex` (~line 1132), `update_mathjax` (~1150), `update_pdf` (~1330), and `update_typst` (~1615) — all call `create_web_engineview()` / `create_web_engineview_with_pdf()`, which can now return `None`, then unconditionally call `.setHtml()` on it inside the `if not w:` branch.

leo/plugins/viewrendered.py
```diff
@@ -383,15 +338,14 @@ def show_scrolled_message(tag: str, kw: Any) -> None:
     c = kw.get('c')
     if not c:
         return
+    vr = c.vr
```
vivainio: `vr = c.vr` is a bare attribute access, but `c.vr` is only set by `initVr()`, which no-ops when `'viewrendered.py'` isn't enabled for `c`. If `show_scrolled_message` fires for a commander where `initVr()` was never called, this raises `AttributeError` before ever reaching the `if not vr: return` guard on the next line. (See the related comment on `qt_gui.py` below for one concrete path that hits this.)

leo/plugins/viewrendered.py
```diff
@@ -1108,12 +1060,23 @@ def update_jupyter(self, s: str, keywords: Any) -> None:
...
 
     # @+node:ekr.20180311090852.1: *5* vr.get_jupyter_source
     def get_jupyter_source(self, c: Cmdr) -> str:
         """Return the html for the @jupyter node."""
+        # # nbformat (@jupyter) support.
+        try:
+            import nbformat
+            from nbconvert import HTMLExporter
+        except ImportError:
+            return ''
```
vivainio: This early `return ''` now runs before the `body.startswith('<')` html-passthrough check below it (previously that check ran first, unconditionally). An `@jupyter` node whose body is already raw HTML used to render fine even without `nbformat`/`nbconvert` installed; now it silently renders blank on any system lacking those optional deps.

vivainio (commented):
Automated review: found a few `c.vr` / optional-dependency guard gaps introduced by this refactor. Two more (in files this PR doesn't touch, so GitHub won't let me anchor them inline) are in a separate comment below.

vivainio:
Two more issues from the same refactor, in files this PR doesn't touch (so I can't anchor them inline) — both are consequences of `c.vr` no longer always being set / `onCreate` being removed:

**`leo/plugins/qt_layout.py:141`** — `vr = c.vr` is a bare attribute access. Since `initVr()` now only sets `c.vr` when `'viewrendered.py'` is in the enabled-plugins list, this raises `AttributeError` if `viewrendered.py` is disabled while `qt_layout.py` is used (e.g. the `layout-big-tree` command). Other call sites touched by this PR use `getattr(c, 'vr', None)` for exactly this reason (see `getVr()` in `viewrendered.py` and the fix in `pyplot_backend.py`).

**`leo/plugins/qt_gui.py:1268`** — `vr.onCreate('tag', {'c': c})` — but this PR removes `onCreate` from `viewrendered.py` entirely, replacing it with `initVr(c, parent)`. This fallback path (reached when neither `viewrendered.py` nor `viewrendered3.py` is already loaded and `scrolledMessageDialog` needs to load one) will raise `AttributeError: module 'leo.plugins.viewrendered' has no attribute 'onCreate'`. This is also the path that can reach the bare `c.vr` in `show_scrolled_message` (viewrendered.py:341) before `initVr` has ever run for that commander.

tbpassin:
I don't think that importing QWebenginexxx causes any problem until a QwebEngineView is instantiated.

tbpassin:
If VR is going to be created differently, and is going to use a different way to check if it is available, some layouts may mistakenly think it's not present, or even fail. I will look into that.

tbpassin:
Even if VR is not enabled in the global enabledplugins setting, a user can enable it in a outline-local setting. Then QWebengineView, etc., will  be imported and available to all outlines. I have been doing local enabling for Freewin, which is how I discovered this possibility.

So there shouldn't be an assumption that the QWebEngineView hasn't been imported just because it's not enabled in the global settings.

OTOH, as I wrote earlier, I don't think that the mere import causes the problems we're trying to prevent.

tbpassin:
> If VR is going to be created differently, and is going to use a different way to check if it is available, some layouts may mistakenly think it's not present, or even fail. I will look into that.

Yes, one of VR or VR3 is always assumed to be present by the layout code. E.g.,

```
# A layout might want to use VR3 if it is present, else VR.
# This is indicated by using the name VRX_PLACEHOLDER_NAME in the layout.
# In building the SPLITTER dict we replace the placeholder
# by VR3_OBJ_NAME if it exists, otherwise VR_OBJ_NAME.
SPLITTERS: dict[str, Any] = dict()
for k, v in layout['SPLITTERS'].items():
    if k == VRX_PLACEHOLDER_NAME:
        k = VR3_OBJ_NAME if has_vr3 else VR_OBJ_NAME
    SPLITTERS[k] = v
```

This assumption is built in precisely because one of (VR, VR3) was needed to display system help messages, etc. at the time the new layout system was designed. These messages were called "scrolled messages", weren't they?

OTOH, the layout code checks to see if the designated widget actually exists before adding it to a splitter:

```
# Move all desired widgets into their home splitters
# SPLITTERS is an OrderedDict so the widgets will
# be inserted in the right order.

splitter_index: dict = {}
for name, target in SPLITTERS.items():
    widget = self.find_widget(name)
    if widget is None:
        widget = self.created_splitter_dict.get(name, None)
    dest = SPLITTER_DICT.get(target, None)

    if widget is not None and dest is not None:
        i = splitter_index[dest] = splitter_index.get(dest, -1) + 1
        if dest is not None:
            dest.insertWidget(i, widget)
```

The layout code tries to find the desired widgets using `find_widget()`. The code does not try to instantiate a VR instance if one does not exist, unlike the case for VR3. 

So the layout system should continue to work properly even if those initialization details change for VR - as long as the object name itself isn't changed.

tbpassin:
As I thought I remembered, VR is currently loaded even if it is not enabled in the settings files.

leo/plugins/viewrendered.py
edreamleo: All fixed at rev debe2b4e06f.

leo/plugins/viewrendered.py
edreamleo: Fixed at rev fc9d56f6.

leo/plugins/viewrendered.py
edreamleo: Fixed in recent revs.

edreamleo:
@vivainio Thanks for your comments. Recent revs have added guards for c.vr.

As for LeoQtGui.runScrolledMessageDialog, rev 7011517c removes the call to vr.onCreate.

However, there are no calls to x.runScrolledMessageDialog anywhere in Leo! Rev ffd20fb3 removes this method.

edreamleo:
@tbpassin Thanks for all your comments.

- Recent revs (various new guards) should insure that there will be no crashes if  neither VR nor VR3 is loaded.
  In particular, note the new guards in qt_layout.py.

- Despite its name, LeoQtGui.runScrolledMessageDialog was never involved in so-called "scrolled" message.
  I am investigate the actual calling sequence when the user selects the "about" part of a plugin in the Plugins menu.
  The goal will be to fall back to showing such messages in the log pane.

- I agree that importing QWebengine modules probably doesn't cause any problems, but the point of import security is to depend on as few imports as possible. I'm happy to use the "import on use" pattern in the VR plugin.

edreamleo:
@vivainio @tbpassin Recent revs revise LeoQtGui.put_help so that it falls back to putting messages to the "Help" tab.

Imo, this behavior is simple, intuitive, and best. There is no reason to enable VR just for help messages.

edreamleo:
@tbpassin This PR is ready for review. You may find parts of it useful for VR.
@vivainio This PR is ready for review.

Besides the typical unit tests, I exercised the VR plugin using tests in leo/test/test.leo.

leo/plugins/viewrendered.py
```diff
@@ -1351,7 +1292,7 @@ def update_pdf(self, s: str, keywords: Any) -> None:
 
         # Create a new QWebEngineView.
         w = self.create_web_engineview_with_pdf()
-        if not has_webengineview:
+        if not w:
             g.print_unique_message('@pdf rendering requires PyQt6-WebEngine')
             w.setHtml(s)
```
vivainio, line 1296: `w.setHtml(s)` here inside the `if not w:` branch is now dead code — `create_web_engineview_with_pdf()` can't actually return `None` any more, since `create_web_engineview()` (which it calls first) always returns a widget (falling back to `QTextBrowser`). Not a live bug today, but if that invariant ever changes again this becomes the same `AttributeError: 'NoneType' object has no attribute 'setHtml'` as before. Same pattern in `update_typst` below.

leo/plugins/viewrendered.py
```diff
@@ -1624,7 +1578,7 @@ def update_typst(self, s: str, keywords: Any) -> None:
         """Display the typest text in `s` in the VR pane."""
         # Create a new QWebEngineView.
         w = self.create_web_engineview_with_pdf()
-        if not has_webengineview:
+        if not w:
             g.print_unique_message('typst rendering requires PyQt6-WebEngine')
             w.setHtml(s)
```
vivainio, line 1503: Same dead-but-landmine `w.setHtml(s)` inside `if not w:` as in `update_pdf` above — see that comment.

vivainio (commented):
Re-reviewed this revision — the earlier `c.vr` / setHtml-on-None / jupyter-passthrough issues are all fixed (thanks!). One small leftover, not urgent:

leo/plugins/viewrendered.py
edreamleo: @vivainio Thanks. Fixed at rev ea29217f.

edreamleo:
@tbpassin @vivainio Oops. I may have found the reason why the old code deallocated widgets. Switching nodes doesn't reload the web engine content. Surely there must be a better way than allocating a new web ending widget!

tbpassin:
> Surely there must be a better way than allocating a new web ending widget!

I thought that during switching between display widgets, the old widget was not lost and was reused later. In fact I'm sure that in VR3, a new QWebEngineView isn't created each time. I can be sure because I've been setting an environmental variable to prevent the engine from using the GPU. It emits loud complaints about it. I only ever see them when loading an outline, not when switching nodes.

edreamleo:
@tbpassin @vivainio

Alright. Recent revs *do* reallocate graphics widgets when changing nodes. The new code *hides* (but never deallocates) previous graphics widgets. 

I can find no other way to get the job done. The GC is likely going to recycle old graphics widgets, provided there aren't any outstanding Qt updates pending. So this scheme *might* be a bit safer than the legacy (pre-PR) scheme, but I wouldn't bet my life on it.

For sure, never explicitly deleting widgets is a step forward.

Your comments, please.

edreamleo:
@tbpassin I reverted to creating new widgets because of strange problems switching nodes.

I'll see if I can (once again) reuse graphics widgets after taking a break.

tbpassin:
> I can find no other way to get the job done.

I think these widgets can be cached for reuse. They don't have to be subject to garbage removal.

edreamleo:
@tbpassin Yes, I think you are correct.

After talking with Rebecca, I think I know the problem. Depending on timing, a widget that has just been hidden (self.w.hide()), may become visible again as the result of a pending event. In that case, the user has *no way* to recover.

The solution will likely be a new method, hide_all, that hides not just self.w, but the two graphics widgets explicitly. Furthermore, hide_all probably should call QApplication.processEvents. That should get the job done. We shall see.

edreamleo:
@tbpassin @vivainio All's well that ends well. Rev 16d30c88 looks like it fixes the ancient problems.

This PR is ready for review.

PR #4970, opened 2026-09-01
PR: Move three plugins to the attic.

edreamleo:
- [x] Delete leo/plugins/free_layout.py, moving the code to the attic.
- [x] Delete leo/plugins/nested_splitter.py, moving the code to the attic.
- [x] Delete leo/plugins/richtext.py, moving the code to the attic.
- [x] Delete all mention of these plugins from Leo's code.
- [x] Update LeoPyRef.leo.

## Discussion

- The free_layout and nested_splitter plugins are intertwined. I see no reason to keep either one.
- The bookmarks plugin still works without free_layout.
- The richtext plugin is known to be incompatible with Qt6. It would have to be rewritten anyway.


edreamleo:
@tbpassin Ready for review.

leo/core/LeoPyRef.leo
```diff
@@ -3303,34 +3290,22 @@ jobs:
...
         with:
-          python-version: "3.12"
+          python-version: "3.14"
           cache: pip
-      # Install optional packages that Leo's devs should have installed.
-      # - run: pip install "mypy&lt;2" mypy-extensions typing_extensions
-      #     types-docutils types-Markdown types-paramiko types-PyYAML
-      #     types-requests types-six
-      - run: pip install "ty==0.0.71"
-          "PyQt6&gt;=6.6" PyQt6-QScintilla PyQt6-WebEngine
```
tbpassin: Why is ">" getting escaped with the `&gt;` entity here (also line 3304)? We don't need to escape them anywhere else in a Leo outline. And for JSON output these entities are meaningless.

edreamleo: @tbpassin I have no idea. Why do you care?

tbpassin: > @tbpassin I have no idea. Why do you care?

It's jarring for no apparent purpose and it might make some other dev think the substitution should be done other places

edreamleo:
@tbpassin Remember that these escapes occur only in the .xml (.leo) file, not in the actual sources.

The exact details of how Leo writes .leo files don't matter here, and haven't (and won't ever) change.

In short, please ignore the escapes.

PR #4980, opened 2026-09-11
fix @adoc command drops text when a node contains two or more @langua…

### LINKED ISSUE #4979: `@adoc` command drops text when a node contains two or more `@language` directives
See PR #4983.

## Summary

The `adoc` command loses part of a node's body when that node contains **two or
more `@language` directives**. Specifically, all text *before* the first
`@language` directive is silently discarded, and any section whose language does
not match the language in effect is discarded too.

With a **single** `@language` directive the output is correct, which is why the
problem is easy to miss.

## Environment

- Leo: `6.8.11.dev1`
- OS: macOS (darwin)
- Python: 3.x

## Steps to reproduce

1. Add a new node with the headline `@adoc test.adoc`.
2. Give it a child node whose body is the following (AsciiDoc with two
   embedded `@language html` directives inside source blocks):

   ```
   Some intro text before the first source block.

   .App.svelte parent component
   [source,html]
   ----
   @language html
   <script>let message = $state('hello');</script>
   ----

   .FancyInput.svelte child component
   [source,html]
   ----
   @language html
   <input bind:value={value} />
   ----
   ```

3. Run the `adoc` command.
4. Inspect the generated `test.adoc`.

## Expected behavior

The generated `test.adoc` contains the **entire** body (the `@language html`
directive lines themselves are correctly stripped):

```asciidoc
Some intro text before the first source block.

.App.svelte parent component
[source,html]
----
<script>let message = $state('hello');</script>
----

.FancyInput.svelte child component
[source,html]
----
<input bind:value={value} />
----
```

## Actual behavior

Everything before the **first** `@language html` is missing:

```asciidoc
<script>let message = $state('hello');</script>
----

.FancyInput.svelte child component
[source,html]
----
<input bind:value={value} />
----
```

Notice that the intro sentence, the `.App.svelte parent component` caption, the
`[source,html]` marker and the opening `----` are gone.

Using only **one** `@language html` directive produces correct output.

## Root cause

`MarkupCommands.write_body` (`leo/core/leoMarkup.py`) calls `g.getScript(...)`:

```python
script = g.getScript(self.c, p, useSentinels=False)
```

`g.getScript` unconditionally calls `g.extractExecutableString(...)`
(`leo/core/leoGlobals.py`):

```python
s = textwrap.dedent(s)
s = g.extractExecutableString(c, p, s)
```

`g.extractExecutableString` is designed for **script execution**: when a node
mixes several `@language` sections, only the section matching the current
language should be returned. Its logic:

```python
language = c.getLanguage(p)
...
matches = list(re.finditer(pattern, s, re.MULTILINE))
if len(matches) < 2:
    return s                      # 0 or 1 directive: no filtering

extracting = False                # <-- starts False
result = []
for line in g.splitLines(s):
    if m := re.match(pattern, line):
        extracting = m.group(1) == language
    elif extracting:
        result.append(line)
return ''.join(result)
```

Consequences when there are two or more `@language` directives:

- `extracting` starts as `False`, so **every line before the first `@language`
  directive is dropped** (never appended).
- When a section's `@language` differs from the language in effect,
  `extracting` becomes `False`, so **that section is dropped** as well.
- With exactly one directive the `len(matches) < 2` shortcut returns the whole
  body unchanged, which masks the bug.

This filtering is correct for running scripts, but wrong for `@adoc`
(and `@pandoc`/`@sphinx`), which need the entire body rendered as literal
markup.

## Proposed fix

Let `@adoc` bypass the language filtering by adding an opt-out flag to
`g.getScript` (default keeps existing behavior):

`leo/core/leoGlobals.py`:

```python
def getScript(
    c: Cmdr,
    p: Position | None,
    useSelectedText: bool = True,
    forcePythonSentinels: bool = True,
    useSentinels: bool = True,
    useExtraction: bool = True,
) -> str:
    ...
        s = textwrap.dedent(s)
        if useExtraction:
            s = g.extractExecutableString(c, p, s)
```

`leo/core/leoMarkup.py`:

```python
script = g.getScript(self.c, p, useSentinels=False, useExtraction=False)
```

`@others`, `@<<section>>` expansion is unaffected (it happens later in
`composeScript`), and the `@language` directive lines are still removed by
`remove_directives`.

## Tests

A regression test has been added in
`leo/unittests/core/test_leoMarkup.py`:

- `test_adoc_preserves_body_with_multiple_language_directives`
- `test_getScript_useExtraction`

Both pass with the fix and the first one fails without it.

```bash
python -m unittest leo.unittests.core.test_leoMarkup
```

wangzhaohe:
Fixes #4979.

**Problem**
When a node body contains two or more `@language` directives, the `@adoc` command silently drops:
1. all text before the first `@language` directive, and
2. any section whose `@language` differs from the active language.

With a single `@language` the bug is masked because `extractExecutableString` short-circuits and returns the whole body unchanged.

**Root cause**
`MarkupCommands.write_body` renders the body via `g.getScript`, which calls `g.extractExecutableString`. That function is designed for *script execution* (extract only the section matching the current language) and is inappropriate for markup commands, where the entire body must be rendered as literal text.

**Fix**
Add a `useExtraction` flag to `g.getScript` (default `True`, preserving existing behavior) and call it with `useExtraction=False` from `write_body`. `@others`/`@<<section>>` expansion still works; `@language` directive lines are still removed by `remove_directives`. This also fixes `@pandoc`/`@sphinx`, which share `write_body`.

**Changes**
- `leo/core/leoGlobals.py`: `getScript` gains `useExtraction: bool = True`.
- `leo/core/leoMarkup.py`: `write_body` calls `g.getScript(..., useExtraction=False)`.

**Test**
`python -m unittest leo.unittests.core.test_leoMarkup` — includes `test_adoc_preserves_body_with_multiple_language_directives` and `test_getScript_useExtraction`.

### DIFF
--- a/leo/core/leoGlobals.py
+++ b/leo/core/leoGlobals.py
@@ -7858,6 +7858,7 @@ def getScript(
     useSelectedText: bool = True,
     forcePythonSentinels: bool = True,
     useSentinels: bool = True,
+    useExtraction: bool = True,
 ) -> str:
     """
     Return the expansion of the selected text of node p.
@@ -7876,7 +7877,8 @@ def getScript(
             s = p.b
         # Remove extra leading whitespace so the user may execute indented code.
         s = textwrap.dedent(s)
-        s = g.extractExecutableString(c, p, s)
+        if useExtraction:
+            s = g.extractExecutableString(c, p, s)
         script = g.composeScript(
             c,
             p,
--- a/leo/core/leoMarkup.py
+++ b/leo/core/leoMarkup.py
@@ -497,7 +497,7 @@ class MarkupCommands:
         """Write p.b"""
         # We no longer add newlines to the start of nodes because
         # we write a blank line after all sections.
-        script = g.getScript(self.c, p, useSentinels=False)
+        script = g.getScript(self.c, p, useSentinels=False, useExtraction=False)
         s = self.remove_directives(script)
         self.output_file.write(g.ensureTrailingNewlines(s, 2))
 
--- a/leo/unittests/core/test_leoMarkup.py
+++ b/leo/unittests/core/test_leoMarkup.py
@@ -1,0 +1,100 @@
+"""Tests of leoMarkup.py"""
+
+import io
+
+from leo.core import leoGlobals as g
+from leo.core import leoMarkup
+from leo.core.leoTest2 import LeoUnitTest
+
+
+class TestMarkup(LeoUnitTest):
+    """Test cases for leoMarkup.py"""
+
+    def test_adoc_preserves_body_with_multiple_language_directives(self):
+        """
+        The @adoc command must write the *entire* body of a node.
+
+        Regression test for a node that contains two or more @language
+        directives.  g.getScript() used to call g.extractExecutableString(),
+        which:
+
+        (a) discarded all text *before* the first @language directive, and
+        (b) discarded the text of every section whose @language does not match
+            the language in effect.
+
+        The @adoc command now calls g.getScript(..., useExtraction=False).
+        """
+        c, p = self.c, self.c.p
+        p.b = self.prep(
+            """
+            intro text
+
+            .App.svelte
+            [source,html]
+            ----
+            @language html
+            <script>let message = $state('hello');</script>
+            ----
+
+            .FancyInput.svelte
+            [source,html]
+            ----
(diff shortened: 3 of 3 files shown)


wangzhaohe:
#4983 fixed it.

PR #4981, opened 2026-09-11
PR fixes #4977: improve rust importer

### LINKED ISSUE #4977: Improve Rust importer
Remove unnecessary `unnamed use` nodes that are direct children of the imported `@clean/@file` nodes.

For ruff_python_formatter, one inner node remains.

See PR #4981.

edreamleo:
See #4977.

- [x] Improve regexes.
- [x] Rewrite rust_i.function: move_module_preamble.

### Testing

- [x] All tests pass.
- [x] Import all files in [ruff_python_formatter](https://github.com/astral-sh/ruff/tree/main/crates/ruff_python_formatter/src) without `unnamed use` nodes. 
    Only one such node remains. It's in a unit test.


edreamleo:
@boltex The new code may affect LeoJS.
@boltex @tbpassin Ready for review!

leo/plugins/importers/rust.py
```diff
@@ -433,41 +436,49 @@ def move_module_preamble(lines: list[str], parent: Position) -> None:
...
-
-            # Next, move leading lines to the parent, before the @others line.
-            while child1.b.startswith('\n'):
-                if '@others' in parent.b:
-                    # Assume the importer created the @others.
-                    parent.b = parent.b.replace('@others', '\n@others')
+            # Scan across blank lines, /// comment lines, and use lines.
+            lines = g.splitLines(child1.b)
+            i = 0
+            for line in g.splitLines(child1.b):
```
tbpassin, line 445: I don't understand the algorithm here. First it splits `child1.b` into lines. Then it iterates though another copy of the same lines, but only does anything if a line is blank or starts with `('///', 'use')`. In this case it appends the line to the initial list of lines and then terminates the loop (using `break` - did you mean `continue`?).

Once out of the loop, it then chops off the very line(s) it just appended. This just doesn't make sense to me.

leo/plugins/importers/rust.py
```diff
@@ -433,41 +436,49 @@ def move_module_preamble(lines: list[str], parent: Position) -> None:
...
+                """
+                if lines:
+                    i = parent.b.find('@others') if '@others' in parent.b else 0
+                    parent.b = parent.b[:i] + ''.join(lines) + parent.b[i:]
+                    new_child_lines = g.splitLines(child.b)[len(lines) :]
+                    child.b = ''.join(new_child_lines)
+
+            move_lines(child1, lines)
+
+            # Move the body text all unnamed child `unnamed use` nodes.
```
tbpassin, line 473: This comment is not understandable. Should it read 

" # Move the body text of  `unnamed use` nodes."?

But where should that body text be moved? Please clarify the comment.

If the answer is "read the code", then this comment adds nothing and should be deleted. I would rather that it clearly states the intent.

edreamleo: @tbpassin Thanks for you comment! I see why you're confused. Rev c0f860fd changes the outer docstring to:
```python
"""
Move the preamble lines from the parent's children to the start of parent.b.
For Rust, the preamble consists of leading 'use' statements and any /// comments that precede them.
"""
```
So
a) the initial scan must include blank lines, /// comments and "use" statements, but
b) we want to associate any *trailing* /// comments in the first struct, def, whatever.

The code in question is almost certainly correct. It imports all the files in the ruff_python_format sources.

leo/plugins/importers/rust.py
edreamleo: @tbpassin Rev 9e78d5a improves the docstring further:

```python
    """
    Move the preamble lines from the parent's children to the start of parent.b.

    For Rust, the preamble consists of all leading blank lines, "use"
    statements, and /// comments.

    However, *trailing* /// comments belong to following enum, struct, function, etc.
    """
```

tbpassin (approved):

edreamleo:
@tbpassin Thanks for your approval. Your comments helped improve the code.

PR #4986, opened 2026-09-21
PR: Fix flashes in tree pane and log panes

edreamleo:
- [x] Fix flash in tree pane: don't call `g.es('')` in LM.load.
- [x] Fix flash in log pane:
    - LM.load updates all logs very late instead of LM.finishOpen.
    - LeoQtLog.put does *not* call w.repaint.


edreamleo:
@tbpassin This PR is ready for review. How did we ever live with the initial flashes?

leo/plugins/qt_frame.py
```diff
@@ -2496,7 +2496,6 @@ def put(
         w.moveCursor(MoveOperation.End)
         w.insertHtml(s)
         sb.setSliderPosition(0)  # Force the slider to the initial position.
-        w.repaint()  # Slow, but essential.
```
tbpassin, line 2499: If a repaint here used to be considered essential, why is it not essential any more?

tbpassin:
> @tbpassin This PR is ready for review. How did we ever live with the initial flashes?

I never even noticed flashes and now that I looked for them I still don't see them. When do you see them?

leo/plugins/qt_frame.py
edreamleo: Great question. I don't know the answer:
- Perhaps the repaint was essential in an earlier version of Qt.
- Perhaps the repaint is essential is some special cases, and now those cases are broken.
- Perhaps the repaint wasn't ever really essential.

edreamleo:
@tbpassin The previous PR fixed an ugly thingy inside the tree pane that appeared briefly (about a second) during startup. The previous PR fixed this by not drawing the outline initially.

*This* PR fixes a tiny flash in both the outline and log panes. Both flashes seem to be the result of calls to g.es.

In essence: this PR updates the log pane only once during startup.

tbpassin:
I tried three tests, and saw no flashes:

1. With 5 outlines open in Leo, I opened a new session loading only the Workbook. I got the usual dialog about an outline being open.
2. I opened a new session of Leo asking for a non-existent outline.
3. I closed Leo and opened it, letting it re-open the five that were open when the previous session closed.

I will go ahead and approve the PR. Please note that the entire startup process is complicated and I just have to assume that the rest of it is going to be OK.

tbpassin (approved):

edreamleo:
@tbpassin Thanks for your approval, Thomas. I agree that the startup process is (too?) complicated, but now the drawing part is less so.

Imo, only wider testing has any chance of revealing subtle problems. I don't expect such problems, but we shall see...

